PR #95, opened 2023-07-18
Document the cloud resources we've created

mgwalker:
This PR adds light documentation around the infrastructure resources we've created in the WOC. This should not be the final form of this documentation, but it serves our immediate needs – not forgetting what we've done. 😄


cannandev (approved):
This is great!

PR #119, opened 2023-07-31
Update github templates to reflect stories and epics

colinmurphy01:


colinmurphy01:
@mgwalker can we update the project board logic to **not** automatically pull in tickets with the "⛰️ Epic" tag. And instead have them go to [the Roadmap board](https://github.com/orgs/18F/projects/52/views/1)?

mgwalker:
I believe that works. It looks like the automation takes a filter, which should also let us do something like `-label:⛰️ Epic`

mgwalker:
Yep, we can do that. And it's done now.

.github/ISSUE_TEMPLATE/story-template.md
```diff
@@ -0,0 +1,22 @@
...
+about: Use this template to create a story and track in our sprint board.
+title: ''
+labels: 'Feature, :rock: Story'
+assignees: ''
+
+---
+
+### Description
+
+**Related Boulder: [Epic]**
```
mgwalker, line 12:
```suggestion
**Related Epic: [#Epic]**
```

mgwalker (approved):
Suggested changing `related boulder` to `related epic`, but this is good to go whenever!

PR #123, opened 2023-08-01
Tinkering: Add observation station location and distance; some cacheing updates

mgwalker:
- For current conditions, the UI now displays the location of the observation station providing the data plus the (rough) distance from the user's location to that observation station
  - The `Observation` class was updated to also capture the lat/lng/altitude of the observation station and to add a helper function for computing the distance from the obs station to an arbitrary lat/lng. Altitude is currently ignored.
  - The main page was updated to add a placeholder in the dom for the ob station location and distance
  - If we decide to display distance to obs stations, we might want to see if we can get someone at the National Geodetic Survey (a NOAA office) to verify our approach is good enough. Basically, I've implemented the [Haversine distance function](https://en.wikipedia.org/wiki/Haversine_formula), which gives you the distance between two points on the surface of a sphere. My concerns here are:
    - The distance is computed based on the surface of the sphere. It does not account for points are different altitudes, which may be significant in some areas with more sparse obs station coverage.
    - The Earth is not a perfect sphere. I think we'll be dealing with relatively small distances (dozens to maybe a hundred miles), so maybe it's _close enough_ that we don't have to worry about it. Especially since I think it's also okay for our results to have an error of a mile or two, too.

    @beechnut and I talked about this, and they thought this approach was probably good enough, but I don't see any harm in getting NGS to review it if we do go that route. Matt has a background in geography, so I'm willing to trust their judgement. Certainly more than my own. 😆

- The user's browser location data is now cached in their browser. If the location data in cache is less than 30 minutes old, we use it and then update the cache in the background.
  - In the future, the background update should also trigger some downstream updates: if the user's forecast grid point changed, we should re-fetch observation stations, observations, and forecasts, and then update the UI

- The service worker now supports cacheing arbitrary requests. For a request to be cached, it must be added to a list of cacheable endpoints. When a request comes into the service worker for a matching endpoint, it is routed through the cacheing mechanism instead of fetched directly. It will only be fetched if it is not already cached. These caches currently never expire, so we'll eventually want to come back to that. (Noted in comments.)

### DIFF
--- a/src/lib/api/observations.js
+++ b/src/lib/api/observations.js
@@ -1,5 +1,26 @@
 import convert from "https://cdn.jsdelivr.net/npm/convert@4";
 
+// equatorial mean radius of Earth (in miles)
+const R = 3_963.19;
+
+function toRad(x) {
+  return (x * Math.PI) / 180.0;
+}
+function hav(x) {
+  return Math.sin(x / 2) ** 2;
+}
+
+function haversineDistance(a, b) {
+  const aLat = toRad(a.latitude);
+  const bLat = toRad(b.latitude);
+  const aLng = toRad(a.longitude);
+  const bLng = toRad(b.longitude);
+
+  const ht =
+    hav(bLat - aLat) + Math.cos(aLat) * Math.cos(bLat) * hav(bLng - aLng);
+  return 2 * R * Math.asin(Math.sqrt(ht));
+}
+
 const wmoToLib = new Map([
   ["wmoUnit:degC", "celsius"],
   ["wmoUnit:degree_(angle)", "degrees"],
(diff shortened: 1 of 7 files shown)


src/lib/api/observations.js
```diff
@@ -1,5 +1,26 @@
 import convert from "https://cdn.jsdelivr.net/npm/convert@4";
 
+// equatorial mean radius of Earth (in miles)
+const R = 3_963.19;
+
+function toRad(x) {
+  return (x * Math.PI) / 180.0;
+}
+function hav(x) {
```
elijah-wright, line 9: is there a style guide? if so, maybe the brackets should be removed if there's only one thing being removed

elijah-wright:
I said something stupid earlier but if there's only one return statement in toRad() and hav(), do you want to use a one line arrow function?

mgwalker:
Hi @elijah-wright. We have not adopted a style guide yet. This is very much all prototype code and will likely all be thrown out before long. Once we start writing code that we expect to stick around, we'll use [eslint with the Airbnb preset + prettier](https://engineering.18f.gov/javascript/#style). So for now, we're not super interested in code quality or maintainability. 🙂

elijah-wright:
awesome, good to know

PR #137, opened 2023-08-10
Tinkering: a map view of current temperatures associated with a WFO

mgwalker:
This adds some tinker code that pulls a list of approved observation stations for a WFO from the API and puts them on a map centered over Nashville. It's just here to get it out of my head, as a thought about what it means to display current conditions for a place.

Also adds a very basic stress testing script. It uses multiprocess curl to hit a target however many times specified. Coupled with `time`, it'll give us something to compare CMSes with. While this test doesn't tell us how well a CMS will perform in our production environment, it should give us good data about how they compare to each other.

### DIFF
--- a/src/grids.html
+++ b/src/grids.html
@@ -53,7 +53,7 @@
           forecastOffice,
         ).then((r) => r.json());
 
-        approvedObservationStations.map(async (stationUrl) => {
+        approvedObservationStations.forEach(async (stationUrl) => {
           const {
             features: [
               {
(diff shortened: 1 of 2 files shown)


src/grids.html
```diff
@@ -0,0 +1,80 @@
...
+    <script
+      src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"
+      integrity="sha512-XQoYMqMTK8LvdxXYG3nZ448hOEQiglfqkJs1NOQV44cWnUrBc8PkAOcXy20w0vlaXaVUearIOBhiXZ5V3ynxwA=="
+      crossorigin=""
+    ></script>
+  </head>
+  <body>
+    <div id="map"></div>
+    <script type="text/javascript">
+      const nashville = [36.15, -86.77];
```
mgwalker, line 25: Roughly the middle of Nashville. Coordinates taken from Wikipedia.

PR #138, opened 2023-08-10
Add zscaler root certificates to our Docker container

mgwalker:
For 18F users, all our outbound internet traffic is now routed through a proxy that terminates the SSL connection so it can do traffic inspection. The proxy uses certificates issued by a root that is not included in any OS or browser default stores (and for good reason: the certificate is being abused). When the proxy configuration was pushed to our machines, the root CA was also installed to our system certificate stores.

Docker containers on macOS don't use the system certificate store. As a result, outbound internet traffic over SSL invariably fails. In production that might actually be fine because I don't think we'll want our CMS initiating outbound traffic. In dev, though, it's problematic because it prevents us from using system tools to install plugins.

This PR adds a Dockerfile for creating a custom image instead of using the base Drupal image. The Dockerfile adds the zscaler root CA to the container, configures curl to use it, and configures PHP to use it. The PHP configuration part is a little heavy-handed, but as this is for dev, I was more interested in "make it work" than "make it work well."


drupal-volume/default-site/settings.php
```diff
@@ -264,7 +264,7 @@
  *   $settings['hash_salt'] = file_get_contents('/home/example/salt.txt');
  * @endcode
  */
-$settings['hash_salt'] = 'e-xFAOi5RgKgkAYZSraObr3CW8ZxalG8dmDk6-SUjT_PiHdFpRWmZaNJOyxXsmrzW9Zka7DROQ';
+$settings['hash_salt'] = 'bQxfHu3DvQrom6_zyl9lncicvDN6uoOpqiQZhwYpSYLimSLERTTbuUA-bfcd0KHTb3xgJpOyCw';
```
mgwalker, line 267: I think I rebuilt my site and Drupal automatically changed this value. 🤷🏻‍♂️

drupal-volume/default-site/settings.php
```diff
@@ -839,7 +839,9 @@
   'prefix' => '',
   'host' => $_ENV['DRUPAL_DB_HOST'],
   'port' => $_ENV['DRUPAL_DB_PORT'],
+  'isolation_level' => 'READ COMMITTED',
   'namespace' => 'Drupal\\mysql\\Driver\\Database\\mysql',
   'driver' => 'mysql',
   'autoload' => 'core/modules/mysql/src/Driver/Database/mysql/',
 );
+$settings['config_sync_directory'] = 'sites/default/files/sync';
```
mgwalker, line 847: It seems that Drupal will ask to setup the database again if these values are missing. By adding these values, Drupal will happily "just work" provided that the database is already configured.

PR #143, opened 2023-08-11
Normalize our Docker and Drupal config

mgwalker:
## What it does

- Pulls more into our Dockerfile instead of relying on the base Drupal container
   - Specifically, our Dockerfile is now roughly a duplicate of the Docker-maintained image, except we create a composer project using the `drupal-composer/drupal-project` template instead of the `drupal/recommended` project. The `drupal/recommended` project does not include drush, and no amount of fiddling with it made drush work correctly without breaking the site.
- Gets rid of the `drupal-volumes` folder and moves its contents into the project root
   - This is following the direction @cannandev went with developing a custom theme. Might as well do it the same way across the board!
   - Gets rid of the `profiles` folder entirely.
   - Updates the `docker-compose` to map the new folders into the container
- Moves the saved config from `config/config-localhost_8080-2023-08-11-00-39` into just `config`
   - Updates the config to enable our custom toy content API module
- Adds `config` to the folders mapped into the container, and update `settings.php` to point to it as our config sync directory
- Adds docker tools for installing the site from our config, importing from config, and exporting to config
   - Pulls the repeated parts of our docker service definitions into a separate file so they can be reused, instead of repeating everything 4 times.
- Adds a Makefile to wrap the docker tool commands in easier-to-type commands.
- Updates the README

## Recommendation for after merging

When this gets merged, devs should stop and remove their current docker containers. This will wipe out your database. Then, pull the latest from `main`, and run the following:

```shell
docker compose up
make install-site
```

This will install a minimal Drupal site with our saved configuration. User 0 will be `admin` with password `root`. The installation will include the custom theme, but that theme won't be enabled because when the theme is enabled, there's no way to get to the login page. So... login first, _then_ enable the theme. (I assume we'll have a login page in our theme later, at which point the theme can be enabled in our standard config.)

## Going forward

As we make changes to site configuration, we can export those to files so we can version/share them by running:
```shell
make export-config
```
or
```shell
docker compose run --rm export-config
```
Then commit the changes in the `config` directory and include them in your pull request.

To import config changes, run:
```shell
make import-config
```
or
```shell
docker compose run --rm import-config
```

There is also a helper tool for clearing the Drupal caches:
```shell
make clear-cache
```
or
```shell
docker compose run --rm clear-cache
```


mgwalker:
Converting to draft. The way the helper scripts handle drush breaks the site. So... I'm going to prioritize working on that today before we merge this. :sigh:

cannandev:
Works great for me! There's just one error I get when running `make install-site`. I get:

`bash: line 1: drush: command not found`

Does Docker care that I'm running zsh? 

But I can get to the install profile and config just fine. And yes, we'll be able to login with the hello_world theme. Then we can add it to the default enable themes and modules in  `composer.json`.

PR #150, opened 2023-08-21
Mgwalker/hello world theme

mgwalker:
## What's wrong?

The setup from Friday doesn't work right with 3rd-party modules. It uses `composer create-project` with a Drush-provided project template, which includes its own composer files. To add 3rd-party modules, you would then need to get a shell inside the container and install them yourself. However, they wouldn't be persisted because the container had its own isolated composer files.

## How does this PR fix it? 

Instead of using `composer create-project`, the Dockerfile now copies our composer files into the container at build-time and simply runs `composer install --dev`. Then, `docker-compose` makes our composer files into the container. That way, the initial build uses our composer files and any changes we make get persisted back into our repo.

Since I was in here anyway, I also changed how `settings.php` gets shared into the container, so it's only that one file instead of the whole directory. This way, we get all of Drupal's default files that it requires, but we don't actually have to keep them in our repo.

## Types of changes

## Checklist:


.gitignore
```diff
@@ -4,5 +4,4 @@ vendor
 /.editorconfig
 /.gitattributes
 default-site/files
-modules/contrib
```
mgwalker, line 7: The `contrib` directory is now explicitly excluded from our host-to-container mapping, so we don't need to exclude it from git anymore.

Dockerfile
```diff
@@ -69,10 +69,15 @@ RUN echo "openssl.cafile=/var/certs/zscaler-root.pem" >> /usr/local/etc/php/php.
...
+# Copy in our composer files to start with.
+COPY composer.json /opt/drupal/composer.json
+COPY composer.lock /opt/drupal/composer.lock
+
 WORKDIR /opt/drupal
+
 RUN set -eux; \
   export COMPOSER_HOME="$(mktemp -d)"; \
-  composer create-project --no-interaction "drupal-composer/drupal-project:10.x-dev" ./; \
+  composer install --dev; \
```
mgwalker, line 80: Instead of using `composer create-project`, we can just use our own composer files. But make sure to install dev dependencies too.

README.md
```diff
@@ -14,55 +14,59 @@ We will know that we are progressing forward on this problem if people find the
 
 ### Our vision is that anyone can understand the impact of impending weather. When it comes to making decisions to save life and property, every word and every minute matters.
 
-**Our mission** is to rebuild weather.gov to reflect the integrity and care NWS has for the people you serve.
```
mgwalker, line 17: My prettier config cleans up Markdown files. I highly recommend reviewing this file with the rich diff instead of the source diff.

config/filter.format.basic_html.yml
```diff
@@ -37,12 +37,6 @@ filters:
     status: true
     weight: 9
     settings: {  }
-  filter_image_lazy_load:
-    id: filter_image_lazy_load
-    provider: filter
-    status: true
-    weight: 15
-    settings: {  }
```
mgwalker, line 45: I'm not sure where these came from, but there's not an associated plugin with these so keeping them in causes Drush to complain when it's importing the settings. So... I just got rid of these. Seems to work? 🤷🏻‍♂️

cannandev:
>This way, we get all of Drupal's default files that it requires, but we don't actually have to keep them in our repo.
Brilliant 💡

config/filter.format.basic_html.yml
cannandev: The default format for **Body** and other long text fields is `basic`. I think I as playing around in there. Totally fine to remove it.

mgwalker:
I think for testing this, you should be able to checkout the branch and type `make rebuild`. That will delete your current container, rebuild from source, and restart the container.

cannandev:
Works as expected for me!

PR #152, opened 2023-08-25
Cannandev/update drupal core version 10.1.2

cannandev:
## What's wrong?
📢 New Drupal core update available. Core should be updated as soon as a new version is available, since it usually involves security updates.

## How does this PR fix it? 
Updates to 10.1.2

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,3 +128,29 @@ are reflected in the host folder.
 That's it! Now when you make changes to theme files, they will sync to the
 Docker instance. Whenever you make a change to a .twig template, make sure
 to **rebuild the cache** using the `make clear-cache` command.
+
+## Installing Drupal core updates
+
+### Development
+When a new version of Drupal core is available, use Composer to intall the latest updates. See [Updating core](https://www.drupal.org/docs/updating-drupal/updating-drupal-core-via-composer#s-update-drupal-core-code) docs.
+
+Run `composer show drupal/core-recommended` to see the latest Drupal core version. Then run:  
+
+```
+composer update "drupal/core-*" --with-all-dependencies
+```
+
+This will update the required projects: `drupal/core-recommended drupal/core-composer-scaffold drupal/core-project-message`
+
+The update the db using drush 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -128,3 +128,29 @@ are reflected in the host folder.
...
+```
+make shell
+drush updatedb
+drush cache:rebuild
+```
+
+### Production
+1. Push the changed composer.json and composer.lock files to production.
+1. run `composer install --no-dev` on production, rather than composer update.
+1. run `drush updatedb` or visit update.php
```
mgwalker, line 156: @loganmeetsworld Just want to call your attention to this in particular. I think I have a couple of big questions here:

1. Rather than pushing updated composer files, would we want to rebuild our container image and redeploy it whole? That is no, no in-place updates, only replacements? I think we'd have to, since new containers could spin up in response to demand at any time, and they wouldn't have the updated composer files unless we put those in a shared place. Even then, spinning up a container would necessarily require running `composer install` again, which would make them slower and I think kinda negate the scale-up performance benefits of containers?
2. Regardless of how we update Drupal, how do we update the database? Maybe more of a question for @cannandev, but what happens if we have, say, 10 Drupal containers running. Half of them are running a newer version of our container, and half are still on the previous release. If we update the database, will the older half stop working?

This is all out of scope of this PR. I just wanted to capture the questions I was having.

README.md
```diff
@@ -128,3 +128,29 @@ are reflected in the host folder.
...
+```
+
+This will update the required projects: `drupal/core-recommended drupal/core-composer-scaffold drupal/core-project-message`
+
+The update the db using drush 
+
+```
+make shell
+drush updatedb
+drush cache:rebuild
```
mgwalker, line 150: Does `updatedb` make any changes to configuration that we need to capture? That is, do we also need to run a `make export-config` afterwards?

mgwalker (approved):
vroom vroom

PR #153, opened 2023-08-25
Performance testing scripts

mgwalker:
## What's wrong?

- The basic `stress-test.sh` utility defaults to much too large a number of requests
- We don't have a consistent way of profiling API endpoint performance

## How does this PR fix it? 

- Changes the `stress-test.sh` default to 10 requests per process instead of 100
- Adds a `api-perf-test.js` script to use Node.js's high-precision performance monitoring APIs and make calls to some URL lots and lots of times to get average, median, min, and max response times

## Types of changes

## Checklist:


loganmeetsworld (approved):
Could we put these under a scripts folder?

PR #172, opened 2023-09-05
Overhaul the file and directory structure now that we've picked a CMS

mgwalker:
## What's wrong?

Our repo currently contains files we don't need and the folder structure developed organically as-needed. It's fine, but if we're past this part of prototyping, now we can pause and take the time to intentionally design a directory structure that's easy for us to use.

## How does this PR fix it? 

* The parts necessary to create our Drupal instance are pushed into the `drupal` directory
   * settings.php
   * PHP composer files
   * exported configuration files
   * Dockerfiles
* Custom themes and modules are pushed into the `src` directory
* Simplifies the Docker composer by pulling environment variables and volumes into the root composer file and eliminating the "tool" containers
   * Tool runs are and were accomplished with `docker compose exec` commands in a Makefile, but there's no point in having separate containres to run those in. This update runs those `exec`s in the main Drupal container.
* Removed `scripts` and `src` directories that were related to prototyping. If we want to refer back to it, it'll be in our git history, but we don't need to keep carrying it with us.

## Types of changes

- [x] ⚒️ Breaking change (fix or feature that would cause existing functionality to not work as expected)

Strongly recommend that devs completely destroy their current dev environment and start fresh. If you're working on plugins, themes, etc., export your config first so you can re-apply any changes!

## Checklist:

   - We need to adopt a code style!


cannandev (approved):
😆 Page kept freezing and crashing mid-review. LGTM 🚀 
Also, thanks for updating the `make` commands to directly use drush! that way I don't have to shell in.

PR #176, opened 2023-09-06
Linting

mgwalker:
## What's wrong?

Nothing yet, but things might be wrong soon if we don't adopt a linter and formatter!

## How does this PR fix it? 

* Adds an action to pull requests to run PHP_CodeSniffer on our modules and themes
* Adds two `make` commands:
  * `make lint` will run the same linting as the PR action
  * `make format` will format the code in modules and themes according to the same rules as the linter
* Removes the prototype `content-api` module

## Types of changes

## Checklist:


mgwalker:
There's a VS Code plugin that might be useful on dev machines. It would let us setup arbitrary commands to run on file save, so we could have it run `make format` automatically. This is useful for devs who don't run composer locally and are working entirely from the container.

If you're running composer locally, I think I found some other phpcs plugins that will work but I'll need to go searching again.

Either way, both of ☝🏻 those would be local-only configuration, not anything that would be committed to the repo, but if they can be helpful for dev, we should still document it!

.github/workflows/code-standards.yaml
```diff
@@ -0,0 +1,13 @@
+name: code standards
```
loganmcdonald-noaa, line 1: what else do you imagine putting in this workflow?

loganmcdonald-noaa (approved):

Issue #175, opened 2023-09-05
e3.v1 current conditions (for a location)


janelx:
[Figma link](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=363-2510&mode=dev)
On pages/layouts: #174. Maybe: #173, #160  

## Supported user stories
- 3.1 … I want to understand that the weather information I'm seeing is relevant to my specific location
- 3.2 … I want to view current conditions

## Support needed
- Weather APIs
- [New icons and mapping](url) 
- Automated text summary ?? weather.gov might have a current solution??

- [x] complete #34  
## Acceptance Criteria

*Not part of this story: search/change location.*

**As a member of the public I can easily understand:**

- what the current weather is for a location
- Use a screen reader to listen/read a quick and easy to understand summary of what is shown on screen

----
### General

- [ ] Responsive
- [ ] USWDS friendly
- [ ] Section 508 compliant

### Display current weather for a single location
[Figma link](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=363-2510&mode=dev)
![Image](https://github.com/weather-gov/weather.gov/assets/93934808/a280c0b2-ccc4-446d-98a8-67d8c80c571f)
[current conditions api (cc)](https://github.com/weather-gov/weather.gov/blob/main/docs/data.md#current-conditions-observational-data)

### Timestamp

<kbd> ![Image](https://github.com/weather-gov/weather.gov/assets/93934808/3fb0b431-4b1a-4c73-8210-430b20b910e1) </kdb>

- 11:11 AM CDT > [timestamp] 
- gray-cool-50 
- 0.75 rem

### Basic weather information
![Image](https://github.com/weather-gov/weather.gov/assets/93934808/474e1074-9f00-4633-a3aa-ffeb37d5b799)

**Weather icons**
- New NOAA Icons - See [google sheet for mapping](https://docs.google.com/spreadsheets/d/1Er4nW_WA2nu-9GdeoUwT4JkSbfp7B84hUnXa32PQcdk/edit#gid=1809319678)
  - Replace old NWS weather condition icons with corresponding NOAA icon. 
  - Some day/night mapping may use the same icon. 
- Height = 64px
- Width = unconstrained/relative to height
[...]

Issue #189, opened 2023-09-11
Install USWDS and dependencies


cannandev:
Install US Web Design System according to [developer instructions](https://designsystem.digital.gov/documentation/getting-started-for-developers/) using Node Package Manager (npm). 
Incorporate compilation of SASS/SCSS style files into our developer workflow.

cannandev:
Related to #188

colinmurphy01:
Interesting delivery process use case here. 

Since this is more foundational architectural work... should this go through Des, Eng, PO review prior to DONE or just Eng? @greg-does-weather interested in your take?

greg-does-weather:
I think this can just be engineering. How we _use_ USWDS should go through design and product review, but just for getting it installed and setup, there's not really much to review outside of engineering.

Issue #161, opened 2023-09-01
e2.v1 styles, header, footer


janelx:
🎨 [Link in figma](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=345-2102&mode=design)

### Supported user stories
… I want to know this is an official NWS and gov website (2.2)

### Design (define/discovery)
- [x] Done - part of layout and visual exploration #111 

### Support needed
- USWDS & NOAA style guide - Banner, Footer ([example of approved implementation on NOS](https://oceanservice.noaa.gov/))
- CMS:  footer links
- Editor - who can update these links?
- [x] #189
- [ ] Build menus and forms to capture site-wide info (logos, site title, site alt, etc)
- [x] Build responsive base template with regions for banner, alerts, footer menus, etc.

### Permissions and roles
- [ ]  Need definition

### Acceptance criteria
_Not included in this story (will be a separate story): navigation, alerts, site search)_

**User story**
- [ ] As a member of the public I can quickly identify and trust that weather.gov is an official website of the federal gov't and the official website for NWS

**General:**
- [ ] Responsive
- [ ] USWDS friendly
- [ ] Section 508 compliant
- [ ] Translation friendly

#### Styles:
- [ ] Based on USWDS - default to USWDS unless setting mentioned below:
  - [ ] Font-family: "Nunito Sans","Open Sans","Helvetica Neue",sans-serif
  - [ ] Brand colors: 
    - [ ] sky blue = #003087 (color-primary)
    - [ ] water blue = #0085CA (color-primary-light)
    - [ ] white =  #fff
    - [ ] dark gray = #333
- [ ] Body links: color = #003087 (color-primary)

#### Header:
![image](https://github.com/18F/weather.gov/assets/93934808/18d97d5e-4a7c-4e3e-a418-b22e32a9f622)
*The header follows NOAA style guidelines and based on [USWDS header component](https://designsystem.digital.gov/components/header/). Note: NOAA has a guideline for displaying alerts but will be part of the alert ticket.*

- [ ] Shows on every page
- [ ] Correct NOAA logo link to github [svg](https://github.com/18F/weather.gov/blob/main/docs/designs/e2v1-styles-header-footer%20/noaa-logo-rgb-2022.svg)) 
  - [ ] Links to weather.gov
  - [ ] Alt text should be “National Oceanic and Atmospheric Administration logo” and coded as a link 
[...]

PR #196, opened 2023-09-19
e3.v1 - Static implementation of the current conditions component

greg-does-weather:
## What does this PR do? 

This PR adds a new module called `weather_blocks` and updates the `hello_world` theme to define how it is rendered. The site configuration is also updated to include the module.

### Service

This module defines a new service for fetching weather data. Its purpose is to abstract away the sources of truth and allow our blocks, etc., to only worry about the service. The service will handle getting data from the original sources.

Currently uses `Drupal::httpClient()` and should be refactored to get this via dependency injection to make it easier to test. We can probably also write a lightweight wrapper around api.weather.gov calls since they all come back the same way and it'd be nice to have fewer instances of `json_decode($thing->getBody())`.

### Block

This module defines a new current conditions block. The block has no rendering associated with it, instead injecting data into any Twig templates that implement the block. The following data is available:

- conditions
  - long - long description of current conditions
  - short - short description of current conditions
- feels_like - a temperature in degrees Fahrenheit
- humidity - relative humidity in percent
- icon - icon name (name *only*, not a full path)
- location - location for these conditions (e.g., "Nashville, TN")
- temperature - a temperature in degrees Fahrenheit
- timestamp
  - formatted - the timestamp of the conditions, formatted to the America/New_York timesone
  - utc - UTC timestamp, as seconds since the Unix epoch
- wind
  - direction - direction the wind is coming from as one or more letters (e.g., "E" or "NE")
  - speed - angular direction of the wind; north = 0, east = 90, south = 180, west = 270

For this PR, all of the data is hardcoded except the timestamp, which just uses the current time.

### Theme

1. Copied the NOAA weather icons into the theme's assets
2. Adds a template to handle the new block type
   > [!IMPORTANT]  
   > For now, the template defines its own styles right inside the template file, for simplicity. Longer term, we need to decide how we want to do all that. It's kind of nice having the styles defined right alongside the things that use them, though.
3. Adds a Javascript file that finds all `<weather-timestamp>` elements, reads the UTC timestamp from their `data-utc` attributes, and replaces their contents with an appropriately-formatted timestamp for the user's locale.

   The template renders timestamps like this:
   ```html
   <weather-timestamp data-utc="{{ timestamp.utc }}">
      {{ timestamp.formatted }}
   </weather-timestamp>
   ```

   `timestamp.formatted` is in our default timezone, `America/New_York` (EST or EDT, depending). This Javascript will replace the server-rendered timezone with a client-rendered one instead. If Javascript is disabled or fails, the default timezone display will remain.

---

Part of #175.

The next step here will be to hook all of this up to a real location. That's likely to come somewhere later, though... Still, there should probably be an issue for it before we merge this.

Screenshot 2023-09-19 at 9 02 02 AM

## Types of changes

- [x] ⚒️ Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


greg-does-weather:
If this is merged before #189 is finished, we should create a tech debt issue to come back and fix the styles. Right now they are embedded inside the Twig template, which might not be where we want them. But we don't have to solve that right now; moving text between files is easy. 😛

claire-skies-noaa (approved):
LGTM

* We can revisit the in-template styles later. I'm going to talk @janelx about standardizing margins and padding with #161

claire-skies-noaa (commented):
Just caught this 

`src/themes/hello_world/images/icons/humidity.svg` should be in the `weather` folder.

PR #197, opened 2023-09-19
Rearrange local filesystem to match container filesystem

greg-does-weather:
Since the bulk of this PR is just moving things on disk, I figured the most reasonable way to review it would be based on what the file tree would look like after merging:

```
.github/
docs/
web/
   |__config/
   |  |__<saved config files>
   |__modules/
   |  |__<weather.gov custom modules>
   |__sites/
   |  |__default/
   |     |__settings.dev.php
   |     |__settings.prod.php
   |     |__settings.php
   |__themes/
      |__<weather.gov custom themes>
.gitignore
composer.json
composer.lock
CONTRIBUTING.md
docker-compose.yml
Dockerfile.dev
Dockerfile.prod
LICENSE.md
Makefile
package.json
package-lock.json
README.md
zscaler-root.pem
```

After this, running `composer` locally should have the same effect as running `composer` inside the container, maybe, hopefully.


loganmcdonald-noaa (approved):
Seems reasonable to me as long as it works for you locally! I have a change to make to how we're doing settings but will do after PR merges and I get my cloud.gov stuff setup.

PR #203, opened 2023-09-21
QASP draft

greg-does-weather:
## What's wrong?

We do not have documentation about what we think is important for building a sustainable, maintainable system.

## How does this PR fix it? 

This PR creates a QASP that defines what we think are important for that purpose, along with what we think good performance towards those things are.


claire-skies-noaa (approved):
LGTM

warrickmoran:
System Level QASP: 

Considering abstraction levels, a system QASP would be based on our operational requirements and the strategies and activities for ensuring the quality and performance of the "system." Aspects of configuration management, documentation, reviews/audits, and metrics would be detailed with an understanding of roles and procedures (where needed.)  

As mentioned, this may be an out-of-scope of the current conversation and an NWS responsibility.

warrickmoran:
Although implied under Code 1 - should we include configuration management/branching scheme (strategy) as an item?

colinmurphy01:
@loganmcdonald-noaa do you think any of the standard ATO documentaton will cover the system-level requirements Warrick is mentioning above? I want to avoid duplicating documentation if that's something we will do during that process. 

If not, then yes we can spin up a separate ticket and prioritize for later sprints.

PR #213, opened 2023-09-25
Setup the ability to deploy our container to cloud.gov

loganmcdonald-noaa:
# What this PR does

This pull request sets up the configuration to deploy what we are currently running locally in Docker to cloud.gov. It does this mainly in the `manifest.yml`, but there are some additional changes that needed to be made. This can be deployed to any provisioned cloudfoundary organization - right now it is sitting in a sandbox that can be reached at https://weather.app.cloud.gov/

# What this PR does NOT do

- While s3fs is installed, it is not yet configured to run, meaning any file uploads will be lost on deploys.
- post-deploy needs to be setup correctly to ensure we are managing configuration the right way and not losing anything on re-deploy. 
- Eventually, we would need to setup on a real cloud.gov organization so that we can have various environments. Until then, it will not be possible to setup continuous deployment through CI.
- Yet add all of the operations documentation around cloud.gov which I would want to add if we use this moving forward

# Notes to reviewers
- I am following the recommended way of handling settings.php which means y'all will have to copy the example for local into a new file in order to run this locally.
- There are changes to .gitignore, please review
- Should we move composer requirements all to `require` and remove from `require-dev` besides the ones actually just needed in dev?

### DIFF
--- a/cronish.sh
+++ b/cronish.sh
@@ -1,0 +1,8 @@
+#!/bin/bash 
+set -euo pipefail
+
+while true
+do
+  drush --root=$HOME/web core:cron
+  sleep 15m
+done


greg-does-weather (approved):
LGTM!

PR #247, opened 2023-10-04
161 e2v1 styles fonts

claire-skies-noaa:
## What's wrong?
Site is using default USWDS fonts. We need the following font-family: "Nunito Sans","Open Sans","Helvetica Neue", sans-serif

## How does this PR fix it? 

- Removes default fonts: merriweather, sans-source, etc.
- Adds Nunito variable and static font
- Sets `sans` **font-type** as Nunito stack (font-family: "Nunito Sans","Open Sans","Helvetica Neue", sans-serif)
- Sets heading and body **font-roles** to `sans`, code to `mono`
- Sets a couple other theme settings that override uswds-core values: Sets image path location, turns on compile warnings, turns off notifications.

before PR:
image

after PR:
image

## Types of changes
 🆕 New feature (non-breaking change which adds functionality)

## Checklist:


greg-does-weather (approved):
I don't know what's up with all the whitespace-only changes in the ADRs, but it doesn't affect the content so 🤷‍♂️.

I can confirm that when I switch to this branch, the main body font family switches to Nanito.

PR #255, opened 2023-10-05
For demo purposes, swtich to Nashville

greg-does-weather:
## What's wrong?

The location we display in the current conditions is static for now. It currently points to Minneapolis. We want to switch it to Nashville.

## How does this PR fix it? 

- Changes the coordinates from downtown Minneapolis to downtown Nashville
- Modifies the Twig template to make wind speed/direction in the narrative text conditional, based on whether there is any wind.

image

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -123,9 +123,11 @@ class WeatherDataService {
   public function getCurrentConditions() {
     date_default_timezone_set('America/New_York');
 
-    // Roughly Minneapolis.
-    $lat = 44.98;
-    $lon = -93.27;
+    // 36°09′44″N 86°46′28″W
+
+    // Roughly Nashville.
+    $lat = 36.16;
+    $lon = -86.77;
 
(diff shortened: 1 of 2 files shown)


greg-does-weather:
The "Nashville-Davidson metropolitan government (balance)" text is coming from the API. I think that's why I used Minneapolis instead.

colinmurphy01:
Oooo the title is an interesting use case with the title. Are there any alternatives for this? 

Sidebar: what does (balance) mean?

greg-does-weather:
I have no idea what the "(balance)" thing is about. I did see that the city and county have a unified city/county government, so that part makes sense even if it's not really what people think about when they say "Nashville."

Re: alternatives, I don't know. I tried a few lat/long variations in the near vicinity of Nashville when I was doing this originally, and it would switch from this long name to, like, an airport. I couldn't get it to just return the city. This may be something to discuss with NCO about how the API is getting those names so we can decide if that works for our needs or if we need to do something else. "Something else" in this scenario is probably not fun.

colinmurphy01 (approved):
Despite the weird title, let's ship this for now and then we'll create a ticket for investigating the Location title issue.

colinmurphy01:
Tracking title issue in #258

PR #268, opened 2023-10-10
Add phpunit tests

greg-does-weather:
## What's wrong?

We are not running automated tests in our CI/CD processes.

## How does this PR fix it? 

Adds unit tests to our CI/CD. We don't have any tests yet (and only very little PHP-testable code), so I added a test class for one of the generated classes. It just asserts that `1 = 1`, but it serves as a placeholder for what tests might look like.

Also fixes a bug in the phpcs step, where the output of phpcs was supposed to create GitHub annotations. I'd forgotten to copy over the annotation problem matcher when I switched us from [guix77/phpcs-drupal-action](https://github.com/guix77/phpcs-drupal-action).

And finally, I think this fixes a bug with how stuff was being cached. This new cacheing might be more aggressive, but that could mean saving on execution times.

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -100,6 +100,7 @@
         "drupal/coder": "^8.3",
         "drupal/devel": "^5.1.2",
         "drupal/single_content_sync": "^1.3",
-        "drush/drush": "^12.1"
+        "drush/drush": "^12.1",
+        "phpunit/phpunit": "^9.6"
     }
 }
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -44,7 +44,7 @@ jobs:
       - name: run phpcs
         run: |
           echo "::add-matcher::${{ github.workspace }}/.github/workflows/phpcs-problem-matcher.json"
-          vendor/bin/phpcs -v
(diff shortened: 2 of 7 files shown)


composer.json
```diff
@@ -100,6 +100,7 @@
         "drupal/coder": "^8.3",
         "drupal/devel": "^5.1.2",
         "drupal/single_content_sync": "^1.3",
-        "drush/drush": "^12.1"
+        "drush/drush": "^12.1",
+        "phpunit/phpunit": "^9.6"
```
loganmcdonald-noaa, line 104: Is there a point to using --dev at this point? I think it's useful to have things like phpunit and coder not added in production but not sure the flag is still the recommended way of excluding them?

loganmcdonald-noaa (approved):
LGTM just one question about our requirements.

composer.json
greg-does-weather: Separating them in composer is still recommended, but you don't use the `--no-dev` flag when installing anymore. Instead you set a `COMPOSER_NO_DEV` environment variable: https://getcomposer.org/doc/03-cli.md#composer-no-dev

PR #270, opened 2023-10-10
Report code coverage

greg-does-weather:
## What's wrong?

We have committed to 90% test coverage, but we're not currently capturing or reporting it.

## How does this PR fix it? 

This PR updates the PHP unit test step to capture code coverage, and then adds two additional steps to report coverage and enforce a minimum percentage.

Currently our coverage is at 0%, so we'll soon need to dedicate some time to tests while the amount of code we have is still low.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   php-lint:
-    name: Drupal code standards lint
+    name: PHP lint
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


greg-does-weather:
@loganmcdonald-noaa ☝️ I propose we ship this with 0% coverage. I was having trouble getting the coverage thing to run locally so this report is the only way I have of seeing how we're doing.

loganmcdonald-noaa (approved):
Ok to ship with 0%

PR #275, opened 2023-10-11
Ignore more composer files.

claire-skies-noaa:
## What's wrong?
When running `composer install` or `composer require`, there are additional drupal files loads. Those should come from the composer package.

## How does this PR fix it?
Adds those files to our .gitignore.

## Types of changes

## Checklist:


greg-does-weather (approved):
Okay, this explains why those files keep showing up in my git log and I keep having to delete them. 😅

PR #278, opened 2023-10-12
Add Cypress

greg-does-weather:
## What's wrong?

We don't have end-to-end, in-browser testing.

## How does this PR fix it? 

This PR adds Cypress for local and CI/CD e2e browser testing. It adds a very simple test for time localization across 4 different locales (American English, American Spanish, Mexican Spanish, and Puerto Rican Spanish). We can build out real tests from here.

It also fixes two bugs in `main`. One was uncovered explicitly by the tests and the other was uncovered serendipitously:
- ~The `get_noaa_icon()` function the weather data module was defined twice. It was causing a PHP error. This was introduced in #211 and went unnoticed because we didn't have any tests that actually attempt to run our PHP code. I removed the older version of the function.~
   - retroactively fixed in #212
- There was a typo in two of the weather icon filenames. We got lucky finding these – the weather conditions reported by the API for Nashville just happened to be conditions that triggered the buggy icon filenames, and then the Twig template threw an error because the files didn't exist. I fixed the filenames.

Most of the files in the `cypress/` directory were generated. Some of them are placeholders or examples. They may end up just being cruft, but I opted to leave them for now.

## Types of changes

## Checklist:


.github/workflows/code-standards.yaml
```diff
@@ -133,3 +133,27 @@ jobs:
...
+
+      # Give the containers a moment to settle.
+      - name: wait a tick
+        run: sleep 10
+
+      - name: populate the site
+        run: |
+          cp web/sites/example.settings.dev.php web/sites/settings.dev.php
+          make install-site
+          docker compose exec drupal drush content:import scs-export/node-page-427c45cf-9fb6-4e79-a0f5-5a54f82b0bfb.yml
```
loganmcdonald-noaa, line 156: Will this yaml file name change?

loganmcdonald-noaa:
Should we add some documentation about how to run these tests locally and what we are using for the tests? Like a `docs/testing.md` file with information background on cypress, how to run tests locally, how to run or view them in github actions. I image we will add accessibility testing info there as well and maybe specific how cypress will be used to create things like users in the future.

.github/workflows/code-standards.yaml
greg-does-weather: Yes. After #277 is merged, I'd want to come back and remove line 156 here, as `make install-site` will handle loading content in a more generic way.

greg-does-weather:
Yes we definitely should! I'll create an issue for that. #277 also begins creating developer documentation, so it'll fit nicely in there!

loganmcdonald-noaa (approved):

Issue #281, opened 2023-10-12
Developer documentation: testing


greg-does-weather:
Our developer documentation should describe how to run tests locally, what we're using for those tests, etc. (See https://github.com/weather-gov/weather.gov/pull/278#issuecomment-1760180464).

Blocked by #277

greg-does-weather:
Done in #328

Issue #253, opened 2023-10-05
Clean up and organize developer documentation


greg-does-weather:
Currently, documentation is kind of haphazard. It's not really a problem today, but as our documentation grows, the lack of organization will make it harder to find what you need. We will create structure for our documentation and begin filling it in.

PR #295, opened 2023-10-16
Mgwalker/cypress a11y

greg-does-weather:
## Types of changes

## Checklist:


loganmcdonald-noaa (approved):
Well done! Only question is if we need to add anything to documentation around this?

greg-does-weather:
Yes, we should definitely add documentation. I'm about to start writing a bunch of docs for #253 and #281, and I'll cover it there. Hopefully after that it can be about keeping documentation updated instead of creating it.

Issue #198, opened 2023-09-20
Address inline styles in the current conditions weather block


greg-does-weather:
Once we have our USWDS theme in place, then we can start to figure out how we want to go about applying styling to various components. For now, I just put them right in the Twig template. That's probably not where they should live forever, though, so let's remember to go clean that up.

Introduced in #196.

greg-does-weather:
We could use some design help on this one! In [ADR 0007](/weather-gov/weather.gov/blob/main/docs/architecture/decisions/0007-we-will-use-utility-first-approach-to-css.md), we decided to use a utility-first approach to our CSS.

@claire-skies-noaa or @janelx – We need to find a mapping between these font sizes and their USWDS utility counterparts, any adjustments we need to make to the USWDS theme settings, or any new CSS classes we need to create.

Font-sizes:
- 0.7 rem
- 3 rem
- 1.5 rem
- 0.75 rem

I think we'll need a pair of new classes for the icons here, though. One for the current condition icon (64px tall, black fill) and one for the wind and humidity icons (16px x 16px).

What do y'all think?

janelx:
I can do the suggested mapping w/ @claire-skies-noaa to review.

greg-does-weather:
I've made an initial implementation in #313. Here are the mappings I used:

|rem|USWDS utility|
|---|---|
|0.7|`.font-body-3xs`|
|3|`.font-body-3xl`|
|1.5|`.font-body-lg`|
|0.75|`.font-body-3xs`|

PR #298, opened 2023-10-16
Fix current conditions timestamp and "feels like" text color

### LINKED ISSUE #199: Fix the gray color in the current conditions block theme
Introduced in #196. This can be fixed after the USWDS theme is in place.

greg-does-weather:
## What's wrong?

The text color of the timestamp and "feels like" in the current conditions component are wrong. It's too light and doesn't have enough contrast.

## How does this PR fix it? 

Changes the text color to `text-gray-50`. Closes #199.

## Types of changes

## Checklist:

### DIFF
--- a/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
+++ b/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
@@ -7,7 +7,6 @@
       }
 
       & .timestamp {
-        color: #7F7F7F;
         font-size: 0.7rem;
       }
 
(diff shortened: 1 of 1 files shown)


greg-does-weather:
Coverage is addressed in #282. This PR doesn't add or change unit-testable code, so coverage is not relevant here.

claire-skies-noaa (approved):
**Approved** with a note. Great using utility class for color. Can you do the same for the text size? 

[`font-body-3xs`](https://designsystem.digital.gov/utilities/font-size-and-family/) is the smallest we should use for readable text.

Should not prevent merging this PR.

greg-does-weather:
Eggselent, thank you! That note is super helpful for #198!

PR #303, opened 2023-10-17
Fix ADR 0007

greg-does-weather:
## What's wrong?

The automation for creating ADRs seems to strip out code blocks. OH NO!

## How does this PR fix it? 

It doesn't. But it fixes one of the ADRs that got bitten. As far as I can tell, this is the only one that's been affected.


claire-skies-noaa:
@greg-does-weather Is this going to happen for future Markdown files with code blocks?

claire-skies-noaa (approved):

mgwalker:
@claire-skies-noaa Probably so. I created #304 so we get it fixed, and I'm going to dig into it right now.

PR #313, opened 2023-10-17
Remove inline styles from the current conditions component

### LINKED ISSUE #198: Address inline styles in the current conditions weather block

greg-does-weather:
## What's wrong?

When it was originally created, the styles for the current conditions component were hardcoded right into the component code. Instead, it should use USWDS utility classes.

## How does this PR fix it? 

- Swaps out font size styles for USWDS utility classes
- Swaps out margins for utility classes
- Pulls the styling that's unique to the component into its own sass file

This component is probably going to need some more styling work in order to be *right*, but this is a step in the right direction.

image

Closes #198.

## Types of changes

## Checklist:


web/themes/new_weather_theme/assets/sass/components/_current-conditions.scss
```diff
@@ -0,0 +1,16 @@
+.weather-gov-current-conditions {
+  & .icon svg {
+    height: 64px;
```
claire-skies-noaa, line 3:
```suggestion
    @include u-height(8);
```

web/themes/new_weather_theme/assets/sass/components/_current-conditions.scss
```diff
@@ -0,0 +1,16 @@
...
+
+    & path.st0 {
+      fill: black;
+    }
+  }
+
+  & .item {
+    & svg {
+      height: 16px;
```
claire-skies-noaa, line 12:
```suggestion
      @include u-height(2);
```

web/themes/new_weather_theme/assets/sass/components/_current-conditions.scss
```diff
@@ -0,0 +1,16 @@
...
+      width: 16px;
```
claire-skies-noaa, line 13:
```suggestion
      @include u-height(2);
```

web/themes/new_weather_theme/assets/sass/components/_current-conditions.scss
```diff
@@ -0,0 +1,16 @@
+.weather-gov-current-conditions {
+  & .icon svg {
+    height: 64px;
+
+    & path.st0 {
+      fill: black;
```
claire-skies-noaa, line 6:
```suggestion
      fill: color('black');
```

claire-skies-noaa (changes requested):
This is great, Greg! Thanks for creating the component sass. Let's start using USWDS utility mixins instead of hardcoding values. I've included code suggestions. As long you add 
`@use "uswds-core" as *;` at the top of the SASS partial, you'll have access to all utility mixins and tokens. 

References:

[Height and width mixins](https://designsystem.digital.gov/utilities/height-and-width/#utility-mixins)
[The color tokens and mixins](https://designsystem.digital.gov/design-tokens/color/theme-tokens/#using-theme-color-tokens-2)

greg-does-weather:
@claire-skies-noaa Thank you! I'm going to have to get into the habit of keeping the USWDS reference open more often.

PR #315, opened 2023-10-17
Update .stylelintrc.js

claire-skies-noaa:
## What's wrong?
When Stylelint checks sass files, it throws an error on any classes with a double underscore (__). This is because the default settings is expecting kebab-case classes only. We're using USWDS, which uses a combo of double underscores and hyphens (aka kebabs). 

## How does this PR fix it? 
Adds a rule to stylelintrc.js to ignore selector class patterns, since we should only be using USWDS utility classes. Reference: https://github.com/WordPress/gutenberg/issues/28616

error:
image

## Types of changes

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
   extends: "stylelint-config-standard-scss",
+	"rules": {
+		"selector-class-pattern": null,
+  }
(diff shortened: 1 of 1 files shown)


greg-does-weather (approved):
Yay, thank you for fixing this!

PR #317, opened 2023-10-18
Integrate location

greg-does-weather:
## What does this PR do? 🛠️

- Adds a prototype location search box. Note that this UX component is entirely prototype. It's just here to demonstrate the capabilities in the system.
- Adds a new page at `/prototype-search-location` that exposes the search box
- Adds route handlers for location endpoints:
   - `/point/{latitude}/{longitude}` - This route will attempt to resolve a geographic point into a WFO grid cell. If it succeeds, it will redirect the user to a grid URL. If it fails, it will return a 404 error.
   - `/local/{wfo}/{gridX}/{gridY}/{locationName}` - This route doesn't add any user-facing functionality, but it provides route parameters and a route ID that other components can use to populate themselves. Like the current conditions block!
- Updates the current conditions block to only display on the `/local/{wfo}/...` route
- Updates the current conditions block to pass its routes into the WeatherDataService when it requests data
- Updates the WeatherDataService to take a route when providing data, and to use that route to determine what location to fetch data for (using the WFO, grid X, and grid Y parameters).
- Updates our configuration to display the location search prototype, and to make the current conditions block available on all routes that start with `/local/`. The block will handle hiding itself on `/local/*` routes that don't have grid information.

I've found this to be pretty fragile, but I think it's mostly because I kept forgetting to update some configuration. Hopefully I captured everything in this PR (zapping results in a functional system for me), but if you run into any weird behavior, feel free to ping me and we can troubleshoot.

@claire-skies-noaa If you'd like to pair to review it, that would be great. I'd love to be able to talk through it and make sure the approach makes sense and is Drupal-ish. If this is a sensible way to do it, I can see how we use this basic framework for all other location-related stuff that needs to happen.

https://github.com/weather-gov/weather.gov/assets/142943695/2ad353a4-7cad-4705-a83a-3d9364821296


greg-does-weather:
No review necessary at this point, but a heads-up and an invitation for feedback!
@janelx @kmranjo @colinmurphy01

claire-skies-noaa (approved):
Thanks for the walkthru Greg! 🚀

PR #329, opened 2023-10-23
Make the current conditions component translation-ready

### LINKED ISSUE #204: The text on the current conditions component (e3.v1) is not translation-ready
Introduced in #196.

We cannot address this bug until we build out the infrastructure for supporting translations. I think that should be pretty straightforward with Drupal and its built-in language modules. Once we know how we're going to do translations in general, we can come back and fix this component.

On this one in particular, the displayed text is combined from different sources, so that may introduce additional complexity. We'll evaluate once we get there, but noting here because we may need a little design help.

greg-does-weather:
## What does this PR do? 🛠️

- In the Twig template, runs translatable strings through the `t` filter.
- In the WeatherDataService, sends the current conditions text (e.g., "mostly sunny") through the translation manager
   - I did this translation in the service because it is data from the API. Doing the translation in Twig would be funky because the current conditions text is a variable, so it would have to be double-translated, and that seemed dirty:
   ```php
   {{ "The weather in @place is {{ $data["condition"] | t }}" | t }}
   ``` 
  Also, ☝️ that doesn't work.
- Updates tests to inject a mock translation manager.

Closes #204

None, I guess. We don't actually have translations setup, so we can't be sure that this works as expected yet. But that'll come. I reckon for now we're just making sure the code makes sense.

### DIFF
--- a/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
+++ b/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
@@ -7,7 +7,8 @@
     <h1>{{ content["#data"]["location"] }}</h1>
 
     <p class="margin-0 font-body-3xs text-gray-50">
-      Weather as of
+      {{ 'Weather as of' | t }}
+      {# Datetimes are localized in client-side Javascript. #}
       <weather-timestamp data-utc="{{ content["#data"]["timestamp"]["utc"] }}">
         {{ content["#data"]["timestamp"]["formatted"] }}
       </weather-timestamp>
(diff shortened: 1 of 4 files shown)


web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
```diff
@@ -33,25 +34,40 @@
...
 
     <p class="conditions-narrative long">
-      The weather in <strong>{{ content["#data"]["location"] }}</strong> is
-      <strong>{{ content["#data"]["conditions"]["long"] | lower }}</strong>.
-      Temperature is <strong>{{ content["#data"]["temperature"] }}
-      &deg;F</strong>. Humidity is <strong>
-      {{ content["#data"]["humidity"] }}%</strong>{% if content["#data"]["wind"]["speed"] > 0  %} with
-      <strong>{{ content["#data"]["wind"]["speed"] }}mph winds</strong> from
-      the {{ content["#data"]["wind"]["direction"] }}{% endif %}.
+      {{ "The weather in <strong>@place</strong> is <strong>@conditions</strong>.
```
claire-skies-noaa, line 52: 😍

claire-skies-noaa (approved):

PR #331, opened 2023-10-23
Fix the current conditions component layout

### LINKED ISSUE #330: The grid layout on the current conditions component is wonky as heck
### Description

- The utility icons on the right are sized really weirdly
- The alignment of pieces is also funky
   - The icon should be vertically centered relative to the temperature and conditions text
   - The temperature units should be vertically top-aligned

image

This issue replaces #207 

### Acceptance criteria

- ~Tests have been implemented or modified to capture the bug~
    N/A because we're not currently doing snapshot testing. Layout bugs aren't functional, so we don't have a way of capturing them right now.

greg-does-weather:
## What does this PR do? 🛠️

The current conditions component was kind of a layout mess. It was *close*, but not quite right. This PR essentially re-implements this Twig template to work better. It uses grids + flexboxes to make all the pieces align nicely.

Closes #330

* @claire-skies-noaa I only used utility classes. As a result, the elements have a lot of classes attached to them. I wasn't sure if this was the right way to go with utility-first, or if I should created a couple of classes and used mixins instead. If the latter, it should be easy for me to make that change, just let me know!
* @janelx I think with this one, we should be able to close issue #175. What do you think?
* The humidity and wind icons had heights and widths set in the SVG file, which resulted in them not scaling properly. I removed the height and width from those files so we can control them via CSS.

Before:
before

---

After:
after


claire-skies-noaa (approved):
## Code Review Checklist

### The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation


## Reliability

## Accessibility
    - [x] Tab order is logical

## Device Matrix
- [x] firefox (renders correctly and user interactions work)
    - [x] at 480px
    - [x] at 640px
    - [x] at 1024px

janelx:
@greg-does-weather would you like me to do a design review on the built component? Is there a link to where I should review? (apologies if I'm missing it in this PR)

greg-does-weather:
Here's a good one for review: [https://beta.weather.gov/local/FFC/51/87/Atlanta](https://beta.weather.gov/local/MPX/108/72/Minneapolis)

There's a [bug on some pages](https://beta.weather.gov/local/MPX/108/72/Minneapolis) where the component isn't full width. I didn't create an issue for it yet, but I should, and it'll be something we can address properly when we put the whole local page together.

PR #334, opened 2023-10-24
Bug fix: catch when the icon is empty or null

### LINKED ISSUE #333: The site fails if the API returns null or an empty string for the icon for an observation
### Description

In the weather data service, we map from the API's icon into a new icon and a text description of the current conditions. We currently do not handle the case where the API returns null or an empty string. This failure cascades down and crashes the whole site.

Example API result: https://api.weather.gov/stations/KBNA/observations/2023-10-24T11:53:00+00:00
This link might not work forever.

### Acceptance criteria

- [x] Tests have been implemented or modified to capture the bug
- [x] Changes have been tested for accessibility

greg-does-weather:
## What does this PR do? 🛠️

- Adds tests to cover the cases where the API returns null or an empty string for the icon of an observation
- Updates the weather data service to properly handle those cases.

Closes #333.

This one's pretty straightforward, I think.

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -23,7 +23,7 @@ function get_api_condition_key($observation) {
    */
   $icon = $observation->icon;
 
-  if($icon == NULL or strlen($icon) == 0) {
+  if ($icon == NULL or strlen($icon) == 0) {
     return "no data";
   }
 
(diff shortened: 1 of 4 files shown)


loganmcdonald-noaa (approved):
Thanks for the test!

greg-does-weather:
Heck yeah! In the event of a bug, write a test to catch it before fixing it, and then you shouldn't have that specific bug again. I don't remember who first told me that eons ago, but boy has it saved me a lot of trouble over the years.

PR #365, opened 2023-10-26
Skip tests when only documentation changes have been made

### LINKED ISSUE #357: Exempt the documentation directory from code standards tests
### Description

When we're editing the documentation, there's no reason to run our full suite of code standards tests. We should modify them so that if only the `/docs` directory is modified, they all pass immediately.

> [!WARNING]  
> We can't just use the GitHub workflow filter to disable the runs altogether because they are required to pass by our branch protection rules.

### Acceptance criteria

- [ ] For pull requests that only change the `/docs` directory, tests run very quickly because they pass and don't actually test anything.

greg-does-weather:
## What does this PR do? 🛠️

Updates our automated tests to pass quickly if only documentation has been changed.

Closes #357

Whew. What a hassle.

### DIFF
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -4,9 +4,61 @@ on:
   merge_group:
 
 jobs:
+  should-test:
+    name: check if we should run tests
+    runs-on: ubuntu-latest
+    outputs:
+      "no": ${{ steps.skip-tests.outputs.no }}
+      "yes": ${{ steps.skip-tests.outputs.yes }}
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: list files that have changed from main
+        id: diff
+        run: echo "diff=$(git diff --name-only origin/main)" | tr "\n" "\t" >> $GITHUB_OUTPUT
+
+      - name: determine whether to skip tests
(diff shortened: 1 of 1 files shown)


.github/workflows/code-standards.yaml
```diff
@@ -23,32 +75,28 @@ jobs:
...
-      - name: cache composer bin
-        uses: actions/cache@v3
-        id: bin-cache
-        with:
-          path: ${{ steps.composer-paths.outputs.bin-dir }}
-          key: composer-bin-${{ hashFiles('composer.lock') }}
-
       - name: install dependencies
-        if: ${{ steps.bin-cache.outputs.cache-hit }} != 'true'
+        if: needs.should-test.outputs.yes == 'true'
```
greg-does-weather, line 85: It turns out that cacheing this bin directory doesn't actually work properly, and coupled with these new conditionals, it failed very consistently. So... yank it. The composer intermediary cache is maintained, and that one works and gives us all the speed up benefits anyway.

greg-does-weather (commented):
For most of these test jobs, I only disabled the actual test to keep the configuration a little simpler. For the ones that start a Docker container (e2e and a11y), I disabled all the steps because otherwise the do-nothing tests would still take several minutes.

loganmcdonald-noaa (approved):

PR #395, opened 2023-11-13
Ca/177 e3v2 short term forecasts hourly

claire-skies-noaa:
## What does this PR do? 🛠️

Creates a new block called **Hourly forecast** with placeholders for short term data.

## What does the reviewer need to know? 🤔

Connection to actual API data and HourlyForecastBlock.php.test will be done in another branch.

**Steps:**

1. Run `make import-content` to import Hourly Forecast block settings.
2. Run `make cc` to clear cache.
3. Confirm **Hourly forecast** block is enabled and placed on [homepage](http://localhost:8080/admin/modules) in your browser
4. Confirm layout on mobile, tablet, and desktop breakpoints per #177 

## Screenshots (if appropriate): 📸
See #177

Issue #177, opened 2023-09-06
e3.v2 short term forecasts (hourly)


janelx:
Product flow (in figma): [From weather.gov](https://www.figma.com/file/2bDju7teT0H71ldhCkYORF/Weather.gov-2.0-MVP-Flow-%2F-Share-doc?type=whiteboard&node-id=1-197&t=4fI7ZiRfDaM0r8hF-4)

### Supported user stories
- 3.3 … I want to view short ~~and longer~~ term weather forecasts 
  - long term forecast story is here #178 

### Story details: As a member of the public 
- I am able to understand what the basic weather will be for each hour for the next 4 hours

### Design (define/discovery)
- Completed during prototype phase

### Support needed
- Weather API (forecastHourly, forecastGridData)
- Weather icons
- Speech to text for summaries (need to discuss between PD, eng, and PM if part of this issue)

### Acceptance criteria
[Criteria outline in figma](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=345-2238&mode=design)
**As a member of the public I can easily understand:**
- [ ] what the weather will be for each hour for the next 4 hours for a specific location
**Additional UX research/continuous discovery**
- [ ] Qualitative experience of accessibility:
  - [ ] Screenreader experience of listening to the weather components
  - [ ] Experience of navigating weather data for screenreader, select, keyboard shortcuts
- [ ] Are we showing the most accurate and useful data?
  - [ ] For is it accurate/useful information to show PoP in the contents of the weather icon? API response examples.
- [ ] Is it easy to understand? Is the understanding accurate (enough to make a decision)?

----
### General
- [ ] Responsive
- [ ] USWDS friendly
- [ ] Section 508 compliant

### Short term forecasts >= tablet  - 12 col grid

<kbd><img width="583" src="https://github.com/weather-gov/weather.gov/assets/93934808/7801252c-a78a-4616-a1eb-9b80d29a387b.png"></kbd>

**Hour title (12 PM, 1 PM, etc)**
- [ ] h3
- [ ] hourly beginning with next hour. Ex. If it is 11:01 to 11:59 show for 12 PM, 1 PM, 2 PM and 3PM (four hours)
- [ ] time zone is relative to the forecast location not browser location

**Conditions (icon and description)**
- [ ] NOAA weather icons - max height: 64px or max width: 72px
- [ ] Icon description from Icons api
- [ ] text: 'body'

[...]

PR #412, opened 2023-11-14
Make the max_items configuration for hourly forecast settable in the Drupal UI

greg-does-weather:
## What does this PR do? 🛠️

Adding a first coat of polish to #395, this PR makes the `max_items` configuration something that admins can set in the Drupal UI instead of having to bonk around in code. This configuration is then pulled at build time and pushed down to the template.

1. View the site. Note that there are 4 hours of forecast displayed.
2. [Log in](http://localhost:8080/user/login) as an admin
3. Navigate to the [hourly forecast block configuration](http://localhost:8080/admin/structure/block/manage/new_weather_theme_hourlyforecastblock)
  - Or click to it. It's under Home > Administration > Structure > Block layout, then find the Hourly Forecast block, and click the "Configure" button 
5. Note that there is a form element that says "*Maximum items to display*" and it is set to 4
6. Change this value to another number and click the "Save block" button at the bottom of the page
7. View the site again. Note that there are now a different number of hours of forecast display.

### DIFF
--- a/web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
+++ b/web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
@@ -3,6 +3,7 @@
 namespace Drupal\weather_blocks\Plugin\Block;
 
 use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Form\FormStateInterface;
 
 /**
  * Provides a block of the hourly (short term) weather conditions.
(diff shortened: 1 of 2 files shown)


greg-does-weather:
There's not a ticket for this, but two things happened that made me want to pursue it *right now*:
1. Claire's note in the comments was super valuable in helping me connect the dots to how a custom block can present configuration fields in the Drupal UI.
2. While I was eating lunch, I poked at documentation and saw how to do it. Rather than saving off links to the documentation, I figured it'd make more sense to just do it and then we'd have our own reference in the future.

eric-gade (commented):
This works for me!
Just two related and I think non-blocking questions:
1. For this block, do we want to impose a "max" restriction?
2. We will want to deal with styling issues wherever someone puts in a number that is greater than 8 (see the screenshot below)
  
<details>
<summary>
Check out this screenshot with the config set to 50
</summary>
Screen Shot 2023-11-14 at 5 26 55 PM
</details>

greg-does-weather:
Yes to both of those things! I think they're both design questions, and it looks like in #177 the designed-for number of hours is four. So it's entirely possible that we don't actually need to make this configurable at all, in which case I guess this PR could be closed and we'd just have it as a reference, which is fine too.

@janelx What do you think? Should the number of hours to display in the hourly forecast be configurable? And if so, what guardrails should we put around it to ensure someone doesn't accidentally set it to 0 or 50 (or 100!)

janelx:
OMG yes! that's amazing. let's bring to PO. I think we can add a horizontal scroll for the time being. Show 7 and scroll bar?

greg-does-weather:
@janelx Do we need to make considerations around whether we're showing this in the short-term forecast versus longer-term? (I'm showing some ignorance here! I'm not even sure if hourly data is present on longer-term forecasts.)

greg-does-weather:
In the meantime, I'm going to change what my dev deployment is pointing to, but we can change it back to this when we're ready to look at it some more.

greg-does-weather:
Given ongoing discussion in Slack, I think it might make sense to prioritize this one over #437, and then we can fiddle with it as much as we want without having to redeploy anything. If we later decide we don't want this to be configurable, we can take it back out then. 🤷‍♂️

eric-gade (approved):

PR #419, opened 2023-11-15
Fix page layout vertical flexing

eric-gade:
## What does this PR do? 🛠️
We update the main page layout templates so that the `<main>` element will flex to fill space vertically, even if it contains little content. The upshot is that the footer is always at the bottom of the document regardless of window size.
  
The canonical case of "doesn't have a lot of content" is, for the moment, a 404 page.

- Navigate to a page that doesn't exist in the live beta site ([example](https://beta.weather.gov/doesnt-exist/)). Expand your window vertically if needed, and notice how the footer is not at the bottom.
- With this branch running locally, navigate to a page that doesn't exist ([example](http://localhost:8080/doesnt-exist/)). Notice how the main area flexes vertically to fill in the space, and that the footer is at the bottom of the page.

<details>
<summary>Before:</summary>

Screen Shot 2023-11-15 at 12 53 42 PM
</details>
  
<details>
<summary>After:</summary>
  
Screen Shot 2023-11-15 at 12 53 42 PM
</details>


web/themes/new_weather_theme/templates/layout/off-canvas-page-wrapper.html.twig
```diff
@@ -0,0 +1,26 @@
...
+ * - children: Contains the child elements of the page.
+ *
+ * @ingroup themeable
+ */
+#}
+{% if children %}
+  <div class="dialog-off-canvas-main-canvas height-full" data-off-canvas-main-canvas>
+    {{ children }}
+  </div>
+{% endif %}
```
greg-does-weather, line 26: Ooh, what's this file?

greg-does-weather (commented):
This works as expected on 404 pages, but on the front page and on location pages, the footer ends up on the right instead of the bottom?

<details>
<summary>screenshot</summary>

image

</details>

eric-gade:
> This works as expected on 404 pages, but on the front page and on location pages, the footer ends up on the right instead of the bottom?
> screenshot

Oof. Good catch, forgot to unset wrapping. Should be fixed in https://github.com/weather-gov/weather.gov/pull/419/commits/64ef27f327ffed25d7e11fcbfdaf110455c873fb

web/themes/new_weather_theme/templates/layout/off-canvas-page-wrapper.html.twig
eric-gade: As far as I can tell, Drupal offers (modal?) dialog capabilities, and this `div` wraps the main page content so that the dialog can distinguish it. 
  
Because I've been using USWDS [utility classes](https://designsystem.digital.gov/utilities/) where possible rather than editing a stylesheet, I needed to be able to edit the attributes of this div. It turns out [this file](https://api.drupal.org/api/drupal/core%21modules%21system%21templates%21off-canvas-page-wrapper.html.twig/10) is part of the standard theme structure for Drupal, and by adding a custom version to our theme I'm simply overriding the default. Btw, the default is nearly identical to this file, with the exception of the `height-full` class on the div. Perhaps there's a better way to accomplish the same thing...

greg-does-weather (approved):

web/themes/new_weather_theme/templates/layout/off-canvas-page-wrapper.html.twig
mgwalker: > Perhaps there's a better way to accomplish the same thing...

Claire once told me that when working with Drupal, the way that works is the right way. So... there you go. :joy:

PR #422, opened 2023-11-15
Hook up the hourly forecast to location and real data

greg-does-weather:
## What does this PR do? 🛠️

When we implemented the hourly forecast, we hardcoded the data. This PR hooks up the hourly forecast block to real location-based data.

Most of the new code is tests. You'll also need to update your config with `make import-config`.

<details>
<summary>screenshot</summary>

image

</details>


eric-gade (approved):
Definitely going to crib some of this going forward. Working and LGTM 👍

PR #423, opened 2023-11-16
Disable internal page cache

### LINKED ISSUE #370: Caching broken for location prototype
### Description

Forecasts are cached for a specific point location and don't get updated for some time resulting in out of date forecasts in weather.app.cloud.gov

greg-does-weather:
## What does this PR do? 🛠️

Removes the Drupal Internal Page Cache. This module aggressively caches pages sent to anonymous users, which doesn't work for us because our pages are highly dynamic. [Drupal's documentation](https://www.drupal.org/docs/administering-a-drupal-site/internal-page-cache) recommends disabling the module in this case, and it fixes the caching problem we've seen with the current conditions component.

Closes #370.

My [dev environment](https://weathergov-greg.app.cloud.gov/local/MPX/108/72/Minneapolis) has been running this branch for about 24 hours. You can see that the last updated time is within about an hour (consistent with what the API returns, and what appears on the existing point forecast), and I can attest that the temperature and conditions are accurate relative to what I see out my window. So I guess... I'm saying check out the docs, ask any questions you have, and if there's nothing evidently wrong, we might just have to accept this and watch to see how it behaves in staging because I don't know how else we can test/prove this.

### DIFF
--- a/web/config/core.extension.yml
+++ b/web/config/core.extension.yml
@@ -32,7 +32,6 @@ module:
   mysql: 0
   node: 0
   options: 0
-  page_cache: 0
   path: 0
   path_alias: 0
   s3fs: 0


eric-gade (approved):
lgtm!

PR #427, opened 2023-11-16
Wait for USWDS if necessary

### LINKED ISSUE #425: The location search box sometimes doesn't do anything
### Description

When you type in the location search box, sometimes nothing happens.

### Acceptance criteria

- Typing in the location search box should always do something.

greg-does-weather:
## What does this PR do? 🛠️

We're using the USWDS combobox component. Under the hood, it inserts an `<input>` element into the DOM for us, but we can't be sure when that will happen. #425 is caused basically by a race condition. This PR adds some logic to our search box to essentially ask the DOM periodically if the input element is present yet and once it shows up, then it hooks up the event listeners.

The new logic only does the wait-and-see thing 20 times before giving up entirely. The reason for this is that if there's anything outside of our control that interferes with the element being added to the DOM, we don't want to just keep running a poll forever and ever and chewing up users' CPUs and/or batteries.

Closes #425.

### DIFF
--- a/web/themes/new_weather_theme/assets/js/components/locationSearch.js
+++ b/web/themes/new_weather_theme/assets/js/components/locationSearch.js
@@ -28,11 +28,33 @@ class LocationSearch {
   #select;
 
   constructor(node) {
-    this.#input = node.querySelector("input");
-    this.#select = node.querySelector("select");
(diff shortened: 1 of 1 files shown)


greg-does-weather:
Closing in favor of #428

PR #428, opened 2023-11-17
Updating to use event bubble / document event handlers

### LINKED ISSUE #425: The location search box sometimes doesn't do anything

### LINKED ISSUE #426: The location search box has usability and accessibility deficiencies
### Description

The location search box currently uses a [USWDS combo box](https://designsystem.digital.gov/components/combo-box/). The usability and accessibility guidance sections say:

> **Avoid auto-submission.** Don't use JavaScript to automatically submit the form (or do anything else) when an option is selected. Offer a "submit" button at the end of the form instead. Users often change their choices multiple times. Auto-submission is also less accessible.

> **Avoid auto-submission.** Don’t use JavaScript to automatically submit the form (or do anything else) when an option is selected. Auto-submission disrupts screen readers because they select each option as they read them.

Our current behavior submits on select, which is exactly what USWDS discourages. USWDS actually recommends using a different component entirely for the time being.

**The location search box was introduced as a prototype.** The whole search flow still needs to be designed, and this box is just a placeholder to allow us to get to other functionality (like location-specific current conditions, forecast, etc.). That said, so long as we're putting this on the internet for people to see, we ought to make it accessible.

### Acceptance criteria

- The search box no longer submits when a user selects an item.
- The search includes a submit button users can use to trigger a submission instead.

eric-gade:
## What does this PR do? 🛠️
We use the event bubble to capture input/change events on the location input field.
  
This is in direct reference to #425 

- Spin up the branch locally
- Open the browser console
- Enter a zip or location into the location search input and expect results
- Check the console for errors
- Refresh the page, repeat several times

### DIFF
--- a/web/themes/new_weather_theme/assets/js/components/locationSearch.js
+++ b/web/themes/new_weather_theme/assets/js/components/locationSearch.js
@@ -22,84 +22,80 @@ const createOneTimeJSONP = () => {
   return { load, name };
 };
 
-class LocationSearch {
-  #input;
+const locationSelected = async (event) => {
 
-  #select;
-
-  constructor(node) {
-    this.#input = node.querySelector("input");
-    this.#select = node.querySelector("select");
-
-    this.#input.addEventListener("input", this.searchChanged.bind(this));
-    this.#select.addEventListener("change", this.locationSelected.bind(this));
+  // We only care about change events that have
+  // come from select elements with the "weather location"
+  // name.
+  // Note that we cannot use the id here, because USWDS strips it,
+  // giving the input element it inserts that id value instead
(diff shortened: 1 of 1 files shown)


eric-gade:
@greg-does-weather question: do we want to bind the listeners to the `form` rather than the document? Simple enough change if so

greg-does-weather:
Since this component is a prototype and we're going to throw it our or completely rebuild it anyway, I'm personally not too worried about it right now. This is a good solution!

greg-does-weather (approved):
***And*** you got rid of that annoying class! 🤩

PR #430, opened 2023-11-18
Put weather stories on the location page

greg-does-weather:


greg-does-weather:
Closing in favor of #439

PR #432, opened 2023-11-21
Switch to WOFF2 fonts

greg-does-weather:
## What does this PR do? 🛠️

USWDS assumes fonts configured via its settings to be available in WOFF2 format. Our fonts are currently in TTF. This PR replaces the TTF versions of our fonts with WOFF2. This should fix all the 404 errors on page load, and it should result in us using the font we've been designing for.

This should be the easiest one to test ever because it shouldn't be impacted by Drupal caches at all, and there's no content or configuration, so no need to do anything to try it out - just check it out and verify that whatever your current configuration still works.


eric-gade (approved):
lgtm 👍

PR #433, opened 2023-11-21
remove blank selectElement

elijah-wright:
this line removes the input if a location is selected and the user selects outside of the input


greg-does-weather:
Hi @elijah-wright! Thanks for your contribution! This is a good catch. We'll review it soon.

greg-does-weather:
Hi @elijah-wright. Can you tell us more about the behavior you were seeing that this PR fixes?

I should also let you know, the whole search box and that whole experience has not been through design. I put it in quick-and-dirty just so we had a way for people to get to location pages. We're probably going to rebuild it entirely later on. But that said, it's still helpful to us to know what the problem was so we can make sure we don't cause the same problem again. 😄

elijah-wright:
I put in a location but didn't enter it, and then I clicked off the search box. because the location was over three characters long and the input changed, the search box cleared

greg-does-weather:
Ah-ha, now I understand.

It turns out this PR does something else! This search component is kind of a mess so it's hard to see what's going on. We're using [USWDS](https://designsystem.digital.gov/), and it controls the search box. We are using the USWDS [combo box](https://designsystem.digital.gov/components/combo-box/) in a way they did not intend for it to be used, so we kind of hacked it together. Anyway, USWDS controls the behavior of that text box; our code only reads the value of the box, it does not it. Instead, USWDS is clearing it whenever you click into unless there is a selected value.

The code you changed on line 80 definitely looks suspicious, though. It turns out, that line empties the search suggestions so we can put new suggestions in it. If we don't empty them, the list will just grow forever because new suggestions will be added to the bottom.

elijah-wright:
interesting, reading over it it does seem like it would do that. I tested the component with that line removed and it seemed to work fine?

with what you said, do you want me to see what I can do to not get the component to clear?

greg-does-weather:
Certainly feel free to continue if you're interested in trying to solve it, but since we almost certainly will not use this USWDS combo box for long, it would mostly just be for fun. (Unless you find a good solution, in which case I'd be up for helping you get it submitted to USWDS as a permanent fix.)

I'm going to close this PR so it's not on our project board, but feel free to keep pushing to it and using this as a place to document what you're finding. And thanks again for dropping in to help!

PR #439, opened 2023-11-22
Location page

greg-does-weather:
# Hold for daily forecast.

It'll be much easier to move the daily forecast into this PR than vice versa, I think.

## What does this PR do? 🛠️

Updates the page displayed at `/local/*` addresses.

#### Configuration

- Added the beta banner as a theme region, so we can manage it with the CMS. If there is no associated beta banner block content, the banner will not be displayed.
- Added location, current conditions, forecast, and WFO local content as theme regions.
- Added a beta banner block type
- Added a view for weather stories; this view filters to the most recent published weather story for a WFO based on the URL; it presents itself as a block that can be placed in the block layout
- Added a view for WFO blurb (attribution?); same as above
- Modified the WFO field used in weather stories, etc., to include the 3-letter codes so we can query by it
- Modified the WFO address field so it has a max of 500 characters instead of a max of 20 characters
- Update the block layout to add the beta, location, current conditions, and forecast regions, and put the appropriate blocks in each
   - **Beta banner region:** beta banner block
   - **Location region:** location search block
      - We should rename this block; more info in the backend section
   - **Current conditions regions:** current conditions block
   - **Forecast region:** hourly forecast block (daily forecast will go here too)
   - **WFO local content:** weather story view and WFO blurb (attribution?) view

#### Backend

- Updated the `WeatherDataService` module to cache its results for a given request, so we can request data from it in multiple places without having to back to the API each time.
- Updated the `LocationSearchBlock` to query the `WeatherDataService` for the location currently being viewed

#### Frontend

- Updated the template for the location search block to display the current location, if available, along with an as-yet non-functional link to change location
   - For now, the "link" is just text, and the search box is visible on all pages
   - Because this block isn't just for location searching anymore, I think it would make sense to rename it to just a location block, or something like that.
- Added a new page template for URLs that start with `/local/*`.
   - This seemed like the easiest way to get a different layout for a different part of the site. However, there's now a fair bit of duplication between the default page and the local page. I think we could handle that with some Twig imports.
- Added styling for the new regions and blocks
- Updated styling of the existing blocks to make them more responsive

- closes #248 
- closes #310 
- closes #311
- closes #404
- closes #429
- implements #181

## Screenshots (if appropriate): 📸

<details>
<summary>desktop view</summary>

Screenshot 2023-11-24 at 09-11-42 Weather gov National Weather Service

</details>

<details>
<summary>mobile view</summary>

Screenshot 2023-11-24 at 09-12-39 Weather gov National Weather Service

</details>

PR #440, opened 2023-11-24
Create a weather block base class

greg-does-weather:
# Hold for #439

This PR builds on #439, so after that one is merged, this one should be rebased on top of `main` and then it will be ready for review.

## What does this PR do? 🛠️

Inspired by @eric-gade, this PR pulls the common boilerplate code out of our custom module blocks and puts it into a base class. It also configures autoloading of our blocks so we can reference them by namespace instead of having to `include()` them in tests.

Behaviorally, nothing should change. If everything works, then I guess this is good?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,11 @@
         },
         "sort-packages": true
     },
+    "autoload": {
+        "psr-4": {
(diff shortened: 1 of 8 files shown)


greg-does-weather:
Superseded by #448

PR #442, opened 2023-11-24
Document high-level data flow

greg-does-weather:
## What does this PR do? 🛠️

Adds a high-level data flow diagram. This was living in my head and I thought it should be written down somewhere. Once I got it down, it seemed simpler than I thought it was, but... maybe we should keep it? 🤷‍♂️


warrickmoran:
Thank you!

PR #447, opened 2023-11-27
remove the beta site notice and leave just search

loganmcdonald-noaa:
## What does this PR do? 🛠️
This removes the content body where it says "This is a beta site" and leaves just the search block. We already have the banner, do we need this text?


greg-does-weather (approved):
Does what it says. Maybe a product question whether to merge it, but whether it's *fit* for merge... yes it is.

PR #464, opened 2023-11-29
Add screenshots to our end-to-end tests

greg-does-weather:
# Hold

I don't think this is useful unless we can integrate the newly-captured screenshots into our review process. Ideally I think we'd want the screenshots to be added to a comment that gets updated on each push (instead of new comments each time). GitHub doesn't provide an API for uploading images into comments, so we'll have to dance around that a little bit. There are existing actions that do it, but they add new comments each time, so that'd be noisy.

## What does this PR do? 🛠️

Adds an additional test to our end-to-end tests that iterates over a list of pages we want to screen shot and gets screen shots of all of those pages at phone, table, and desktop viewport widths.

Nothing much. Running end-to-end tests will create screenshots (in the git-ignored) `tests/e2e/cypress/screenshots/screenshots.cy.js/` directory.

### DIFF
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -262,6 +262,13 @@ jobs:
           project: tests/e2e
           cache-key: cypress-e2e-${{ hashFiles('package-lock.json') }}
 
+      - name: save screenshots
+        if: needs.should-test.outputs.yes == 'true'
+        uses: actions/upload-artifact@v3
+        with:
+          name: screenshots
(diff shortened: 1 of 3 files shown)


greg-does-weather:
It took me a long time to come back around to this, but agreed. It'd be neat to add them to PRs, but the GitHub API doesn't seem to really want you to, so... okay then. At least we'll have them locally!

eric-gade (approved):
Print it! (No checklist because we've had this one for a while)

Issue #455, opened 2023-11-28
Update daily and hourly forecasts to use API short descriptions


greg-does-weather:
### Description

Currently, both components use a mapping from API icons to condition descriptions. This is, from the words of NCO, a "no-go." (For a lot of reasons, not least that those icons will never give us as nuanced results as actual text will.) We should instead use the short description provided by the API.

Because the text descriptions from the API may be significantly longer than the ones in our mapping, we may need to also pair up with design to tweak things accordingly.

> [!NOTE]  
> We know that the API returns different short forecast text than the legacy point forecast page does. This is a concern, but Brian is tracking that and we can come back and adjust when we know what we need to do, if anything.

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

colinmurphy01:
Can be picked up after #439 is merged.

greg-does-weather:
There's a draft PR #472 for this. I don't remember why I put it into draft rather than marking it ready for review – I think maybe there was still some discussion about whether this is really what we wanted to do?

colinmurphy01:
From a PO perspective, @shadkeene were pretty clear that we should leverage shortForecast. (also validated by Brian)

greg-does-weather:
Well then! I'll put that PR in review.

PR #472, opened 2023-11-30
Use API text for conditions descriptions

greg-does-weather:
## What does this PR do? 🛠️

We currently key off the icon returned by the API to map into a condition description. However, (we think) we should be using the text from the API. This PR updates the current conditions, hourly forecast, and daily forecast components to use the text provided by the API.

Addresses #455 

A simple cache clear should have this ready to run locally.

I think it would be good to do a design/eng huddle to look at this together. It would also still be super helpful to have a complete and definitive list of text values the API may return so we know what kind of text lengths we need to accommodate.

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.getHourlyForecast.php.test
+++ b/web/modules/weather_data/src/Service/WeatherDataService.getHourlyForecast.php.test
@@ -177,7 +177,8 @@ final class WeatherDataServiceGetHourlyForecastTest extends TestCase {
 
     $expected = [
       [
-        "conditions" => "Thunderbolt", // Conditions should be sentence-cased
+        // Conditions should be sentence-cased.
+        "conditions" => "Thunderbolt",
         "icon" => "nodata.svg",
         "probabilityOfPrecipitation" => 11,
         "time" => "12 PM",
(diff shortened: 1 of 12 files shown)


web/modules/weather_data/src/Service/WeatherDataService.getHourlyForecast.php.test
```diff
@@ -227,14 +236,14 @@ final class WeatherDataServiceGetHourlyForecastTest extends TestCase {
 
     $expected = [
       [
-        "conditions" => "Rain",
+        "conditions" => "Frightening",
```
eric-gade, line 239: 👍 👶

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -438,10 +426,8 @@ public function getDailyForecastFromGrid($wfo, $gridX, $gridY, $now = FALSE, $de
       // Get any mapped condition and/or icon values.
       $obsKey = $this->getApiObservationKey($daytime);
 
-      // The short forecast name should be mapped to
-      // the legacyMapping and translated.
-      $shortForecast = $this->legacyMapping->$obsKey->conditions;
-      $shortForecast = $this->removeParenthetical($shortForecast);
+      // Sentence-case the forecast description.
+      $shortForecast = ucfirst(strtolower($daytime->shortForecast));
```
eric-gade, line 430: Good call

web/modules/weather_data/src/Service/legacyMapping.json
eric-gade, line 1: Do we actually want to get rid of this information though? Maybe we can just rename the field "iconDescription" or something like that, since it's actual information from the icon API (not sure if we'll ever need it though...)

eric-gade (approved):
Looks good 👍 . I was going to look into some styling additions but we should wait for a design pass

web/modules/weather_data/src/Service/legacyMapping.json
greg-does-weather: I tend towards the "if you're not using it, take it out" philosophy, which I guess is like one far end of YAGNI. And if we do need it, it's in our git history so we can get it back.

PR #479, opened 2023-12-01
Refactor: Have a single method for fetching API data and cache it

greg-does-weather:
## What does this PR do? 🛠️

- Adds a public `getFromWeatherAPI` method to the weather data service, and modifies the other methods to use it
- The new method caches data from any given URL for the duration of the response
- Removes independent caching from `getCurrentConditions`
- Adds tests for the new method, and removes caching test for `getCurrentConditions`

Main this probably to disable whitespace in the review, because `getCurrentConditions` gets unindented, undoing (but generalizing!) a change from #439.

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -52,6 +52,13 @@ class WeatherDataService {
   private $currentConditions;
 
   /**
+   * Cache of API calls for this request.
+   *
+   * @var apiCache
+   */
+  private $apiCache;
+
+  /**
    * Constructor.
    */
   public function __construct(ClientInterface $httpClient, TranslationInterface $t) {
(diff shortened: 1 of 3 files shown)


web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -62,13 +69,34 @@ public function __construct(ClientInterface $httpClient, TranslationInterface $t
...
   }
 
+  /**
+   * Get data from the weather API.
+   *
+   * The results for any given URL are cached for the duration of the current
+   * response. The cache is not persisted across responses.
+   *
+   * Disable phpcs on the next line because it does not like method names with
+   * sequential uppercase characters, but... I'm not camel-casing "API".
```
eric-gade, line 88: Maybe we can disable this rule globally for phpcs because I dont like camel casing acronyms either!

Issue #474, opened 2023-11-30
The place name on the location page comes from the URL


greg-does-weather:
### Description

The place name displayed on the location page is taken from the URL. This is undesirable because you could make it whatever you wanted and we'd just... show it:

[Nashville's page with modified place name](https://beta.weather.gov/local/OHX/50/57/Not%20Memphis)

We should get the place name from... somewhere else.

### Acceptance criteria

- [ ] We have determined a canonical source of place names (probably the gridpoint API?)
- [ ] We've updated the weather data service to fetch the place name instead of pulling it from the URL

PR #481, opened 2023-12-01
Refactor: Make blocks location-aware

greg-does-weather:
## What does this PR do? 🛠️

Instead of pushing route controllers into the data service and having it figure out the location to work with, have blocks do that. This PR adds a `getLocation` method to the block base class that we can build on in the future to allow more different ways of specifying location. It also updates the weather data service `getCurrentConditions` method to `getCurrentConditionsFromGrid` to accept the WFO grid as arguments instead of deriving it from the route.

I will do the other weather data service methods in a follow-on PR.


web/modules/weather_blocks/src/Plugin/Block/WeatherBlockBase.php
```diff
@@ -67,4 +67,31 @@ public function getCacheMaxAge() {
     return 0;
   }
 
+  /**
+   * Get the current location that the block should be associated with.
+   *
+   * If the current route is on the grid, pulls the WFO grid information from
+   * the URL.
+   */
```
eric-gade, line 75: Leave or toss, but I was thinking some pointers in the comment might help us keep track of the structure of this location object as we modify it going forward:
```suggestion
  /**
   * Get the current location that the block should be associated with.
   *
   * If the current route is on the grid, pulls the WFO grid information from
   * the URL.
   * @returns array $location An array with location info in the following
   * structure:
   *     array['grid'] An NWS API grid response
   *.                       ['wfo'] The NWS WFO code for the location
   *                       ['gridX'] The X value of the grid coordinate
   *                       ['gridY'] The Y value of the grid coordinate
   */
```

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -268,7 +250,7 @@ public function getCurrentConditions($route) {
         'feels_like' => (int) round($feelsLike),
         'humidity' => (int) round($obs->relativeHumidity->value ?? 0),
         'icon' => $this->legacyMapping->$obsKey->icon,
-        'location' => $location,
+        'location' => "SOMEWHERE - FIX THIS SOON",
```
eric-gade, line 253: I assume related to https://github.com/weather-gov/weather.gov/issues/474

eric-gade (approved):
Checks out on my end! 👍 
  
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

web/modules/weather_blocks/src/Plugin/Block/WeatherBlockBase.php
greg-does-weather: Good call! I was also wondering if we should create some explicit types for these sorts of things (this location, current conditions, hourly forecast, etc.). That might be overkill, though, when we can just document them at the points they're created and they're each only created in one place. 🤔

PR #499, opened 2023-12-05
Restructure tests

greg-does-weather:
## What does this PR do? 🛠️

- moves our PHP tests into a subfolder to be a bit tidier
- creates a test base class to handle setting up the object instead of having it repeated all over
- shortens the test class names to not be redundant, since they're namespaced

Big de-dupe. Also the git diff got confused about `GetFromApi` being renamed to `Base`, but in fact `Base` is an entirely new file, so that makes reviewing it kind of wonky.


eric-gade (approved):
👍

PR #500, opened 2023-12-05
Remove test coverage reporter

greg-does-weather:
## What does this PR do? 🛠️

Disables the test coverage reporter step of our code standards automated tests. This step is the one responsible for the extra comment on all our pull requests with the funky coverage histogram, which is a really odd way of presenting that data. It's not very useful and it creates extra notifications. Since we already have a step for enforcing a 90% lower-bound on code coverage, we don't really need this.


eric-gade (approved):
🙏

PR #504, opened 2023-12-05
No longer import content on deployments

greg-does-weather:
## What does this PR do? 🛠️

The content we store in our repo is mostly just to get a representative site stood up quickly. Now that our deployed environments are up, we no longer need to import this content into them.

### DIFF
--- a/scripts/post-deploy.sh
+++ b/scripts/post-deploy.sh
@@ -22,11 +22,5 @@ drush list | grep "config:import" > /dev/null || install_drupal
 echo  "Updating drupal ... "
 drush state:set system.maintenance_mode 1 -y
 drush deploy -y
-# TODO: Remove and only pull content down
-# Temporarily importing content that was created locally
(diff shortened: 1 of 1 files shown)


loganmcdonald-noaa (approved):
Can we please add to the documentation what the process for content changes will be?

greg-does-weather:
Tagging in @kmranjo for thoughts as well. I *think* the process for pushing content from local/dev/staging to beta will manual, using the single-content-sync plugin. If someone creates content that needs to be migrated up, they can export and reimport it since they'll have the knowledge and context of what should and shouldn't be synced.

For sync down from beta to staging/dev/local, maybe periodically someone runs a content export from beta, imports it into staging, and puts it in git so it's available for dev environments. This flow feels like it would be safe to automate, provided that users and stuff like that don't get synced, which I believe they do *not*.

greg-does-weather:
Adding #518 to make sure we document the process.

Issue #516, opened 2023-12-07
V1 Location Page - Accessibility Considerations


colinmurphy01:
### Mac VoiceOver + Chrome
- [ ] Beta Banner: Full website url is read aloud (consider hyperlinking weather.gov)
- [ ] Header: Logo + Weather.gov both have links, duplicative?
- [ ] Search: Redundant "change location" AND "search for location" AND "search for location in search box"
- [ ] Current conditions: doesn’t  have a H2 header
- [ ] Current conditions: Consider skipping to “weather as of [Date][Time] + [Summary]”
- [ ] All Icons don’t have alt text
- [ ] Hourly forecast: navigates by each individual component (4PM, image, mostly cloudy, 2% chance, 46, ºF) should we group these so they're read together by hour?
- [ ] Hourly forecast: I’d like to be able to exit out of Hourly forecast easily and then go to Daily Forecast (without having to tab by component) - wrap in a div? (this might exist in the VO tool / be a possible human error)
- [ ] Hourly forecast: Temperature and ºF are not grouped to be read together
- [ ] Daily forecast: Repetition temperature reads “day will be high of “54” degrees 54º night will be low of 48ºF”

### Chrome Wave Extension
- [ ] No alt text in gov banner for flag
- [ ] Footer links are grouped in multiple <nav> elements with aria labels?

greg-does-weather:
It does make sense! I have some thoughts on a few of the items you found, but generally it's a good list to work from.

### Mac VoiceOver + Chrome
- Header: Logo + Weather.gov both have links, duplicative?
  - Ideally they'd be part of the same link, but I don't think we can structure it that way in the HTML. Whether or not there should be two links... I don't really have an opinion. Easy change, either way!
  - Incidentally, I noticed that below the "weather.gov" text is a the next "National Weather Service" and that links to noaa.gov, which seems wrong. 😛 We probably need to change either the text or the link so they match.
- Search: Redundant "change location" AND "search for location" AND "search for location in search box"
  - Yes. I think this should be noted in the search design ticket #185. We can fiddle with it, but it doesn't seem like a good use of time right now to me. There's also a little bit of VO funkiness at play with the "search for location" and "search for location in search box" stuff. It's reading the label when you land on it, then when it gets to the input, it reads the associated label again. I believe this is normal and expected behavior, but we should definitely look into it.
- Current conditions: Consider skipping to “weather as of [Date][Time] + [Summary]”
[...]

PR #531, opened 2023-12-13
SVG Icon Partials

eric-gade:
## What does this PR do? 🛠️
This PR implements conversion of SVG icon files to twig template partials that can be used with parameters. This allows us to individually add CSS classes and ARIA attributes as needed each time we use an icon.
  
This work is in reference to comments from #516 (as well as long standing discussions we've had)

### SVG to template generation ###
We have added a [script](https://github.com/weather-gov/weather.gov/blob/eg-516-icon-svg-partials/scripts/svg-to-twig.js) and a [corresponding npm task](https://github.com/weather-gov/weather.gov/blob/eg-516-icon-svg-partials/package.json#L43) that will take all relevant SVG icon files in our theme and generate twig templates for them in a [directory we use for storing these partials](https://github.com/weather-gov/weather.gov/tree/eg-516-icon-svg-partials/web/themes/new_weather_theme/templates/icons).
  
The purpose of the script is to:
1. Remove any specific XML markup to make this more HTML compliant;
2. Insert twig template values for incoming variables in parts of the SVG element (aria-label and class, specifically)
3. Add template comments
4. Create a corresponding twig template for every icon SVG currently checked into the repo
  
### Using the templates ###
The template partials take two arguments when they are included in another template:
| argument | description |
|----------|-------------|
| label | An aria label string that will be used by the svg |
| classNames | A string containing any additional class attribute values that should be applied to the svg element |
  
### Block adjustment ###
Our WeatherBlocks `build` methods will now need to add another key/val to the object they respond with: `iconTemplate`, which is simply the name of the template file for the given icon. This amounts to swapping the `.svg.` for `.html.twig` suffix.
  
### Current usage ###
For this PR, we have updated the DailyForecast block and its template to use these new partials. However, it's unclear that any of these icons actually need a separate aria label, since the surrounding context probably provides enough information. Either way, **we no longer have descriptions of the icons** available for use and might need to restore these in the `legacyMapping.json` file. TBD.


greg-does-weather:
Oooh, this is cool!

eric-gade:
Closed in favor of #543

PR #536, opened 2023-12-13
Hourly/Daily forecast blocks using lists

eric-gade:
## What does this PR do? 🛠️
Structuring the markup of daily and hourly forecast cards as lists. This makes more sense semantically and might be better from an accessibility perspective.
  
This is in relation to the corresponding item in #516 

No screenshots, because they appear identically as before.


greg-does-weather (approved):
The semantic improvements are 👍 and will be good for accessibility for sure. It still reads each piece separately so you have to keep navigating to get from "5pm" to "sunny" and so on, but having them identified as a list of items is an improvement!

PR #539, opened 2023-12-14
Enable Twig debugging and disable markup caching in dev

greg-does-weather:
## What does this PR do? 🛠️

Enabling Twig debugging causes it to add HTML comments to the markdown that tell you what templates it is searching for along the way, which is super helpful for figuring out what a new template should be named.

Disabling the markup cache means that any change to a template file will be rendered immediately so you do not need to manually clear the cache. You'll still need to clear the cache for module changes.

`make import-config` will get this change configured. Thereafter, `zap` and `reset-site` will also do it. These two settings are ***NOT*** included in exported configurations, so there's no risk of accidentally pushing them into production.

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ export-content: ## Export all content to web/scs-export
 
 import-config: ## Import the Drupal configuration from the config directory into your site
 	docker compose exec drupal drush config:import -y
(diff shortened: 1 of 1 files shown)


eric-gade (approved):
This is excellent!

Issue #521, opened 2023-12-11
Find an alternative to embedding SVG icons


greg-does-weather:
For weather icons, we currently embed whole SVGs. The reason we do that is so we can control the color of them with CSS. The downside, however, is that they increase the size of the HTTP response. If the SVG was not embedded but was instead a separate resource, the separate resource could be cached, resulting in a better user experience.

There are several options available, including (but not limited to):

1. Modify the SVG files themselves to be the colors we want. The downside is that we can't control the color with CSS, which means if we want multiple colors of icons, we'll need multiple files.
2. Use the SVGs as a [background mask](https://github.com/weather-gov/weather.gov/issues/516#issuecomment-1847896218). Unsure of the accessibility implications of this approach, however.

greg-does-weather:
Closed by #543

PR #543, opened 2023-12-14
Use SVG spritesheet for our icons

eric-gade:
## What does this PR do? 🛠️
**NOTE: This PR relates to #531 and closes that PR without merging in its changes**  

We have decided to follow the [pattern recommended by USWDS](https://designsystem.digital.gov/components/icon/#accessibility-icon), using SVG partial symbols and selectors to display icons. This gives us more control over the styling and accessibility qualities whenever we need.
  
This PR implements the following:
- A script that will take all icon SVG files, modify them slightly, and generate a single `spritesheet.svg` file.
- Updates hourly and daily blocks to return the plain filename (without extension) of each icon
- Updates hourly and daily block templates to use the `<use xlink:href>` style selection of a single icon from the spritesheet, along with necessary styling to make them appear as they already were

To run the spritesheet compiler:
```shell
npm run compile-svg
```


greg-does-weather:
This will close #521 too.

scripts/compile-svg-sprite.js
```diff
@@ -0,0 +1,115 @@
...
+ * @return string A string with extraneous
+ * newlines removed
+ */
+const cleanupWhitespace = (str) => {
+  const lines = str.split("\n");
+  return lines.map(line => {
+    return line.replace(/^\w+\n$/, "");
+  }).filter(line => {
+    return line !== "";
+  }).join("\n");
```
greg-does-weather, line 56: I believe this multiline regex should work as well:

```suggestion
return str.replace(/\n+/gm, '\n');
```

scripts/compile-svg-sprite.js
```diff
@@ -0,0 +1,115 @@
...
+ * Removes the XML header from the string
+ *
+ * @param str string The source xml string
+ * @return string The cleaned output xml string
+ */
+const removeXMLDeclaration = str => {
+  const lines = str.split("\n");
+  return lines.filter(line => {
+    return !line.startsWith("<?xml");
+  }).join("\n");
```
greg-does-weather, line 69: I'm partial to regular expressions, but I'll defer to you on readability. I think what you have here is probably more readable, tbh.

```suggestion
 return str.replace(/\n<?xml[^\n]*/gm, "");
```

greg-does-weather (commented):
This is clever. I like it a lot! But you're going to hate me.

When I run it locally, I can't actually see the icons because they have a white fill. And... it turns out... you can't target stroke and fill on SVG sprites with CSS. 😬

Instead of setting the fill color, we *could* apply a brightness filter:

```css
.icon svg {
  filter: brightness(0);
}
```

That starts to fill icky, though. Or at least less obvious than `fill: black;`. I'm sorry, I feel like I sent you down this path and now it's a bad path.

scripts/compile-svg-sprite.js
eric-gade: I'll mess with the regex. There are places where the SVGs were generated (cough cough, hello Illustrator) with lines that have only spaces and/or tabs

greg-does-weather: Oooh, I missed that nuance.

partly-igor:
> When I run it locally, I can't actually see the icons because they have a white fill. And... it turns out... you can't target stroke and fill on SVG sprites with CSS. 😬

Hopefully I am not overstepping here, and I might be missing something, but it seems like it's possible in general to target with CSS

image

mgwalker:
You're definitely not overstepping, and you're also correct. I think the problem I was seeing was elsewhere and it resulted in the style I was setting just not being applied. But my "not possible" statement was wrong.

eric-gade:
@partly-igor to bring you up to speed: it looks like they don't accept the styling in FF. There's all kind of chatter online about what to do but none of it seems to work when I try it...

eric-gade:
Alright @greg-does-weather cracked the case! Because we are still loading individual SVGs elsethere on the page, and these SVGs have style tags that define the `st0` class, that style is available globally. Our spritesheet was removing style tags but not class attributes that referenced `st0` (and the devtools were not forthcoming about where this was happening).
  
The solution, pushed in https://github.com/weather-gov/weather.gov/pull/543/commits/5ff33c324b04efdb85cfcdcc74fba74d022e4017 , was to simply remove all class attributes from any element in the spritesheet.

greg-does-weather (approved):

PR #545, opened 2023-12-15
Add documentation about how to export and import content

greg-does-weather:
## What does this PR do? 🛠️

Documents the process of importing and exporting content.

### DIFF
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -1,27 +1,46 @@
 # Environments
 
-There are two tiers of environments - unstable and stable. 
-
-## Unstable
-
-### Local/Development
-
-See [dev docs](dev/) for information on running the application locally. We use docker with composer to manage the application locally.
-
-### Developer Sandboxes/QA
-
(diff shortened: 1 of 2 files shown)


greg-does-weather:
@kmranjo We can pair up on reviewing this if it's helpful!

kmranjo:
@greg-does-weather  Yes, could we set up some time? Both to review/discuss and also because I'm not entirely sure what to do with a pull review request (I dont think I've done one yet in github?).

greg-does-weather:
@kmranjo Based on our conversation yesterday, I made an attempt at plain-languaging our environments documentation, moved the warning about configuration changes to the top of the sync doc, and removed the rest of the configuration documentation since it's probably not applicable to the target audience.

kmranjo (approved):
I think this looks good! We may want to consider spelling out more explicitly that content doesn't have to be created in staging first and then pushed to production, but only in certain circumstances. I think that might be something that can go in a separate doc more about content management as a whole though.

Issue #517, opened 2023-12-07
Anchorage location not working on beta.weather.gov


klsheets:
### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

greg-does-weather:
The API is not giving us grid information for Anchorage (about [latitude 61.218, longitude -149.858](https://api.weather.gov/points/61.218,-149.858)). I remember from the PA that Alaska is "different" in a lot of ways; is this one of those? Does Alaska not use grids?

But legacy forecast is happy to [provide a forecast for those coordinates](https://forecast.weather.gov/MapClick.php?lon=-149.858&lat=61.218), so we know the data exists and it's probably available in the API, we just need to figure out how to get it.

klsheets:
Alaska issues grids, but it’s different grids than the conusAK grids are polar stereo graphic projection and I think 1.25km, or at least some of them.Pacific region grids are also different.  I think there’s 1 fit Hawaii and 1 for the western pacific territories.  All use the Mercator projection and I’ve lost track of their standard resolution my guess is 2.5 or 1.25kmFor records sake.  Conus grids are lambert conformal projection and depending on element 1.35 or 2.5km I think.It’s been 15 years but the grids were part of my master’s project so I have lots of detailed grid attributes buried somewhere in my brain.Sent from my iPhoneOn Dec 7, 2023, at 7:44 PM, Greg Walker ***@***.***> wrote:﻿
The API is not giving us grid information for Anchorage (about latitude 61.218, longitude -149.858). I remember from the PA that Alaska is "different" in a lot of ways; is this one of those? Does Alaska not use grids?
But legacy forecast is happy to provide a forecast for those coordinates, so we know the data exists and it's probably available in the API, we just need to figure out how to get it.

colinmurphy01:
@greg-does-weather add it to the API spreadsheet?

greg-does-weather:
Added. The API is also not returning observation stations for points in Alaska.

greg-does-weather:
The API has a known bug specifically for the Anchorage CWA. It's on the API's roadmap, but there's no proposed workaround for us in the meantime. #532 is for a short- to mid-term solution to at least let users know that we're aware of the problem.

colinmurphy01:
Closing this issue:

- Short-term solve with #532 
[...]

PR #546, opened 2023-12-15
If a given latitude and longitude does not produce a grid, show a "not implemented page

### LINKED ISSUE #532: Create static page for Anchorage locations
### Description

Currently, the API is unable to return grid (and therefore conditions/forecast) information for locations in the Anchorage wfo.
  
We should design and implement a static information page to display to users so that they know we currently cannot provide them this information.
  
From an engineering/cms perspective, this is simply a matter of creating a content page of some sort and capturing all requests to `/AFC/*`, redirecting them to this new page.

### Acceptance criteria

- [ ] There are designs and content for a static page describing how the forecast information is unavailable for the Anchorage area
- [ ] We have implemented this page within the CMS
- [ ] Any request to a URL that begins with `/AFC/` (the Anchorage wfo code) should display the static page we've designed and implemented

greg-does-weather:
## What does this PR do? 🛠️

This PR serves as a general purpose stopgap for #517. If any latitude/longitude (taken from search) does not return a grid cell, it redirects to `/not-implemented`. We will need to add content in each environment at that location, but in the meantime it'll at least result in a 404.

It also logs each time this happens so we'll be able to determine if it's happening anywhere else, though that seems unlikely. (Except for the part where a lat/long may be outside of NWS's grids entirely, but we haven't addressed that yet.)

closes #532

https://github.com/weather-gov/weather.gov/assets/142943695/d78ffaf4-f6c4-44e9-bf3f-8b47c0257556

### DIFF
--- a/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
+++ b/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
@@ -61,6 +61,14 @@ final class LocationAndGridRouteController extends ControllerBase {
       // If we don't get a corresponding grid location, throw a 404.
       throw new NotFoundHttpException();
     }
+
+    if ($grid["wfo"] == NULL) {
+      $logger = $this->getLogger("Weather.gov data service");
+      $logger->notice("location has no grid: $lat / $lon");
+
+      return new RedirectResponse("/not-implemented");
+    }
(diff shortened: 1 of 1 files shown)


eric-gade (approved):
Looks good to me 👍

Issue #550, opened 2023-12-18
Move authentication from using the ICAM dev idp to prod idp


loganmcdonald-noaa:
- [x] Request they add us to the prod idp via our metadata
- [x] Change the redirect URLs for login/logout and update the public cert
- [x] Test change is working with the team

Follow up to #348 and #324

PR #555, opened 2023-12-18
update saml http redirect to prod idp

loganmcdonald-noaa:
Lands #550 

Once approved, before merging I will update the cert in all environments using credentials provisioning with our service.


greg-does-weather (approved):
Let's do eeet

Clip of Moira Rose from the TV show Schitt's Creek, standing outside the motel. She looks disappointed and says, "Do what you  must."

PR #556, opened 2023-12-18
Fetch hazards for a location and put them on the page

### LINKED ISSUE #522: Ingesting hazards and identifying hazard type
### Background
In several features, #155 and #156, we are identifying the capability for the CMS to ingest hazard products and then be able to provide complementary information based on the hazard type. 

How could this work?

greg-does-weather:
## What does this PR do? 🛠️

Very roughly puts alerts on the location page. These haven't been styled according to designs, they're just USWDS alerts plopped on the page, to show that we can get them.

The alerts from the API come in a few different configurations, but the biggest issue I think we'll run into in the short term is the description text. In some cases, the text is ***only*** the "what/when/where/impacts" bits. Other times, those bits are missing entirely and the description is some other blurb. I suspect there are cases where the description is both at once. This PR does not attempt to address all of those cases, though it makes a first pass at pulling out the what/when/where/impacts statements if it can.

Closes #522.

```
make cc
make import-config
```

It may also want a `make rebuild`, but that *shouldn't* be necessary. But who knows, the Drupal cache is pretty mystical.

image_480

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherAlertTrait.php
+++ b/web/modules/weather_data/src/Service/WeatherAlertTrait.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Drupal\weather_data\Service;
+
+/**
+ * Add weather alert methods.
+ */
+trait WeatherAlertTrait {
+
+  /**
+   * Get active alerts for a WFO grid cell.
+   */
+  public function getAlertsForGrid($wfo, $x, $y) {
+    $geometry = $this->getGeometryFromGrid($wfo, $x, $y);
+    $point = $geometry[0];
+
+    return $this->getAlertsForLatLon($point->lat, $point->lon);
+  }
+
+  /**
+   * Get active alerts for a latitude and longitude.
+   */
+  public function getAlertsForLatLon($lat, $lon) {
+    $alerts = $this->getFromWeatherAPI("https://api.weather.gov/alerts/active?status=actual&point=$lat,$lon");
+
+    $alerts = array_map(function ($alert) {
+      $output = clone $alert->properties;
+
+      if ($alert->geometry != NULL) {
+        $output->geometry = $alert->geometry->coordinates[0];
+      }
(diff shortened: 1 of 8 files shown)


web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -0,0 +1,62 @@
...
+      }, $www);
+
+      $www = array_filter($www, function ($item) {
+        $expected = ["what", "when", "where", "impacts"];
+        return in_array($item["name"], $expected);
+      });
+
+      $output->description = explode("\n\n", $alert->properties->description);
+
+      $output->whatWhereWhen = $www;
```
eric-gade, line 54: Eventually we will want to pull this out into some other utility, because as you've said, the parsing is going to potentially get hairy

eric-gade (approved):
Code lgtm 👍 
  
FYI with this extra call, we've pushed the location page into the >2s load time territory (I'm averaging 2.4s on most page loads. Just noting that for later discussions.
  
Here is an example of a curious alert that I found, with a no data for current conditions:
  
localhost_8080_local_HNX_84_134_Mammoth%20Lakes (1)

  
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

web/modules/weather_data/src/Service/WeatherAlertTrait.php
greg-does-weather: Agreed. Created #570 to capture this as debt.

greg-does-weather:
Created #571 to track the page loading time issue. We definitely need to address that.

PR #559, opened 2023-12-19
Validate path parameters for our controlled routes

### LINKED ISSUE #530: Sanitize inputs instead of simply trusting them
Currently our location pages accept URL parameters (WFO, grid X, and grid Y) without verifying or cleaning them. We shouldn't do that. Since we only use them to pass along to the API, the result should only be that users get an error page, but we still shouldn't just do it.

greg-does-weather:
## What does this PR do? 🛠️

- Adds route parameter validation to the `/points/{lat}/{lon}` and `/local/{wfo}/{x}/{y}` routes
  - Enforces that latitude and longitude are decimal numbers
  - Enforces that WFO is 3 alphabetic characters
  - Enforces that X and Y are integers
  - Any deviation results in a 404
- Adds WFO uppercasing to all the API calls that use the WFO because the API will give us a 500 if it's not uppercased
- Removes the location name from the end of the local route - it was only put there as a placeholder until we started fetching the name from the API (or possibly eventually somewhere else, but in any case, it's handled now)

Closes #530

`make rebuild && make cc` maybe necessary. I can never tell what makes Drupal accept changes in this file.


eric-gade:
@greg-does-weather When I test this locally, and I put in a junk value (http://localhost:8080/local/HNX/084/134/), I am getting the dreaded "This website encountered an error" page -- is that the intent? Or is there a 404 page we should be redirecting to?

greg-does-weather:
Wow. It's the trailing slash. Uh... well... good catch. Let's see if I can figure out how to handle that. `:lolsob:`

greg-does-weather:
Drupal won't create URLs with trailing slashes. By default, it simply doesn't permit them. There is an (unsupported?) [module](https://www.drupal.org/project/trailing_slash) that will enable it, but not only does it appear to be unsupported, but it also requires you to know exactly which URLs should be accepted with trailing slashes and we cannot know that. (Or if we could enumerate it, we wouldn't want to.)

Another possible solution would be to configure some redirects in Apache. It might be straightforward in cloud.gov (we already have [some Apache configuration](https://github.com/weather-gov/weather.gov/tree/main/.bp-config/httpd)), but I'm not sure how much effort it'd be to handle in other infrastructure.

The more I think about it, the more I think resolving it in the webserver is the right approach, but I'm just a little concerned about how we manage it such that we don't forget to make the same change in other environments.

greg-does-weather:
@shadkeene Wanted to bring you into this one before we merge it because it makes a change that seems like it could be meaningful. As written, this PR would change what URL people get redirected to when they search for a location or use their browser location. Instead of being:

```
beta.weather.gov/local/PQR/104/114/Glide
```

it would not add the location name anymore:

```
beta.weather.gov/local/PQR/104/114
```

The URL part of the change is not necessary, but I originally added location name to the URL only because that was the only way I had handy to get it. Now that we're getting the location name from the API instead, we don't *need* it in the URL anymore, so I took it out as kind of a cleaning chore.

We can still do the validation we need to do without changing the URL behavior, so if you'd rather we keep the location name in the URL, that's a minor change.

eric-gade (commented):
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

loganmcdonald-noaa (approved):

PR #563, opened 2023-12-20
Forecaster can use admin theme

eric-gade:
## What does this PR do? 🛠️
Adds permissions for the Forecaster role, such that the admin theme will load (ie, when editing or creating content, there is appropriate styling and the correct theme loads)

When editing content as a user who is a Forecaster, you should **not** see any unstyled content.


loganmcdonald-noaa (approved):
thanks!

PR #575, opened 2023-12-22
If a user selects a place from the location search, show that place name on the location page

greg-does-weather:
## What does this PR do? 🛠️

Implements an idea @eric-gade had about using a form `POST` to send the place name from location search into our `/point/{lat}/{lon}` endpoint. When we receive it, we stuff it into a short-term cache and continue our normal redirect to a `/local/{wfo}/{x}/{y}` endpoint. When we get the request for the second endpoint, we pull the place name out of cache, if available, and use it. If there's no place information in cache, we fallback to using the API.

Since I was pulling the Drupal cache into the weather data service anyway, I also took the time to update the `getFromWeatherAPI` method to use this cache as well, instead of using its own pseudo, request-scoped cache. The cache now persists for a minute, so if two people happen to ask for the same gridpoint within a minute of each other, they'll get cached results and speed things up quite a lot. That should pay huge dividends for search results of more densely-populated places. (Also @eric-gade's idea.)

At least `make cc` and possibly `make rebuild`.


web/modules/weather_data/src/Service/Test/GetFromAPI.php.test
```diff
@@ -33,25 +33,22 @@ final class GetFromApiTest extends Base
...
 
-        // Change the response and call it again. We should get the same response.
-        $this->httpClientMock->append(
-            new Response(
-                200,
-                ["Content-type" => "application/json"],
-                '{"here":"response"}',
-            ),
-        );
-        $second = $this->weatherDataService->getFromWeatherAPI("my url");
```
greg-does-weather, line 53: Since we're mocking the cache now, we no longer need to actually make this call twice. :yay:

web/themes/new_weather_theme/templates/block/block--weathergov-location-search.html.twig
```diff
@@ -1,7 +1,9 @@
 <div>
 
   {% if content.place %}
-  <h1 class="margin-bottom-0">{{ content.place.city }}, {{ content.place.state }}</h1>
+  <h1 class="margin-bottom-0">
+    {{ content.place.city }}{% if content.place.state %}, {{ content.place.state }}{% endif %}
+  </h1>
```
greg-does-weather, line 6: This is because the `place.state` property may not always be set anymore. For example, if the user selects "Sea World, 500 Sea World Dr, San Diego, CA, 92109, USA", we won't parse that into city/state. We'll just stuff everything into the `city` property. In that case, we don't want the extraneous comma.

eric-gade (approved):
LGTM 👍 
  
The caching of the location page data is a huge improvement. I saw reductions from 2.8s (initial request) to <300ms (subsequent reload).
  
I do think we might want to have a convo about the TTL for the cache values. For example, I think the place name values -- cached by grid & WFO -- could probably live a little longer, but that's not as big of a deal.
  
One thing I'm still wrapping my head around is how the server even allows the POST request, or where it specifically responds to that. Are we simply allowing all types of requests on the lat/lon route? I think we might want to consider 405-ing for anything that isn't a GET or POST.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

greg-does-weather:
Yes, it appears our route controllers are configured to just accept whatever request type. I haven't looked into it, but I'd hazard a guess that the `routing.yml` file in the routes module has properties for specifying what verbs to accept. And I agree, we should probably do that!

Issue #571, opened 2023-12-20
Location page loads take around 2 seconds, right at the upper limit


greg-does-weather:
### Description

Our location page load times are long (around 2 seconds). This is considered to be about the maximum acceptable load time. We need to investigate our sources of latency and find some mitigations because the user experience is suffering. We should also identify a target page load time (Mozilla recommends ~1 second).

### Acceptance criteria

- [ ] Average location page load times are within our target time
- [ ] Tests have been implemented or modified to capture the bug – an e2e test that loads a handful of location pages in succession and averages the load time would be pretty cool

elijah-wright:
I can't test this, but I think the latency issue is waiting on api.weather.gov to provide data back. if there's no other way to get the data quicker, caching may be necessary

colinmurphy01:
@eric-gade is this still an issue. You mentioned that our caching strategy is starting to work.

Should we close this and continue to monitor or are there more substantial solutions we could explore that we could consider?

elijah-wright:
would it be worth it to configure an external cache service (Cloudflare, Varnish) to cache the API with a TTL of 60 seconds? obviously there's apiCache but that's not a persistent cache.

eric-gade:
@colinmurphy01 The caching branch has not been merged in yet ([the PR still open](https://github.com/weather-gov/weather.gov/pull/575) and has some testing errors failing, and we are waiting for @greg-does-weather to come back first)

eric-gade:
Hi @elijah-wright -- Drupal has a built in cache capability that we are going to use initially, it's just that we haven't merged in those changes yet due to holiday lag

PR #576, opened 2023-12-23
add page load speed test

elijah-wright:
## What does this PR do? 🛠️
adds a page load speed test as described in #571

I have not tested this because api.weather.gov is down but it should work

### DIFF
--- a/tests/e2e/cypress/e2e/page-load-speed.cy.js
+++ b/tests/e2e/cypress/e2e/page-load-speed.cy.js
@@ -1,0 +1,25 @@
+describe("load pages", () => {
+    it("is not a test, is just a utility", () => {
+      const pages = [
+        { name: "location page 1", url: "/local/OKX/33/35/Hoboken" },
+        { name: "location page 2", url: "/local/LOX/155/45/Vernon" },
+        { name: "location page 3", url: "/local/LOT/75/73/Chicago" },
+        { name: "location page 4", url: "/local/HGX/65/97/Houston" },
+        { name: "location page 5", url: "/local/PSR/159/58/Guadalupe" },
+      ];
+      let totalTime = 0;
+  
+      for (const { name, url } of pages) {
+        const startTime = performance.now();
+
+        cy.visit(url);
+        cy.wrap(performance.now()).then(endTime => {
+            totalTime += startTime - endTime;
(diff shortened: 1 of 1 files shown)


tests/e2e/cypress/e2e/page-load-speed.cy.js
```diff
@@ -0,0 +1,25 @@
...
+  
+      for (const { name, url } of pages) {
+        const startTime = performance.now();
+
+        cy.visit(url);
+        cy.wrap(performance.now()).then(endTime => {
+            totalTime += startTime - endTime;
+            cy.log(`${name}: page load took ${startTime - endTime} milliseconds.`);
+        })
+      }
```
greg-does-weather, line 20: What makes the timing measurements tricky is that Cypress is asynchronous, but it tries to look synchronous. I guess they wanted it to be easy to think about, and most of the time it is, but when you need to measure times, it can be tough.

So for example, what's happening in this loop is that we take a timestamp, then enqueue a page visit, then take another timestamp, then enqueue a handler for when Cypress finishes the page "visit" using the second timestamp as the end time. What you've done is pretty clever using `wrap()`. Unfortunately, it's capturing a timestamp that happens immediately after the page visit is enqueued. It may or may not cover the actual time needed to visit the page. (Cypress will do it whenever it gets around to it; we don't control it.)

What I was seeing was that the start and end times for all of the pages were very nearly the same, which makes sense because they are being measured inside the loop, and the loop isn't blocking so the reported page load times are too short. Our page loads are longer than the tests are saying.

Something like this will work:

```js
const startTime = performance.now();

cy.visit(url).then(() => {
  const endTime = performance.now();
  totalTime += startTime - endTime;
  cy.log(`${name}: page load took ${startTime - endTime} milliseconds.`);
});
```

The important thing is not to call `performance.now()` until after Cypress has completed the page visit.

greg-does-weather:
Hey @elijah-wright! Thanks for contributing again. You're right, the API was down for a few days around the new year. There was an upstream system that needed to shut down for maintenance and it affected everything downstream of it, including the API.

Anyway, thanks for tackling this issue for us! I was able to get it running locally yesterday and the tests work as advertised. However, there are a couple of quirks that make performance testing harder than it ought to be in Cypress. I opened [another branch](/weather-gov/weather.gov/pull/653) to take inspiration from your work and try to get around the Cypress timing wonkiness.

Issue #475, opened 2023-11-30
Testing page


colinmurphy01:
We should create a page that pulls in and renders data for multiple places so that we can observe and assess the data coming from the API.

Locations (shad to fill out):
- Crater Lake, OR
- Biloxi, MS
- Caribou, ME
- Cheyenne, WY 
- Centralia, WA

A page for each:
- Current conditions (across 3-5 locations)
- Daily forecast
- Hourly Forecast

Goals:
- Team can monitor for API errors or inconsitencies
- Team can view variances in weather conditions and copy length

colinmurphy01:
@greg-does-weather what locations should we include to get a variety of weather conditions at any point in time.

greg-does-weather:
The thing about the weather is that it changes all the time. By which I mean to say, there is no such list of locations. :trollface:

I think what I'd want to do for a test page would actually be to find different conditions and collect them into a set of mock API responses. That way we don't have to worry about the test page changing with the actual weather; it'd be a static set of conditions that we've identified and want to be able to look at. Going that route, we could also just make stuff up to test various scenarios, like one hour has a super short condition description ("snow") and the next one is super long("snow and fog and thunderstorms and maybe frogs"). We don't necessarily have to be reliant on actual results from the API for this page.

colinmurphy01:
@greg-does-weather and @eric-gade I think I saw you toying with this in this sprint?

greg-does-weather:
We're changing our approach. #509 updates our weather blocks so they are location-configurable, but we think the layout builder may be a more appropriate approach for the testing page.

So I think this ticket should now go back into the backlog.

shadkeene:
@greg-does-weather @colinmurphy01 @eric-gade I've added 5 locations that hopefully give a diverse and abundant look at weather hazards through March at least.

colinmurphy01:
Feedback from PO Review yesterday:

- Approved
- Deploy to staging not beta
- Add last updated timestamps for each component (daily and hourly).
  - only viewable on testing pages
- Add navigation between testing pages

PR #581, opened 2024-01-02
Initial Testing Pages

eric-gade:
## What does this PR do? 🛠️
This PR implements the testing pages outlined in #475 
  
We made some additional important changes in order to make these testing pages possible:
* Enabled Layout Builder and exported the config
* Added a new content type called Testing Page just for these pages
* Enabled Layout Builder _only_ on the individual nodes of Testing Page instances
* Created the testing pages using layout builder
* Added custom URLs for each testing page for easier navigation

The config and content have both been exported, but a deployment will only import the config. Exported content must be manually imported in any deployed environment. So once this is fully committed to beta, we will need to also manually import the testing page content there as well.

## Temporary Testing
For now, I have these pages deployed in my sandbox environment. [See this slack post for the links](https://gsa-tts.slack.com/archives/C05DU11D4J2/p1704224519965539).


loganmcdonald-noaa (approved):
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [ ] I ran the code

## Documentation

## Security
- [ ] security false positives are documented
- [ ] data from external sources is cleaned and clearly marked

## Reliability

## Infrastructure
- [ ] all changes are auditable and documented via a script
- [ ] it is clear who can and should run the script

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
- [ ] firefox (renders correctly and user interactions work)
    - [ ] at 480px
    - [ ] at 1024px

loganmcdonald-noaa:
Do we need to add the pages to accessibility testing or skip because they aren't pages that will be used other than by us?

eric-gade:
@loganmcdonald-noaa good question. At first I thought "no need" since it's just us. But the pages are for testing, after all. The only problem is that the content needs to be imported in order for the page to exist, so we'd have to come back around to that issue after the first deployment

colinmurphy01:
Confirming if we deploy this to beta, this just deploys the config and not the content and therefore the testing pages will not be visible in beta?

eric-gade:
@colinmurphy01 that's correct. Content has to be manually imported

PR #593, opened 2024-01-05
Setup the config split module

### LINKED ISSUE #562: Explore using config_split to manage our configuration
config_split allows us to create complete and partial splits for our configuration to use different configuration for different environments. It is probably handy at this point as we differentiate between local, prod, non-prod and want different modules for those environments. Ex: devel should be local only, samlauth should be cloudgov only, and in the future autologout should likely be partial (different settings for each)

loganmcdonald-noaa:
## What does this PR do? 🛠️
This splits up our config into three folders: sync (everything across all environments), local (only things added to dev like devel), and cloudgov (things like external auth and key management not needed locally).

To make this work in your environment you should only have to pull it down or deploy and then do a config:import. 

Make sure to update `settings.dev.yml` with the new example file since the config has moved and there is an important addition to make config_split work. 

You can see in Greg's environment how there is no devel but there is samlauth/s3fs: https://weathergov-greg.app.cloud.gov/admin/config

and if you pull down locally and import you should see devel but no samlauth

Fixes #562


docs/dev/extend.md
```diff
@@ -0,0 +1,20 @@
...
+1. run `composer require 'drupal/package:^version'`
+2. run `docker compose exec drupal drush en package` to enable, then `make config:export` to pull down the new enabled settings. You can also enable the module in the admin interface. 
+
+## Adding a new module to a singular environment
+
+We use [config split](https://www.drupal.org/project/config_split) so that we can have different settings in different environments. Splits come in full and partial splits so that you can have a module across environments with different settings. Adding a module in only one environment requires some extra steps when installing:
+
+1. run `composer require 'drupal/package:^version'` like you normally would
+2. run `docker compose exec drupal drush en package` to enable, then `make config:export` to pull down the new enabled settings
+3. now, remove the extension from `core.extension.yml` in config/sync. add the extension to `config/SPLIT/config_split.config_split.SPLIT.yml` _and_ add any settings files. Move the settings files that were added to sync to the split instead.
```
greg-does-weather, line 18: I'm not sure I understand this bullet. I get removing the extension from `core.extension.yml`, but after that, I don't have a `config/SPLIT/config_split.config_split.SPLIT.yml` – the directory doesn't exist. Should that be `config/sync/config_split.config_split.SPLIT.yml`?

greg-does-weather (approved):
I can confirm that this works as expected in local dev. I had to run the config import twice in order to get `devel` installed (presumably once to install config-split and the second for it to actually do its magic), but that makes sense.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Infrastructure

docs/dev/extend.md
loganmcdonald-noaa: oh yes you are correct. I changed where this lived and forgot... will update.

PR #600, opened 2024-01-08
replace h4 with h4-styled paragraph

### LINKED ISSUE #599: Prototype alerts create a misordered heading
### Description

Prototype alerts start with a level 4 heading. They appear at the top of our page, right under the level 1 heading. The result is that later level 2 or 3 headings are misordered. Because this is a prototype implementation, I think it would make sense to remove the heading entirely and replace it with heading-styled plain text. We can address this properly in the upcoming real implementation.

### Acceptance criteria

- [ ] Accessibility tests pass

greg-does-weather:
## What does this PR do? 🛠️

Remove the `<h4>` element from the prototype alert component and replaces it with an h4-styled paragraph element. This fixes the heading ordering.

Closes #599 

`make cc`


eric-gade (approved):
Temporary component, so I'll forego the big list (checks are passing)

PR #626, opened 2024-01-11
Mgwalker/api proxy

greg-does-weather:


.eslintrc.js
```diff
@@ -15,6 +15,12 @@ module.exports = {
   overrides: [
     {
       files: ["tests/**/*.js"],
+      rules: {
+        "no-console": 0,
+      },
+    },
+    {
+      files: ["tests/a11y/**/*.js", "tests/e2e/**/*.js"],
```
greg-does-weather, line 23: This only applies to the code in the proxy server. Whatever arguments might exist against `console.log`, they require way too much complexity for our very simple needs here. 😛

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -49,8 +49,11 @@ public function getAlertsForGrid($wfo, $x, $y)
      */
     public function getAlertsForLatLon($lat, $lon)
     {
+        $lat = round($lat, 4);
+        $lon = round($lon, 4);
```
greg-does-weather, line 53: It turns out that in some calls using lat/lon, if they have more than 4 digits of precision, we get a 301 redirect. Drupal has been helpfully following the redirect for us so it hasn't been a bug, but it does cause an extra network hop. So I opted to round these down to 4 decimal points to cut that hop out.

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -105,6 +105,12 @@ public function __construct(
      */
     public function getFromWeatherAPI($url, $attempt = 1, $delay = 75)
     {
+        if (!str_starts_with($url, "https://")) {
+            $baseUrl = getEnv("API_URL");
+            $baseUrl = $baseUrl == false ? "https://api.weather.gov" : $baseUrl;
+            $url = $baseUrl . $url;
+        }
```
greg-does-weather, line 112: Prepends the server URL. This is where we can inject the URL from environment variables.

eric-gade (approved):
I think this is great, and the code looks good and runs well on my end 👍 
  
Two non-blocking comments: 
1. Do we want some kind of visual indication on the page when a saved value has been used to show some information on the page? How might we achieve that and what would it look like?
2. When recording, the system is also trying to save a request to `favicon.ico` and spitting out a json file like:
```
{
  "path": "/favicon.ico",
  "correlationId": "3dd26a0b",
  "title": "Not Found",
  "type": "https://api.weather.gov/problems/NotFound",
  "status": 404,
  "detail": "'/favicon.ico' is not a valid resource path",
  "instance": "https://api.weather.gov/requests/3dd26a0b"
}
```
  Should we just have a policy of not committing these files or are they somehow necessary?

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

greg-does-weather:
Good call. Maybe the recording function should ignore return content-types that aren't JSON and statuses that aren't 2xx?

greg-does-weather:
Some kind of visual indication would be great! We could maybe have our custom server inject something? Like a colorful 🟥  or something in the place name, so it stands out?

PR #645, opened 2024-01-12
Fix current conditions narrative display bug

### LINKED ISSUE #644: The current conditions narrative text sometimes jumps up into the same row as the "widget" view
### Description

At around 1020 pixels wide when the current conditions are no wind, the narrative portion of the current conditions will jump onto the same row as the widgets.

### Acceptance criteria

- [x] Tests have been implemented or modified to capture the bug

greg-does-weather:
## What does this PR do? 🛠️

Basically the fix is to make the widget row be full-width.

Relies on #626. Found a situation where it was happening, recorded the API results, fixed up the dates, and renamed things so they'd make some kind of sense. Then I put in snapshot testing and tweaked the date/time so it would be constant.

- closes #644


eric-gade (approved):
lgtm 👍

shadkeene:
@greg-does-weather and @eric-gade 

Looked at some current conditions output and there may still be some formatting issues with this fix in staging: https://weathergov-staging.app.cloud.gov/local/MFR/105/73 and I see this and it doesn't seem right: 
staging_mfr

This is from beta for same location, and I can see it is bumped up and this is not right: 
beta_mfr

And this is from staging for NY, and this does seem right. 
ny_correct_staging


For the first example at the top, is there a fix for this or is this expected behavior?

greg-does-weather:
Also still seeing it in some cases. I've captured the Medford data so we can reproduce it. Gonna reopen the issue. I think I goofed where I put one of the classes.

greg-does-weather:
Well, I *tried* to capture that data, but I forgot to flush the Drupal cache for so there weren't actually any requests to record. 🤦‍♂️ Still, we know it's still an issue. If anyone happens to see it again, please flag! (It looks like Medford is working as expected now.)

PR #653, opened 2024-01-17
Add page load time testing

greg-does-weather:
## What does this PR do? 🛠️

Heavily inspired by #576

- Adds page load time testing for five locations. Asserts a maximum single page load time of 5 seconds and a maximum average of 3 seconds. These seem like reasonable points to throw big alarms like test failures.
  - I don't think we should make PRs dependent on these timing tests passing. They're too outside of our control. But we still need to see it, and it seems like a big ugly red icon on our PRs is a good way to make sure it's in our faces.
- Tweaks the code standards workflow
  - moves the conditionals out of the various job steps and into the jobs themselves so there's just less duplication
  - reworks the Docker build caching; the old way just didn't work
    - builds our Drupal image into a tar file and caches that
    - about 30 seconds slower when the image needs to be rebuilt, but about a minute faster the rest of the time

nothing special


eric-gade (approved):
Looks good, makes sense, works locally. I pushed one commit to add a `lt` directive to the Makefile for quickly running these tests. 

## Code Review Checklist

## The Basics
- [X] Checks are passing
- [X] I read the code
- [X] I ran the code

## Documentation

## Security
- [ ] security false positives are documented
- [ ] data from external sources is cleaned and clearly marked

## Reliability

## Infrastructure
- [ ] all changes are auditable and documented via a script
- [ ] it is clear who can and should run the script

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

greg-does-weather:
Oh bless it... trying to re-sign the commit and I accidentally dropped the Makefile change... 🤦‍♂️

Issue #589, opened 2024-01-05
Alert description field parsing


eric-gade:
### Description
We need to parse out the raw text of a weather alert's `description` field for better displaying in the UI.
  
According to the [Alert Details document](https://docs.google.com/document/d/1aKfbnZV0FiO5qwELeUczNB63eKZ2I8fmFEni6fGiS5c/edit#heading=h.undl3hq54751), the `description` field can have several formats depending on the type of hazard in the alert and who is issuing the alert. However, many alerts have a consistent textual structure in this field. Specifically, they have the what/where/when structure outlined in the linked document.
  
Note that for this ticket we only want to deal with the [`descrption` ](https://docs.google.com/document/d/1aKfbnZV0FiO5qwELeUczNB63eKZ2I8fmFEni6fGiS5c/edit#heading=h.undl3hq54751)field parts of the "[field exceptions" portion of the Alert documentation](https://docs.google.com/document/d/1aKfbnZV0FiO5qwELeUczNB63eKZ2I8fmFEni6fGiS5c/edit#heading=h.undl3hq54751). Other field exceptions will be dealt with in subsequent tickets.
  
### Prerequisites
#583 should be completed and merged before starting work on this ticket
  
### Designs
[Figma board](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=1336-19052&mode=design)
[Prototype](https://www.figma.com/proto/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?page-id=915%3A1056&type=design&node-id=1336-19384&viewport=760%2C-2982%2C0.16&t=8MRcYGAOfUF81eCv-1&scaling=scale-down&starting-point-node-id=1336%3A19384&show-proto-sidebar=1&mode=design)

### Example Data
You can find some example data of Alert responses with `description` fields [here](https://docs.google.com/document/d/1fp_kUTLF7pNb36sNwc6rvFThNPQ6lCWHIufck0Yf6iE/edit?usp=sharing). There are examples of alerts both with and without the what/where/when structure.

### Acceptance Criteria
- [ ] Unnecessary whitespace elements are removed from the description field, as outlined [here](https://docs.google.com/document/d/1aKfbnZV0FiO5qwELeUczNB63eKZ2I8fmFEni6fGiS5c/edit#heading=h.ekginlwxpntp).
[...]

PR #654, opened 2024-01-17
Alert description parsing

eric-gade:
## What does this PR do? 🛠️
Per #589, this PR implements basic Alert `description` field parsing, along with basic integration into the alert template.

The parsing is regex based. It will look for any word in all caps followed by ellipses, and use that as the label. This means we don't specifically search for WHAT/WHEN/WHERE/IMPACTS tokens. If we want to be more literal to the observed format, and more strict, we can switch over.
  
When the regex matches successfully, the `description` field on the processed alert object will become an associative array whose keys are the labels in lowercase ("what" "where" etc) and whose values are the text content for that label.
  
When the regex fails to match, the plain description string is displayed, with the white space adjusted.

<details>
<summary>Before:</summary>
  
Screen Shot 2024-01-17 at 6 27 25 PM

</details>

<details>
<summary>After:</summary>
  
Screen Shot 2024-01-17 at 6 24 23 PM

</details>

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherAlertTrait.php
+++ b/web/modules/weather_data/src/Service/WeatherAlertTrait.php
@@ -33,6 +33,23 @@ trait WeatherAlertTrait
         return $str;
     }
 
+    static function tryParsingDescriptionText($str)
+    {
+        $regex = "/\*\s*(?<label>[A-Z]+)\.\.\.(?<content>.*)(\n{2}|$)/sU";
+        if(preg_match_all($regex, $str, $matches)){
+            $result = [];
+            for($i = 0; $i < count($matches['label']); $i++){
+                $label = strtolower($matches['label'][$i]);
+                $content = $matches['content'][$i];
+                $result[$label] = $content;
+            }
+
+            return $result;
+        }
+
+        return $str;
+    }
+
     /**
      * Get active alerts for a WFO grid cell.
      */
@@ -110,4 +127,6 @@ trait WeatherAlertTrait
 
         return $alerts;
     }
+
+    
 }
--- a/web/modules/weather_data/src/Service/WeatherAlertTrait.php
+++ b/web/modules/weather_data/src/Service/WeatherAlertTrait.php
@@ -84,9 +84,17 @@ trait WeatherAlertTrait
                 $output->geometry = [];
             }
 
-            $output->description = self::fixupNewlines(
-                $output->description ?? false,
-            );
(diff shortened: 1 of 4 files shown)


web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -67,9 +84,19 @@ public function getAlertsForLatLon($lat, $lon)
...
+                $output->description,
             );
+            if (!is_array($alertDescription)) {
+                $output->description = self::fixupNewlines(
+                    $alertDescription ?? false,
+                );
+                $output->usesParsedDescription = false;
+            } else {
+                $output->description = $alertDescription;
+                $output->usesParsedDescription = true;
```
eric-gade, line 97: There is no good way in a twig template to determine if some attribute on an object in question is or is not an array. Here the `alert.description` field will be an array with named keys. Otherwise it's just the description string itself. We use this boolean field in the twig template to determine how we display the field.

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -33,6 +33,23 @@ protected static function turnToDate($str, $timezone)
         return $str;
     }
 
+    public static function tryParsingDescriptionText($str)
+    {
+        $regex = "/\*\s+(?<label>[A-Za-z\s]+)\.\.\.(?<content>.*)(\n{2}|$)/sU";
```
greg-does-weather, line 38:
```suggestion
        // look for any word in all caps followed by ellipses, and use that as the label, and
        // all text until the next pair of newlines is the content
        //
        // https://regexper.com/#%2F%5C*%5Cs%2B%28%5BA-Za-z%5Cs%5D%2B%29%5C.%5C.%5C.%28.*%29%28%5Cn%7B2%7D%7C%24%29%2F
        $regex = "/\*\s+(?<label>[A-Za-z\s]+)\.\.\.(?<content>.*)(\n{2}|$)/sU";
```

greg-does-weather (approved):
I haven't pulled this down and run it locally yet, but the code looks good!

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [ ] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

Issue #592, opened 2024-01-05
Alert areaDesc formatting


eric-gade:
### Description
The `areaDesc` field of an Alert, if present, lists in plain text the areas impacted by the hazard. Depending on the type and location of hazard, the areas may be counties, cities, mountain ranges, or other types of location. 
  
They are presented in a list, separated by semicolons. 
  
Example: Hamilton; Inland Nassau; Northern Columbia; Atkinson; Pierce; Brantley; Inland Glynn; Echols; Clinch; Inland Camden; Northern Ware; Northeastern Charlton; Southern Ware; Western Charlton

### Prerequisites
#583 must be completed and merged before this work can begin.
  
### Designs
[Figma here](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=1336-19052&mode=design).

### Acceptance Criteria
- [ ] Turn this text list into an unordered list
- [ ] Remove the semicolons
- [ ] Display the unordered list in a bulleted list 
- [ ] For counties effected - use a two column layout

colinmurphy01:
From Des/Eng sync: For counties effected - use a two column layout

PR #661, opened 2024-01-19
Weather alert area description formatting

eric-gade:
## What does this PR do? 🛠️
This PR implements the changes requested in #592

After some messing around with USWDS grid options, we determined the best choice was to use an inline `columns:2` style string on the unordered list.

Screen Shot 2024-01-19 at 10 08 50 AM

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherAlertTrait.php
+++ b/web/modules/weather_data/src/Service/WeatherAlertTrait.php
@@ -104,7 +104,13 @@ trait WeatherAlertTrait
             $output->instruction = self::fixupNewlines(
                 $output->instruction ?? false,
             );
+
             $output->areaDesc = self::fixupNewlines($output->areaDesc ?? false);
+            if ($output->areaDesc) {
+                $output->areaDesc = array_map(function ($description) {
+                    return trim($description);
+                }, explode(";", $output->areaDesc));
+            }
 
             $output->onset = self::turnToDate(
                 $output->onset ?? false,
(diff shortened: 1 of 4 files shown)


web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -104,7 +104,13 @@ public function getAlertsForLatLon($lat, $lon)
             $output->instruction = self::fixupNewlines(
                 $output->instruction ?? false,
             );
+
             $output->areaDesc = self::fixupNewlines($output->areaDesc ?? false);
+            if ($output->areaDesc) {
+                $output->areaDesc = array_map(function ($description) {
+                    return trim($description);
+                }, explode(";", $output->areaDesc));
```
greg-does-weather, line 112: Ugh, the way PHP does this is just so ugly. Is it so wrong to want something more fluent?

greg-does-weather (approved):
My local environment is a mess right now so I didn't pull it down to run it, but all the tests pass and I pulled the screenshots from the automated tests. They look great at all three aspect ratios, and the code reads like it ought to be accessible.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [ ] I ran the code

## Reliability

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #702, opened 2024-01-25
update drupal core

loganmcdonald-noaa:
This updates our modules with dependencies in Drupal Core to 10.2 as recommended for security and feature updates. It also updates to a compatible Drush. There is one change that needed to be made to xdebug - it needs to be downgraded since there is a known issue with 10.2. See https://git.drupalcode.org/project/drupal/-/merge_requests/5774

### DIFF
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -222,7 +222,7 @@ jobs:
         uses: actions/cache@v3
         id: cache
         with:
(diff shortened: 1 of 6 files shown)


loganmcdonald-noaa:
Locally and after deployment: 

Screenshot 2024-01-25 at 11 46 09 AM

greg-does-weather (approved):

PR #714, opened 2024-01-30
Add a `/bundle` endpoint to fully capture all the API data in a single page load

greg-does-weather:
## What does this PR do? 🛠️

The API proxy server now has an endpoint at [localhost:8081/bundle](http://localhost:8081/bundle) that will enable capturing a "bundle" of API responses associated with the same page load. The result is a ZIP file that we can share more easily. Just unzip it locally and then visit the same location page that generated the bundle. (You may have to clear your Drupal cache first, or wait a minute or so for them to expire.)

Also adds a header to our outgoing API requests identifying the Drupal response we're servicing. That's necessary for the whole "bundling" thing, but we might other uses for it in the future. It's not a personal identifier – it's based on the current time.

### DIFF
--- a/docs/dev/intercepting-the-api.md
+++ b/docs/dev/intercepting-the-api.md
@@ -70,3 +70,16 @@ from the current time by adding a value and unit: `date:now +1 hour`. The value
 and unit are passed directly into
 [Day.js's `.add` method](https://day.js.org/docs/en/manipulate/add), so use that
 documentation to know what units are available.
+
+## Recording a whole transaction
+
+Sometimes it's easier to capture a whole series of API calls related to a single
+page load so we can attach it to a bug report or issue for future investigation.
+To help with this, the proxy server has another utility that will capture all
+of those calls and produce a zip file with the results.
+
+To enable this, visit [localhost:8081/bundle](http://localhost:8081/bundle). The
+proxy will capture all API calls associated with the next location page load.
+Once that location response is finished, the proxy will revert back to its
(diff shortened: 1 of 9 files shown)


eric-gade:
I spent a bit of time trying to figure out why I couldn't get the bundle to save, then remembered that the `/local/TST/10/10` test route uses the proxy to fetch preloaded data. That seems to actually crash Drupal and no subsequent requests to the main web server work. This might be because the proxy server itself is also crashing.
  
The server(s?) also crashed in the case where I activated the bundle saving, then went to another tab and simply refreshed a page I had already loaded. I'm thinking this has to do with the fact that we are loading from the cache and not making any api requests at all, though looking at the code I'm not sure why that should crash anything rather than simply not writing a saved bundle.

greg-does-weather:
Good debug info! Thank you! Will look into it.

greg-does-weather:
Alright, so I think everything you saw was a result of the way I was handling ending a bundle. Basically, if you enabled a bundle, it would disable itself automatically after 3 seconds. The original assumption was you'd turn it on, run your requests, and that'd be that, and I needed a way for the server to know that the bundle was finished. Instead of centralizing that, I tried to make both the main and save components track it, and when they got out of sync, it was a bad time for everyone.

I just pushed up a commit that centralizes configuration into one place. Now main doesn't need to worry about it, and the save component is responsible for signaling when the bundle is finished.

There was also a logic error where if bundling was enabled, it would try to skip reading local files, which is very silly.

eric-gade (approved):
👍

PR #717, opened 2024-01-30
Replace static CSS/script imports with Drupal placeholder tokens

### LINKED ISSUE #234: Replace static css and js link tags with Drupal placeholder_token

greg-does-weather:
## What does this PR do? 🛠️

- Defines libraries in our theme so we can import them via name
  - `uswds-init` is for the init script, so we can specify that it goes in the `head`
  - `uswds-customized` for our built `styles.css` and the USWDS scripts
  - `location-search` for powering the location search
  - `localize-timestamps` for powering localization of timestamps
- Updates our theme to, by default, import the two USWDS-related libraries on every page
- Updates our `html.html.twig` template to remove the static `<link>` and `<script>` tags
- Manually attach the `location-search` and `localize-timestamp` libraries in the twig templates that need them, so they only get loaded on relevant pages

Going forward, if there are scripts that only need to appear in certain places, this may be a template for how we can do that.

closes #234

`make cc` may be enough, but the theme cache has to be cleared out in order for Drupal to re-read the libraries stuff.

n/a because nothing should have changed from the user's perspective


eric-gade (approved):
Just a quick approval without checklist here -- this looks good to me. Good that location search is now only on the relevant pages.

PR #720, opened 2024-01-31
Add global template data

greg-does-weather:
## What does this PR do? 🛠️

Implements the `template_preprocess_default_variables_alter` hook to inject some of our own globals. We can add to this later as-needed, but for now, our globals object looks like this:

```
{
  alerts: <bool; default false>,
  place: <bool|string; default false>
}
```

Then we can use those in any of our twig templates with `{{ weather.alerts }}` or `{{ weather.place }}`. This PR also adds the location page place name to the page title using this global data.

I don't think we can do dependency injection on these module-level hook handlers, so it's a little messy and... perhaps untestable? Not thrilled about that but I mean... I suspect we're already testing a lot more than most Drupal sites.

`make rebuild && make cc`

### DIFF
--- a/web/modules/weather_blocks/weather_blocks.module
+++ b/web/modules/weather_blocks/weather_blocks.module
@@ -1,1 +1,46 @@
 <?php
+
+// This hook lets us alter the variables at the page level. These variables are
+// available globally to all templates.
+function weather_blocks_template_preprocess_default_variables_alter(
+    &$variables,
+    $hook,
+) {
+    // Let's create a single global that contains all of our stuff, rather than
+    // scattering them around. Yay, namespaces!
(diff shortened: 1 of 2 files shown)


eric-gade (approved):
It's working!
  
If we start adding more API data to this we will want to make sure our cache implementations are set up right  

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

PR #722, opened 2024-02-01
Add end-to-end tests for alert accordions

greg-does-weather:
# HOLD
This will need potentially significant revision when tabbed navigation goes in, and since that's pretty close, let's just hold this PR until then.

## What does this PR do? 🛠️

- updates saved data so the `TST/10/10` location doesn't need to go to the API at all (oops)
- adds tests to verify that the first alert accordion is expanded and the others are collapsed

# TODO:
- adds tests to verify that alerts expand and collapse when clicked


eric-gade (commented):
I'm thinking we can close this PR. As you suspected, I think the e2e tests for tabbed navigation cover the accordion test cases.

greg-does-weather:
You are correct!

Issue #667, opened 2024-01-19
UX: Upon clicking a hazard banner, expand hazard details


colinmurphy01:
Upon clicking a hazard banner, I want to jump to the corresponding hazard details (current) AND have that accordion automatically expand (desired).

Current state video: first banner is automatically expanded (which is demoed and desired), but after clicking second banner it doesn't expand, and after collapsing first accordion and clicking again it doesn't expand.

https://github.com/weather-gov/weather.gov/assets/24654605/dc0d427e-992c-444a-a77c-d1bdbf803f35

colinmurphy01:
@igorkorenfeld can you confirm that this is desireable?

partly-igor:
I hadn't thought about that interaction before it was implemented, but yes you are right that's likely the interaction a user would expect – clicking the anchor link in the alert jumps down to an expanded text.

greg-does-weather:
We can pick this up after #681. The code in there is already poking around the relevant DOM elements, so it should be fairly easy at that point.

Issue #681, opened 2024-01-22
Add tabbed navigation to basic weather


greg-does-weather:
## Tabbed navigation
Only a couple of notes here, this one likely will surface a lot of questions, but would be happy to pair on this and answer questions as they come up. 

[Annotated artboard](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=2196%3A31948&mode=design&t=bOEgGOPKzi4hAEyK-1)

Component of #647

eric-gade:
Closed by #728

PR #728, opened 2024-02-01
Tabbed Navigation

eric-gade:
## What does this PR do? 🛠️
This PR implements the tabbed based navigation as outlined in #681 and in the current redesigns.

### The `<tabbed-nav>` custom element ###
For this to work, we have implemented a new web component (custom element) called `tabbed-nav`. This component does not make any use of the shadow DOM or anything super fancy -- it merely makes use of the component lifecycle methods to bind the correct events and sets up the ability to switch tabs.
  
You can find the definition of this component [here](https://github.com/weather-gov/weather.gov/blob/eg-681-tabbed-nav/web/themes/new_weather_theme/assets/js/components/TabbedNavigator.js).
  
### Expected behaviors and complexity ###
Due to the presence of the "above-the-fold" alert links -- which themselves link to alert accordion items that may need to be expanded when being navigated to -- and now the additional layer of alerts possibly being in a different tab than the "current" tab, the custom element needs to deal with several complex situations.
  
The ultimate case here is where the tab is on, say, the "Daily" forecast, but the user scrolls up and clicks an alert link. Several things need to happen:
1. The correct tab that contains the alert needs to be selected and its content pane needs to be shown
2. The accordion in which the corresponding alert lives needs to be opened, if it isn't already
3. The browser needs to scroll to the correct position

Note that (3) above carries extra complexity, because the `display-sticky` nature of the tab button area means extra calculation needs to be done in JS to know where to scroll, so that the alert accordion header is visible. This is all handled within the custom element.

### Anchors and hashes ###
The USWDS accordion components work, in part, by setting the hash fragment of the page to the `id` of the accordion (so that the browser scrolls to it by default). A problem emerges when a user shares a link with that hash fragment still present: the page needs to "know" to do steps 1, 2, and 3 outlined above. Our component now handles that situation as well.
  
Additionally, _clicking_ any of the tabbed-nav tabs will update the hash fragment to be the id of the tab container (though programmatically selecting a tab, as in the steps above, will _not_ change the hash). This is so that users can share a link that, for example, ends in `#daily`, and upon loading the page the Daily tab will be activated and scrolled to.
  
### Alerts tab is not yet conditional ###
The "Alerts" tab will currently be present on every location page, whether or not the location has any active alerts. This is temporary. The work in #720 will enable conditional rendering of the alerts tab and it will be handled in a subsequent ticket.

<details>
<summary>Expand</summary>
 
Screen Shot 2024-02-01 at 1 39 59 PM
Screen Shot 2024-02-01 at 1 39 54 PM
Screen Shot 2024-02-01 at 1 39 49 PM
Screen Shot 2024-02-01 at 1 39 43 PM

</details>
  
## Closes ##
#681


greg-does-weather:
Seems like this closes #667 too.

web/themes/new_weather_theme/assets/js/components/TabbedNavigator.js
```diff
@@ -0,0 +1,146 @@
...
+    if (matchedTabButton) {
+      this.switchToTab(matchedTabButton.dataset.tabName);
+      matchedTabButton.parentElement.scrollIntoView();
+      return;
+    }
+
+    const childElement = this.querySelector(`${hash}`);
+    if (childElement) {
+      const tabContainer = childElement.closest(".tab-container");
+      this.switchToTab(tabContainer.id);
```
greg-does-weather, line 66: What happens if the hash refers to an element that isn't side our tab containers? Would it be worth adding an `if(tabContainer)` ahead of switching tabs, just to be safe?

tests/e2e/cypress/e2e/tabbed-nav.cy.js
```diff
@@ -0,0 +1,124 @@
...
+            .should("not.exist");
+        });
+      });
+    });
+
+    describe("Intercepts click events on above-the-fold alert links", () => {
+      it("Clicking an alert link opens the accordion for that link and scrolls to it", () => {
+        cy.get("weathergov-alert-list a").each((anchorEl) => {
+          const alertId = anchorEl.attr("href").split("#")[1];
+          cy.get(`#${alertId}`).as("alertEl").click();
```
greg-does-weather, line 84: This look like it's clicking the accordion element, not the link?

```suggestion
          const alertId = anchorEl.attr("href").split("#")[1];
          cypress.wrap(anchorEl).click();
          cy.get(`#${alertId}`).as("alertEl");
```

Checking my thinking with Cypress, I opened the UI tool so I could rerun just this one test and I noticed that you can expand the test. And then I was surprised you can expand the list of steps, and you can hover over a specific step to see what element it was operating on. Here's the test as-is:

image

Which maybe isn't super illustrative. But anyway, when I make that change above, here's what it looks like:

image

That is so cool! I feel like that's going to be a really useful tool for making sure tests are doing what we intend for them to.

greg-does-weather (approved):
Looks good! Well done.

A question for @partly-igor: should the tab buttons have a pointer cursor to indicate that they're clickable?

I didn't do accessibility testing because of GSA IT. But I'll open a debt ticket to come back and do a11y testing on the tabbed navigation.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

partly-igor:
@greg-does-weather that's a good point, and I also will owe you some hover/active state styles, but was planning on tackling along with the other style updates. 

The other a11y debt I'll note that we likely want to capture at the same time is handling left and right keyboard press, as well as additional ARIA roles for the tabs and tab panel. This is coming from the the w3c recommendations for [tab pattern ](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

.eslintrc.js
```diff
@@ -19,6 +19,12 @@ module.exports = {
         "no-console": 0,
       },
     },
+    {
+      files: ["web/themes/new_weather_theme/assets/js/components/**/*.js"],
+      rules: {
+        "class-methods-use-this": 0,
```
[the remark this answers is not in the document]
eric-gade: Yes. It's a terrible rule, because you can't actually use a static method without referencing the class by name or by constructor, which is clunkkkkkky

web/themes/new_weather_theme/new_weather_theme.libraries.yml
```diff
@@ -28,4 +28,110 @@ localize-timestamps:
...
+  version: VERSION
+  css:
+    component:
+      css/components/search-results.css: {  }
+user:
+  version: VERSION
+  css:
+    component:
+      css/components/user.css:
+        weight: -10
```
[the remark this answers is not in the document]
eric-gade: You know, I have no idea what this is. Must be left over from when I rebased onto main?

web/themes/new_weather_theme/assets/js/components/TabbedNavigator.js
eric-gade: The only other option is for it to be a tab button, which we already deal with in the case(s) above. But -- if we change the structure of the component children then you are right, it could present a problem. We might also want to actually enforce a shadow dom and slots, ensuring that we _don't_ display anything that isn't a button area or a tab container...

tests/e2e/cypress/e2e/tabbed-nav.cy.js
eric-gade: I dealt with it in this commit: https://github.com/weather-gov/weather.gov/pull/728/commits/987cf87d82704314343746cdca5206ea5cc6d0a8

web/themes/new_weather_theme/assets/js/components/TabbedNavigator.js
mgwalker: Yeah, fair enough. Could it pop up if again if we add hash navigation elsewhere on the page? In any case, no need to deal with it right now.

Issue #713, opened 2024-01-29
Parsing: Additional Details being cut off after first paragraph


shadkeene:
### Description

When there are additional details that span more than one paragraph, either beta.weather.gov or API seems to be cutting off the second paragraph.  Images below show the difference.

### Acceptance criteria

beta.weather.gov shows entire "additional details" section for hazards.

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility
![additional_deets_full](https://github.com/weather-gov/weather.gov/assets/139174561/986b0e92-bd42-4e01-8603-1243d0a7d25b)
![additional_deets_partial](https://github.com/weather-gov/weather.gov/assets/139174561/9f8a925a-2768-4105-b935-9312983cdf92)

shadkeene:
It does look like alerts API has both paragraphs
![alerts_fulldetails](https://github.com/weather-gov/weather.gov/assets/139174561/c1aec88b-3d05-4aa6-a3e8-3ddb3933e462)

greg-does-weather:
That flood warning from the API:

[active__status=actual&point=37.7812,-122.4263.json](https://github.com/weather-gov/weather.gov/files/14090344/active__status.actual.point.37.7812.-122.4263.json)

greg-does-weather:
Related to #712

greg-does-weather:
Closed by #748

PR #748, opened 2024-02-05
Alert parsing refactor

### LINKED ISSUE #712: What to do with "Overview" section of hazard
### Description

Not sure if this is a bug or enhancement, but some products include an "overview" section, like the section circled below.  The current beta.weather.gov does not include the "overview" section.  I'm not sure if this was intended or not, or if the "overview" section is not available on the API.

### Acceptance criteria

We are all aware of the "overview" section and we know how to deal with that when it is available.

overview
no_overview

eric-gade:
## What does this PR do? 🛠️
This PR refactors the Alert description field parsing, in major part to address the two issues highlighted in #713 and #712

### Introduction of `WeatherAlertParser` class (and restructuring) ###
This PR moves the parsing functions out of the `WeatherAlertTrait` and into its own parsing class. We chose this restructuring because it allows us to more easily add parsing functionality in the future (if needed) while also maintaining a separation of logic. It also makes the corresponding test files less daunting.
  
### Addition of examples in local dev application ###
I have added examples from each (screenshots below) of the issues into the local proxy data, available at `http://localhost:8080/local/TST/10/10`.
  
### Description of parsing ###
Previously, we were using one long regex to pull out any structure we wanted. The way we implemented this meant that we were also losing information -- extra paragraphs (per #713) or important preamble paragraphs (per #712) that didn't meet the expected structure.
  
The new parser is a little simpler, and certainly more composible. Its primary purpose is to return an array of "nodes," where each node is itself an assoc array that has a `type` and `text` field. For now we only have "heading" and "paragraph" types.
  
The alert description is split up into paragraphs, then each paragraph is tested for further structure. If structure is found, the appropriate nodes are created and added to the final list. If not, we create one paragraph type node, set the "text" of the node to be the whole paragraph, then add that node to the final list. In this way, no text is left out. By default all non-structured text will be wrapped in paragraph nodes.
  
<details>
<summary>Example for #712</summary>

Screen Shot 2024-02-05 at 1 57 34 PM

</details>

<details>
<summary>Example for #713</summary>

Screen Shot 2024-02-05 at 2 33 29 PM

</details>
  
## Closes
#712 
#713


web/modules/weather_data/src/Service/WeatherAlertParser.php
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return $str;
+    }
+
+    public function __construct(string $descriptionString)
+    {
+        $this->descriptionString = $descriptionString;
+    }
+
+    public function parse()
```
greg-does-weather, line 31: Just curious why use an instance method here and a static one above. Could this be

```php
public static function parse($str)
```

And then call it with `WeatherAlertParser::parse($description);`?

I'm not suggesting let's do it, just curious.

web/modules/weather_data/src/Service/WeatherAlertParser.php
```diff
@@ -0,0 +1,123 @@
...
+    {
+        $cleanedDescriptionString = self::fixupNewlines(
+            $this->descriptionString,
+        );
+        $paragraphs = preg_split("/\r\n|\n|\r/", $cleanedDescriptionString);
+        // Remove blank strings
+        $paragraphs = array_filter($paragraphs, function ($paragraph) {
+            return $paragraph != "";
+        });
```
greg-does-weather, line 40: Oh, clever, because the only newlines left should be pairs.

web/modules/weather_data/src/Service/WeatherAlertParser.php
```diff
@@ -0,0 +1,123 @@
...
+     * asterisk, then an all-caps label, followed by
+     * ellipses, and then the rest of the text.
+     *
+     * Append appropriate parsed nodes to the end of
+     * the nodes array if matched and return true.
+     * Otherwise, return false.
+     */
+    public function parseWhatWhereWhen($str, &$nodes)
+    {
+        $regex = "/^\*\s+(?<heading>[A-Za-z\s]+)\.\.\.(?<text>.*)$/";
```
greg-does-weather, line 107:
```suggestion
        $regex = "/^\*\s+(?<heading>[A-Z\s]+)\.\.\.(?<text>.*)$/";
```
The comments say the heading is all-caps, so should we be matching for that? Or is it more... we think it's *supposed* to be all-caps, but we're not sure, and the pattern otherwise holds so let's be flexible? 😛

greg-does-weather (approved):
Faboo!

Wakko Warning typing furiously while his siblings watch on

## Code Review Checklist

Using a smaller checklist since nothing in the UI should have changed in a way that needs testing.

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

web/modules/weather_data/src/Service/WeatherAlertParser.php
eric-gade: @greg-does-weather I went ahead and pushed some changes with what I (probably?) originally intended: https://github.com/weather-gov/weather.gov/pull/748/commits/7512fff23320df33bd55f78342498c1866dd2124
  
LMK what you think

greg-does-weather: 👍 LGTM!

PR #750, opened 2024-02-06
Load spatial data

greg-does-weather:
## What does this PR do? 🛠️

* Adds scripts and documentation for loading geospatial data locally and in cloud environments.
* Adds our current geospatial data sources as git LFS files

closes #723

PR #751, opened 2024-02-06
Use local spatial data for place name and alert filtering

### LINKED ISSUE #724: [GIS] Use local geospatial data to filter alerts
### Description

The API's `/alerts/active?location={lat,lon}` endpoint returns all alerts – land or marine – for that point. However, we can't guarantee that any point we choose from a grid cell will be on land. For example, with grid LIX 113,110, all four corners and the centroid are over water. A visitor from that small part of Biloxi, MS, would never receive land alerts because we would _always_ ask the API for a location over water.

Instead, we should use the API's `/alerts/active?place={state}` endpoint. This only returns land hazards. We can find the state by finding which one intersects the grid polygon. The alerts returned can be either zone-based or polygon-based. For zone-based alerts, we can find the grid's county (also by doing an intersection) and only keep the zone-based alerts whose SAME codes include the county's FIPS code. For polygon-based alerts, we can keep the alerts whose polygon intersects the grid's polygon.

See the [spatial proposal doc](https://docs.google.com/document/d/1Cp0VfgD6HuMU9ZVWX0UKzZVLpVUM7-mpvC1ihvAKxFY/edit#heading=h.5esd92rc2393) for more details.

- For determining which alerts to show, should we intersect an area larger than the grid? Imagine for example that someone in western Nashville has come to the site. They type "Nashville, TN" into the search box, which returns a grid in downtown Nashville. If western Nashville is under a tornado warning but downtown Nashville is not, this user will be unwarned.
  - We should look into how legacy handles this kind of situation.
  - We may want to expand the bounds of the grid polygon before filtering alerts.
  - Using the "use my location" button should produce the most accurate results in most, but not everyone is going to want or be able to do that. We also can't super rely on it – for privacy reasons, many browsers and operating systems now "fuzz" location data.

### Dependencies

- #723

### Acceptance Criteria
- [ ] Alerts are filtered by polygon instead of by point

### LINKED ISSUE #725: [GIS] Use local geospatial data to find place name
### Description

Use the Geonames dataset loaded into our geospatial tables to find a place name instead of relying on the API or ESRI.

- We should do at least some light performance testing to ensure doing the geospatial stuff ourselves isn't slowing down our page load times. I think it should be *faster* but we'll have to see.
- Do we need to clean up the Geonames dataset?
  - We know there's a Carl's Jr. location in there, for example. We should probably take that one out.
  - New York City is represented multiple times, all at exactly the same coordinates. Should we yoink that one too?
  - Should we devise some rules about how we automatically clean this data? We might need to reload it at some point. If we come up with rules, we should encode those into scripts.

### Dependencies

- #723 

### Acceptance Criteria
- [ ] Place names come from the Geonames cities500 dataset
- [ ] The team has fiddled with it in a live environment and generally agrees the results are better than our previous place names

### LINKED ISSUE #726: [GIS] Get a location's timezone from our local geospatial data
### Description

We currently ask the API to tell us what timezone a grid is in. However, we can now get that information from our own local data. Let's do.

### Dependencies

- #723 

### Acceptance Criteria
- [ ] Timezone is pulled from our local data instead of the API

greg-does-weather:
## What does this PR do? 🛠️

- Removes time zone trait, because we weren't using it anywhere
- Modifies `getPlaceFromGrid` to get a grid's polygon and find the place from our `weathergov_geo_places` table that is nearest to it
  - Injects a `$self` variable that defaults to `false`. At runtime, if `$self` is false, it is assigned `$this`. This lets us inject a mocked WeatherDataServer object into our methods so we can isolate them better.
  - If after talking about it, we like this approach, I'll make a debt ticket to update our other tests accordingly
- Updates the weather alert trait `getAlertsForGrid` behavior:
  - fetches all alerts for the state associated with the current place
  - removes marine alerts immediately (no sense wasting cycles to process them)
  - if an alert has no geometry, check if its list of SAME codes includes our county's FIPS code (plus a leading zero, because SAME codes have leading zeroes 🤷‍♂️)
  - if an alert has a geometry, check if the alert geometry intersects our grid geometry
- removes the `getAlertsFromLatLon` method because we're not using it
- fixes up tests

Time zones and place names were easy, so I rolled them into this PR to be done.

- closes #724
- closes #725
- closes #726

## Known issues

- gets all alerts for a county, regardless of zone or issuing WFO
  - we need to load zones, find the zone(s?) for the grid, and filter on that, not on counties
  - deffo a bug
- still shows alerts that have not yet begun (ie, `onset` is in the future)
  - I think this is an undocumented bug?

Don't try reviewing it locally until after #750 is merged and you've loaded spatial data into your local database.

`make rebuild && make cc`


tests/api/data/alerts/active__status=actual&area=AL.json
```diff
@@ -19,127 +19,10 @@
         "id": "urn:oid:2.49.0.1.840.0.21973bd423ae0db39e5260d08fa31bb6eba8b288.001.1",
         "areaDesc": "Stewart; Montgomery; Robertson; Sumner; Macon; Clay; Pickett; Houston; Humphreys; Dickson; Cheatham; Davidson; Wilson; Trousdale; Smith; Jackson; Putnam; Overton; Fentress; Perry; Hickman; Lewis; Williamson; Maury; Marshall; Rutherford; Cannon; De Kalb; White; Cumberland; Bedford; Coffee; Warren; Grundy; Van Buren; Wayne; Lawrence; Giles",
         "geocode": {
-          "SAME": [
```
greg-does-weather, line 22: For all these saved data files, I had to make a few changes:

1. Because we now get alerts based on state, I needed to change the filename from `&point=[lat,lon]` to `&area=[state]`. So I also had to update the `gridpoints/0,0.json` (etc) files to have real coordinates that would resolve to real places. I just started doing states in alphabetical order. (I skipped Alaska because it felt rude to use a state as a test case when we can't actually show anything for large swaths of it.)
2. We also don't just accept the data we get. We apply additional filtering on top of it, so we need to make sure the alerts will pass. Rather than trying to get them to fit a polygon, I opted to just update the SAME codes for all of them.

web/modules/weather_data/src/Service/Test/Base.php.test
```diff
@@ -79,6 +94,10 @@ abstract class Base extends TestCase
 
         $this->cacheMock = $this->createMock(CacheBackendInterface::class);
         $this->requestMock = $this->createStub(Request::class);
+        $this->databaseMock = $this->createStub(Connection::class);
+        $this->selfMock = $this->getMockBuilder(WeatherDataService::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
greg-does-weather, line 100: You'll see below how this gets used, but in a nutshell, it's a way for us to isolate one method from all the others in the service. It should mean less mocking HTTP responses.

web/modules/weather_data/src/Service/Test/GetPlaceFromGrid.php.test
```diff
@@ -2,47 +2,78 @@
...
+                        "WFO",
+                        1,
+                        2,
+                        [
+                            (object) ["lat" => 7, "lon" => 8],
+                            (object) ["lat" => 9, "lon" => 10],
+                        ],
+                    ],
+                ]),
+            );
```
greg-does-weather, line 31: So instead of mocking out an API HTTP response, we can instead directly mock the `getGeometryFromGrid` method.

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -57,27 +57,65 @@ public static function tryParsingDescriptionText($str)
...
-        $geometry = $this->getGeometryFromGrid($wfo, $x, $y);
-        $point = $geometry[0];
+        if (!$self) {
+            $self = $this;
+        }
 
-        return $this->getAlertsForLatLon($point->lat, $point->lon);
-    }
+        $CACHE_KEY = "alerts $wfo/$x/$y";
+        $cache = $this->cache->get($CACHE_KEY);
```
greg-does-weather, line 67: Always use `$this` when using an injected service...

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -57,27 +57,65 @@ public static function tryParsingDescriptionText($str)
...
 
-    /**
-     * Get active alerts for a latitude and longitude.
-     */
-    public function getAlertsForLatLon($lat, $lon)
-    {
-        $lat = round($lat, 4);
-        $lon = round($lon, 4);
+        $geometry = $self->getGeometryFromGrid($wfo, $x, $y);
+        $place = $self->getPlaceFromGrid($wfo, $x, $y);
```
greg-does-weather, line 73: ...and use `$self` for our methods.

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -57,27 +57,65 @@ public static function tryParsingDescriptionText($str)
...
+        $timezone = $place->timezone;
+
+        $alerts = $self->getFromWeatherAPI(
+            "/alerts/active?status=actual&area=$place->state",
+        )->features;
+
+        $geometry = array_map(function ($point) {
+            return $point->lon . " " . $point->lat;
+        }, $geometry);
+        $geometry = implode(",", $geometry);
```
greg-does-weather, line 83: A future refactor would be to have `getGeometryFromGrid` just return the array of points instead of massaging them into a funky object. What's also kind of annoying is that in both places we get this geometry, we still have to transform it into a [WKT representation](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry) anyway.

Maybe it could be `getGeometryWKTFromGrid` instead... 🤔

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -57,27 +57,65 @@ public static function tryParsingDescriptionText($str)
...
+
+        $alerts = array_filter($alerts, function ($alert) use (
+            $place,
+            $geometry,
+        ) {
+            if (AlertPriority::isMarineAlert($alert->properties->event)) {
+                return false;
+            }
```
greg-does-weather, line 91: Previously we removed marine alerts after transforming them. However, since we now do a little bit of math on the alerts, it makes sense to me to filter them out as early as possible.

web/modules/weather_data/src/Service/Test/GetPlaceFromGrid.php.test
```diff
@@ -2,47 +2,78 @@
...
-    public function testHappyPath(): void
-    {
+
         $expected = (object) [
-            "city" => "City",
-            "state" => "Not City, but State",
+            "city" => "Cityville",
+            "state" => "ST",
+            "stateName" => "Statesylvania",
+            "stateFIPS" => "fipsy fips",
```
eric-gade, line 65: valid

web/modules/weather_data/src/Service/WeatherAlertTrait.php
eric-gade: Should we just use `$self` in here across the board, or is the idea only to use it to call methods we actively test? (asking because of the call to `$this->cache` etc)

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -32,27 +32,90 @@ public static function tryParsingDescriptionText($str)
...
+                    // SAME codes are FIPS codes with a leading 0
+                    "0$place->countyFIPS",
+                    $alert->properties->geocode->SAME,
+                );
+            }
 
-        $timezone = $this->getTimezoneForLatLon($lat, $lon);
+            // If there's no geometry, zone, or county information, then we
+            // just... skip this one. But we should totally log this
+            // situation because something is wrong.
```
eric-gade, line 116: Let's link this to #754

eric-gade (approved):
giphy-working
  
(just a few inline comments, non blocking)

## Code Review Checklist
Checking off if true or if not applicable  

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

Issue #733, opened 2024-02-01
The new tabbed navigation needs to be accessibility tested


greg-does-weather:
### Description

The tabbed layout introduced in #728 wasn't accessibility tested as part of the code review. (Blame GSA IT.) We need to do that sooner rather than later.

### Acceptance criteria

- [ ] Changes have been tested for accessibility

eric-gade:
Additionally, per @partly-igor , we should ensure that we are using good ARIA attributes according to the [tab pattern described here](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/). I think maybe we should add that as an AC to this issue.

greg-does-weather:
Done in #753

PR #753, opened 2024-02-07
Tabbed navigation a11y fixes

eric-gade:
## What does this PR do? 🛠️
This PR addresses #733 by adding the expected accessibility behavior and attributes as [recommended by the W3C per these guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/).

Appropriate ARIA attributes and relationships have been added where needed, and we've added a few e2e tests to make sure the navigation that W3C recommends is implemented properly. Specific nav things to look out for:
* If a tab is in focus, pressing the `Tab` key __should not__ move to the next tab button, but rather put the focus into the corresponding tabpanel (ie the tab's content container).
* Left/right arrow keys are used to put the focus on the next or previous tab button, if a tab button is currently in focus (instead of the `Tab` key, see above)
* Left/right arrows that go too far in either direction should simply cycle around to the first/last tab button as needed
* Home key puts focus on the first tab button
* End key puts focus on the last tab button

## Testing
__Please__ be sure the a11y test this locally using VoiceOver or Accessibility Inspector!

### DIFF
--- a/web/themes/new_weather_theme/templates/layout/page--local.html.twig
+++ b/web/themes/new_weather_theme/templates/layout/page--local.html.twig
@@ -133,35 +133,36 @@
 
         {{ drupal_block("weathergov_current_conditions") }}
 
-      <tabbed-nav class="display-block position-relative">
-        <div class="tab-buttons padding-bottom-4 padding-top-4 top-0 z-top display-flex flex-row flex-justify position-sticky bg-white top-4 mobile-lg:grid-col-8 desktop:grid-col-6">
+        <tabbed-nav class="display-block position-relative">
+          <div id="tablist-label" class="usa-sr-only">{{ "Weather information sections" | t }}</div>
+          <div role="tablist" class="tab-buttons padding-bottom-4 padding-top-4 top-0 z-top display-flex flex-row flex-justify position-sticky bg-white top-4 mobile-lg:grid-col-8 desktop:grid-col-6" aria-labelledby="tablist-label">
           {% if weather.alerts %}
-          <button role="button" class="tab-button" data-tab-name="alerts" aria-controls="alerts">{{ "Alerts" | t }}</button>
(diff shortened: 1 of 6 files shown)


web/themes/new_weather_theme/templates/layout/page--local.html.twig
```diff
@@ -133,35 +133,36 @@
 
         {{ drupal_block("weathergov_current_conditions") }}
 
-      <tabbed-nav class="display-block position-relative">
-        <div class="tab-buttons padding-bottom-4 padding-top-4 top-0 z-top display-flex flex-row flex-justify position-sticky bg-white top-4 mobile-lg:grid-col-8 desktop:grid-col-6">
+        <tabbed-nav class="display-block position-relative">
+          <div id="tablist-label" class="usa-sr-only">{{ "Weather information sections" | t }}</div>
```
eric-gade, line 137: @partly-igor Note this text here, which we talked about but perhaps didn't finalize

partly-igor: This is good!

greg-does-weather (approved):
This is excellent. One non-blocking comment.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability
- [X] functionality is tested with unit or integration tests
- [X] dependency updates in composer.json also got changed in composer-lock.json

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

partly-igor:
I tested with VO and keyboard and it was working great – only question I had was about the home/end presses, should we also prevent default on that? Right now it’ll both switch the tabs and scroll to top/bottom.

partly-igor:
I'll note from my conversation with Eric I don't have actual `home` and `End` keys, but instead I press `Fn + Arrow` on a mac keyboard, so it may not be exactly that.

PR #755, opened 2024-02-07
bump mem in staging and beta

loganmcdonald-noaa:
Should fix our deployment issues.


greg-does-weather (approved):
I don't know why our page load time tests have suddenly started failing sometimes, but this PR isn't the culprit.

loganmcdonald-noaa:
Holding off on this until we hear back from cloud.gov.

greg-does-weather:
Converting to draft so we don't accidentally merge it before then

PR #758, opened 2024-02-08
Add DAP to pages

### LINKED ISSUE #682: Implement DAP
Blocked by #497

greg-does-weather:
## What does this PR do? 🛠️

Adds the DAP script to all of our pages.

Question for @shadkeene! The [DAP implementation guide](https://github.com/digital-analytics-program/gov-wide-code/blob/master/documentation/GSA%20DAP%204.1%20-%20DAP%20Code%20Capabilities%20Summary%20and%20Reference.pdf) lists some configurations we must and can set. I only set the two listed below, but if you want to add more (or change the ones I've put in), it's easy to change now and in the future.

* agency: `NOAA`
* subagency: `NWS`
* sitetopic: `weather,safey`

Closes #682

### DIFF
--- a/tests/a11y/cypress/support/e2e.js
+++ b/tests/a11y/cypress/support/e2e.js
@@ -19,3 +19,7 @@ import "./commands";
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+// Ignore uncaught script errors. We can't control what happens with external
+// scripts, so...
+Cypress.on("uncaught:exception", () => false);
--- a/tests/load-times/cypress/support/e2e.js
+++ b/tests/load-times/cypress/support/e2e.js
@@ -1,0 +1,3 @@
+// Ignore uncaught script errors. We can't control what happens with external
+// scripts, so...
+Cypress.on("uncaught:exception", () => false);
--- a/tests/e2e/cypress/support/e2e.js
+++ b/tests/e2e/cypress/support/e2e.js
@@ -18,3 +18,7 @@ import "./commands";
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+// Ignore uncaught script errors. We can't control what happens with external
+// scripts, so...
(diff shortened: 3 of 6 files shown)


tests/a11y/cypress/support/e2e.js
```diff
@@ -19,3 +19,7 @@ import "./commands";
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+// Ignore uncaught script errors. We can't control what happens with external
+// scripts, so...
+Cypress.on("uncaught:exception", () => false);
```
greg-does-weather, line 25: This makes me sad, but for now, we get an unhandled script error from DAP which causes the Cypress tests to fail.

tests/load-times/cypress/support/e2e.js
```diff
@@ -0,0 +1,3 @@
+// Ignore uncaught script errors. We can't control what happens with external
+// scripts, so...
+Cypress.on("uncaught:exception", () => false);
```
eric-gade, line 3: Barely related: we should put in a test to ensure there is no logging to the browser console

eric-gade (approved):
lgtm 👍

tests/load-times/cypress/support/e2e.js
greg-does-weather: For our own code, we could have eslint throw an error for console statements (it currently throws warning, but I don't think we prevent deployments over warnings.)

For 3rd-party code (I guess USWDS and DAP?), should we also try to enforce a no-console-logging policy? I suppose it's worth doing now, and if we ever start getting log messages from some 3rd-party place, it'll force us to see it and decide what to do with it.

eric-gade: Good idea. Let's stick with eslint for now, that should catch us. But if we start bringing in frameworks with dependencies (like, say, a Vue application with all the fixins), we might want the e2e version

PR #760, opened 2024-02-08
Fix DAP

greg-does-weather:
## What does this PR do? 🛠️

The DAP script tag requires an ID. The lack of an ID was causing errors in the console log, but it also prevents DAP from working. This PR adds the required ID. It also reverts the test changes from #758, since those changes were only made to accommodate the errors that no longer show up. 😛


eric-gade (approved):
👍

PR #761, opened 2024-02-08
Update our linter to make console logging an error

greg-does-weather:
## What does this PR do? 🛠️

Logging to the browser console is not a great practice in production, so per @eric-gade's suggestion, let's console logging an error so our lints will fail over it.


eric-gade (approved):
👍

PR #762, opened 2024-02-09
Proxy bundles, take 2

greg-does-weather:
## What does this PR do? 🛠️

* The API proxy has a UI now. Check it out at [localhost:8081/](http://localhost:8081/)
* When you enable bundling, it'll just keep on bundling until you disable it. That's safer than the timing-and-hoping method I used before, but it does mean you can accidentally end up with a bunch of bundles.
* Moves the various TST recordings into a `testing` bundle and play it by default

A "bundle" is just a collection of api.weather.gov responses required to fulfill one beta.weather.gov request. If a bundle is playing and you request something that isn't represented by the bundle, it will be proxied on to api.weather.gov as normal. E.g., when you enable the `testing` bundle and visit `/local/OKX/34/34`, you will receive data from the API, ***not*** from the bundle because the `testing` bundle does not include data for OKX 34,34.

Long way to say: if you visit something not contained in the bundle, you'll get live data. Otherwise you'll get saved data.


eric-gade (approved):
The UI is awesome -- good idea!
  
Tested locally on a local rebase on main, and it's working
  
👍 
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #788, opened 2024-02-14
Add cache rules for static assets and fix-up library versions for cache-busting

### LINKED ISSUE #450: Investigate public/browser caching
### Description

We may need to look into public and browser caching to make sure our assets are cached in a way that's valuable for our users. For example, fonts should be aggressively cached because they're not likely to change very often. We want to make sure that public caches and user browsers don't need to pull duplicate data down very often because that's a waste of time and bandwidth.

Should we spend some time tweaking our Apache settings (via local `.htaccess` files, maybe?) to get the caching behavior we want? Should we look into putting a reverse proxy cache (e.g., Varnish) in front of Apache? Etc.

- Related to/derived from #260

### Acceptance Criteria
- [ ]

greg-does-weather:
## What does this PR do? 🛠️

- Adds web server directives to add cache headers to our static assets
  - Javascript and CSS are cacheable for 30 days. When we make changes to those assets, we need to bump the associated library versions in order to bust stale caches because Drupal will append the version as a query parameter. In reality, any given version of JS or CSS could be safely cached forever, provided we update the version information.
  - Other assets are cacheable for 7 days. Drupal automatically adds etags to these, so as long as we don't actually change them, they don't have to be transferred over the network. Drupal will return a 302 "Not Modified" response and browsers or intermediate caches will just use what's in the cache. When these assets change, they'll get new etags and that'll prompt browsers to fetch them anew.
- Updates our theme's `libraries.yml` file to include proper versions for everything. I opted to use datestamps because that should be good enough.

closes #450 

`make cc` should be enough, but man who knows

I also documented as best I could all the cache headers and stuff. https://docs.google.com/document/d/1feCG3r5xg0_oM11rY-siDuRViEdMHe6ggDwfmBnCbjo/edit#heading=h.x83vhri1gaqt


.github/workflows/update-theme-libraries.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - name: was the library file updated?
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set +e
+          git diff --exit-code origin/main web/themes/new_weather_theme/new_weather_theme.libraries.yml > /dev/null
+          if [ "$?" -ne 0 ]
+          then
+            gh pr comment ${{ github.event.number }} -b "This PR modifies theme Javascript or CSS assets but does not update the theme libraries file. Did you mean to update the appropriate version information in the libraries file?"
+          fi
```
greg-does-weather, line 28: No idea if this will work. It has to go into main before it'll run. Will no doubt be loads of fun to troubleshoot.

Dockerfile.dev
```diff
@@ -13,7 +13,7 @@ RUN echo "openssl.capath=/etc/ssl/certs/" >> /usr/local/etc/php/php.ini
 RUN set -eux; \
   \
   if command -v a2enmod; then \
-  a2enmod rewrite; \
+  a2enmod rewrite headers; \
```
greg-does-weather, line 16: mod_headers lets us rewrite headers. It is already enabled in our cloud environments.

loganmcdonald-noaa (approved):
LGTM! Thanks for the glossary/documentation page - super helpful. I wonder if we should find a way to link to it/add this context somewhere in our markdown docs in this repo?

.github/workflows/update-theme-libraries.yaml
elijah-wright: it should work even if the output goes to /dev/null, but I think you're testing if the file was unmodified, so I think it should be `-ne 1`. is there a reason to use -ne or could we use -eq to make it clearer?

elijah-wright:
if we need the theme libraries file to be modified with data that we can get from the system or wherever the version is (hopefully), then could we maybe use something like husky to modify the file in the event it wasn't modified already?

PR #805, opened 2024-02-21
Basic workbench install

loganmcdonald-noaa:
## What does this PR do? 🛠️
This setups our basic workbench configuration. We may add some things later. This installs, enables, and leverages the WFO taxonomy to setup both workbench and workbench access.


greg-does-weather (approved):
It's hard to review a bunch of config files. 😂 

We should schedule some team time to demo this. I don't see any good reason to hold it from staging, and that'd make for a nice tidy place to do it. Anyway, I'm excited! This sounds cool!

Issue #765, opened 2024-02-09
Log Observation Stations


colinmurphy01:
Based on our current logic for gathering observation stations, if data is missing from the first, we go to the next station in the list, then the third station.

We want to log

- How many point location page results use the first obs station, second, or third
- What is the difference in **distance** from the point to the observation station location
- What is the difference in **elevation** from the point to the observation station location

This will allow us to evaluate the quality of observation data at scale.

eric-gade:
Looks like #754 will be a prerequisite here

PR #806, opened 2024-02-21
Observation distance logging with NewRelic

eric-gade:
## What does this PR do? 🛠️
This PR, in part, implements the logging features outlined in #765 . It also implements additional changes we determined were needed as we went along, in particular the use of NewRelic as the logging endpoint for these distance metrics.
  
To summarize the primary task in the ticket: we want to log the number of stations a user had to cycle through until getting to a "valid" observation station, and to also log the distance from the user to that valid observation station.

### Distance calculation ###
Until we merge in the [point url PR](https://github.com/weather-gov/weather.gov/pull/769), we will not have a reliable point to measure from when getting the distance to the observation station. Instead, this PR implements a workaround: we take the WFO grid cell geometry, find the closest point in that geometry to the observation station, and then use the distance from that point.
  
Because an observation station could potentially be _inside_ of the WFO grid cell, we also include a boolean that tells us if the station is within its corresponding cell.
  
### NewRelic "Metric" integration ###
After some discussion, we decided to use the [NewRelic Metrics API](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/report-metrics-metric-api/) for recording these observation data.
  
To accomplish this, we've created a new class/service called `NewRelicMetrics` that is injected into the `WeatherDataService` as a dependency.
  
#### Guzzle and Async Requests ####
Our Drupal setup uses Guzzle as the HTTP client, and Guzzle has the ability to perform potentially non-blocking requests via its [async capabilities](https://docs.guzzlephp.org/en/stable/quickstart.html#async-requests) and the use of promises. We have implemented the `NewRelicMetrics` object in using these capabilities, in anticipation of future refactoring of all external service HTTP requests.
  
However, the current NewRelic calls are not _currently_ non-blocking -- we are explicitly calling a `wait()` method from within the `WeatherDataService`. In the future, we will want to use some kind of threaded queue object or pool to schedule all async requests. This call to `wait()` should be refactored at that point.
  
### "Stashed" weather data values ###
Because a single `WeatherDataService` instance is used per-request, we have added a property called `stashedGridGeometry` that does a sort of "cache" of any previously requested geometry values, since obtaining these values normally requires an extra API call. If the property is set, we simply use that value. If not, we fetch data for it.

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -85,12 +85,16 @@ class WeatherDataService
     private $database;
 
     /**
-
      * A cached version of any fetched alerts
      */
     private $stashedAlerts;
 
   /**
+     * NewRelic API handler
+     */
+    private $newRelic;
+
+    /**
      * Geometry of a WFO grid cell (stashed per request)
      *
      * @var stashedGridGeometry
@@ -107,12 +111,14 @@ class WeatherDataService
         RequestStack $r,
         CacheBackendInterface $cache,
         Connection $database,
+        NewRelicMetrics $newRelic,
     ) {
         $this->client = $httpClient;
         $this->t = $t;
         $this->request = $r->getCurrentRequest();
         $this->cache = $cache;
         $this->database = $database;
+        $this->newRelic = $newRelic;
 
         $this->defaultIcon = "nodata.svg";
         $this->defaultConditions = "No data";
@@ -609,9 +615,23 @@ class WeatherDataService
      */
     public function logObservationDistanceInfo($obsDistanceInfo)
     {
-        $logger = $this->getLogger("Weather.gov data service");
-        $serialized = json_encode($obsDistanceInfo);
-        $logger->info("[OBSERVATION]  " . $serialized);
+        $promise = $this->newRelic->sendMetric(
+            "wx.observation",
+            $obsDistanceInfo["distance"],
+            [
+                "withinGridCell" => $obsDistanceInfo["withinGridCell"],
+                "stationIndex" => $obsDistanceInfo["stationIndex"],
+                "obsStation" => $obsDistanceInfo["obsStation"],
+                "distance" => $obsDistanceInfo["distance"],
+            ],
+        );
+
+        try {
+            $promise->wait();
+        } catch (Exception $e) {
+            $logger = $this->getLogger("NEWRELIC");
+            $logger->error($e->getMessage());
+        }
     }
 
     /**
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -692,7 +692,7 @@ class WeatherDataService
 
         $obsStationIndex = 0;
         $observationStation = $obsStations[$obsStationIndex];
-        $obsDistanceInfoList = [];
+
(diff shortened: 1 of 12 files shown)


web/sites/default/settings.php
```diff
@@ -761,6 +761,9 @@
   'bower_components',
 ];
 
+// Add the newrelic endpoint for the Metrics API
+$settings["new_relic_rpm.metrics.base_url"] = "https://metric-api.newrelic.com/metric/v1";
```
loganmcdonald-noaa, line 765: Can we just put this in the cloudgov settings? New relic isn't relevant for local since we don't add it to the config.

loganmcdonald-noaa (approved):
I know I wasn't asked to approve, but approving for the mechanism/relationship with new relic!

web/sites/default/settings.php
eric-gade: Yeah that could work for environments. But it we ever want to add a new metric (and test it out first locally), we will want to be able to override in the settings here I think. Unless there is a better way to do that

loganmcdonald-noaa: Ok it's fine to leave it then!

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -491,6 +506,98 @@ public function getPlaceFromGrid($wfo, $x, $y, $self = false)
...
+            $obsDistanceInfo["distance"],
+            [
+                "withinGridCell" => $obsDistanceInfo["withinGridCell"],
+                "stationIndex" => $obsDistanceInfo["stationIndex"],
+                "obsStation" => $obsDistanceInfo["obsStation"],
+                "distance" => $obsDistanceInfo["distance"],
+            ],
+        );
+
+        $promise->wait();
```
greg-does-weather, line 598: I saw the comment about this, but I don't understand what purpose it serves. My naive take is it should be safe to fire-and-forget and it can finish whenever it finishes, but I'm not confident at all because of the nature of PHP. 😬

web/modules/weather_data/src/Service/Test/Base.php.test
```diff
@@ -114,6 +127,12 @@ abstract class Base extends TestCase
...
             $this->cacheMock,
             $this->databaseMock,
+            $this->newRelicMock,
         );
     }
+
+    protected function setUpNewRelicMock(): void
+    {
+        $promise = new Promise();
+    }
```
greg-does-weather, line 137: Is this vestigial?

greg-does-weather (approved):
Code all looks good to me. I don't think I have the right configuration to test it locally, but I'll get my environment setup later.

I hope the change from grid to point won't be a big lift!

web/modules/weather_data/src/Service/Test/Base.php.test
eric-gade: Oops!

web/modules/weather_data/src/Service/WeatherDataService.php
eric-gade: Yes, fire and forget would be the way to go. Unfortunately after hacking at it for a while I discovered that Guzzle [doesn't support that](https://stackoverflow.com/a/44774990). I'm thinking we will need to set up a queue that polls on a different thread and put all of our future async requests into it

Issue #804, opened 2024-02-21
Fire zone alerts are not displayed


greg-does-weather:
### Description

We currently filter alerts based first on geometry, then forecast zone, and finally SAME code. We need to also include alerts based on fire zone. Fire zone-based alert included in the attached bundle.

[albuquerque-no-red-flag.zip](https://github.com/weather-gov/weather.gov/files/14362965/albuquerque-no-red-flag.zip)

### Acceptance criteria

- [ ] Alerts based on fire zones show up correctly
- [ ] Tests have been implemented or modified to capture the bug

PR #807, opened 2024-02-21
Show alerts for a user based on their fire zone in addition to their forecast zone

greg-does-weather:
## What does this PR do? 🛠️

#804


eric-gade (approved):
LGTM 👍

PR #808, opened 2024-02-21
Show hourly (or six-hourly... whatevs) precipitation amounts

### LINKED ISSUE #796: Precipitation amounts table
Image

- [ ] Uses Quantitative precipitation field in /gridpoints
- [ ] Implemented on Today tab
- [ ] Table (not graph)

greg-does-weather:
## What does this PR do? 🛠️

- adds a `getHourlyPrecipitation` method to the data service, using the `/gridpoints/{wfo}/{x},{y}` endpoint
- adds an hourly precipitation block that puts the next 3 6-hour chunks of precipitation into a table
  - the first of these chunks may actually start in the past; I filter out the ones that ***end*** in the past
- removes the "hourly forecast" heading from the hourly forecast block and puts it in the page template, since where it sits may be independent of the hourly forecast block
- adds the hourly precipitation block to the today tab

> [!NOTE]  
> It might be worth holding this one until there's meaningful rain in someone's forecast so we can double-check that it's working? I feel good about it with the tests, but it'd be nice to see it working in the real world.

closes #796

image

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -686,6 +686,7 @@ class WeatherDataService
         foreach ($forecast->quantitativePrecipitation->values as $quantPrecip) {
             $valid = $quantPrecip->validTime;
             $value = $quantPrecip->value;
+            $value = $this->millimetersToInches($value);
 
             $valid = explode("/", $valid);
             $start = \DateTimeImmutable::createFromFormat(
@@ -700,7 +701,7 @@ class WeatherDataService
                 $periods[] = (object) [
                     "start" => $start->format("g A"),
                     "end" => $end->format("g A"),
-                    "value" => round($value * 0.03937008, 1),
+                    "value" => round($value, 1),
                 ];
             }
         }
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -678,7 +678,7 @@ class WeatherDataService
         $place = $self->getPlaceFromGrid($wfo, $gridX, $gridY);
         $timezone = $place->timezone;
 
-        $forecast = $this->getFromWeatherAPI("/gridpoints/$wfo/$gridX,$gridY")
+        $forecast = $self->getFromWeatherAPI("/gridpoints/$wfo/$gridX,$gridY")
             ->properties;
 
         $periods = [];
--- a/web/modules/weather_data/src/Service/WeatherDataService.php
+++ b/web/modules/weather_data/src/Service/WeatherDataService.php
@@ -657,6 +657,57 @@ class WeatherDataService
         ];
     }
 
+    public function getHourlyPrecipitation(
+        $wfo,
+        $gridX,
+        $gridY,
+        $now = false,
+        $self = false,
+    ) {
+        date_default_timezone_set("America/New_York");
+
+        if (!$self) {
+            $self = $this;
+        }
+
+        $wfo = strtoupper($wfo);
+        if (!($now instanceof \DateTimeImmutable)) {
+            $now = new \DateTimeImmutable();
+        }
+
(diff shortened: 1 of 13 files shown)


greg-does-weather:
Holding this until I can walk through it with a screen reader. Want to make sure the table makes sense.

@partly-igor Should this be styled or is the ticket currently just about getting the table in place and we'll iterate later? If there should be styling, I'll schedule us some pairing time to get it right!

partly-igor:
This one is about getting it in place, we can handle styling in a different one.

mgwalker:
Alright, I'm pretty happy with how the table is read. It announces itself as a table, describes the columns, and as you navigate through the table, it reads:

"time period, 6 am, 12 pm" and the next column "forecast precipitation 0 inches"

I will be very eager to see how a screen reader user feels about this, but when I closed my eyes and just listened, I felt like I got a pretty good sense of what the table was trying to convey.

greg-does-weather:
I concur with myself.

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -657,6 +657,57 @@ public function getCurrentConditionsFromGrid($wfo, $gridX, $gridY)
...
+
+        $wfo = strtoupper($wfo);
+        if (!($now instanceof \DateTimeImmutable)) {
+            $now = new \DateTimeImmutable();
+        }
+
+        $place = $self->getPlaceFromGrid($wfo, $gridX, $gridY);
+        $timezone = $place->timezone;
+
+        $forecast = $self->getFromWeatherAPI("/gridpoints/$wfo/$gridX,$gridY")
```
eric-gade, line 681: Is this the only place where we currently call this API route? If not, we should consider adding a stashed variable on the data service instance and checking that first

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -657,6 +657,57 @@ public function getCurrentConditionsFromGrid($wfo, $gridX, $gridY)
...
+            )->setTimeZone(new \DateTimeZone($timezone));
+
+            $duration = new \DateInterval($valid[1]);
+            $end = $start->add($duration);
+
+            if ($end >= $now) {
+                $periods[] = (object) [
+                    "start" => $start->format("g A"),
+                    "end" => $end->format("g A"),
+                    "value" => round($value * 0.03937008, 1),
```
eric-gade, line 703: What's the constant we are using here?

eric-gade (approved):
It's working! 👍 
  
Just a couple of inline comments, plus the following:
  
1. Do we want to check-in a test bundle (I have recorded one just in case).
2. Do we want to display rows in the table when the value computes to zero? See this example (tagging @partly-igor too):
Screen Shot 2024-02-26 at 10 19 30 AM


## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

web/modules/weather_data/src/Service/WeatherDataService.php
greg-does-weather: Good catch. This is the conversion from mm to inches. It should be factored into a constant and/or put into the unit conversion trait thingy. Will do that!

web/modules/weather_data/src/Service/WeatherDataService.php
greg-does-weather: It's not the only place. I think we use this particular endpoint a few times. Good call on pulling it out and storing it in-memory until the end of the response. I think that will also fit well with the testing changes I'm proposing, as it will help delineate a "data layer" vs. a processing layer.

greg-does-weather:
Re: test bundle, good idea! We can add it to an existing point if we want to, but if you've already got one saved, feel free to drop it here and I'll get it added. I should also add an end-to-end test.

PR #828, opened 2024-02-27
Add snow squall warning to land hazards

greg-does-weather:
## What does this PR do? 🛠️

Snow Squall Warning is a fairly new alert type. It is present in our prioritized list of hazards but it is not included in the list of land hazards. This PR adds it.


eric-gade (approved):
Quick approval without checklist here since it's small. Not sure what's going on with these page load tests though

PR #845, opened 2024-02-28
Refactor external data access into a data layer

### LINKED ISSUE #839: Refactor WeatherDataService to extract a distinct data layer
### Description

Part of our [refactored testing strategy](https://docs.google.com/document/d/1l2SgLRqfuBDpp8P_cY1VgloSJnm8yQMmYr2T2rZgEGE/edit#heading=h.i40i4ikjxrlr). This step will help us isolate what we can and cannot reliably test and should make it easier for further refactoring to get to our goal.

greg-does-weather:
## What does this PR do? 🛠️

- Creates a `weather_data_layer` service (`DataLayer` class) and injects it into the WeatherDataService
- Adds a set of methods dedicated solely to fetching data from external sources
- Removes WeatherDataService dependencies on cache, database, and HTTP client
- Updates methods to fetch external data from the data layer
- Removes some old, unused code
   - A big chunk of code that I yoinked was around preserving the place name from search results. It's been broken for a long while and we couldn't use it the way it was originally implemented. I think we should put it back, but it can come after the refactoring so we'll have a better sense of where it should go.

closes #839 

All the unit tests are gonna fail pretty hard. 😅


web/modules/weather_data/src/Service/DataLayer.php
```diff
@@ -0,0 +1,245 @@
...
+            $this->iPlaceNearPolygon[$wktPoints] = $this->getPlaceNear(
+                "POLYGON(($wktPoints))",
+            );
+        }
+        return $this->iPlaceNearPolygon[$wktPoints];
+    }
+
+    public function databaseFetch($sql)
+    {
+        return $this->database->query($sql)->fetch();
```
eric-gade, line 243: Thinking aloud: where do we want to catch and handle DB/query errors, should they emerge? I think in this state of the code, a given block `build()` method is the place that would catch these for now (and that might be the appropriate place)

eric-gade (approved):
mr-clean
  
Nice and clean!
  
I pushed just one commit to fix some leftover old method calls that prevented the hourly forecast block from rendering. Everything else looks good 👍

web/modules/weather_data/src/Service/DataLayer.php
```diff
@@ -0,0 +1,245 @@
...
+        // headers to the API. If we've already gotten an ID for this response,
+        // keep it.
+        $this->responseId = uniqid();
+    }
+
+    /**
+     * Get data from the weather API.
+     *
+     * The results for any given URL are cached for the duration of the current
+     * response. The cache is not persisted across responses.
```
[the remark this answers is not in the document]
greg-does-weather: This is not true. I'll update the comment. The cache TTL is set to 60 seconds. Good catch! We have lots of comments that aren't quite true anymore I'm finding. 😅

web/modules/weather_data/src/Service/DataLayer.php
greg-does-weather: Added #871 so this question doesn't have to just live in your head. 😄

PR #846, opened 2024-02-28
When saving bundles, save API times as relative to the time the bundle is captured

greg-does-weather:
## What does this PR do? 🛠️

When we record bundles, we currently capture the data exactly as provided by the API. As a result, data in bundles can slowly become outdated and no longer displayed by the site. This PR updates timestamps as it records them so that we preserve the relative time from "now" instead of an absolute time. This way, when we replay a bundle, the data presented should be identical to when it was captured, regardless of when it's played back (except for the timestamps, of course, which will be computed as a time relative to when the page is loaded).

### DIFF
--- a/tests/api/serve.js
+++ b/tests/api/serve.js
@@ -20,6 +20,12 @@ const exists = async (file) => {
 };
 
 const processDates = (obj) => {
+  // If the input is null, just bail out. Otherwise we'll accidentally turn it
+  // into an object.
+  if (obj === null) {
+    return;
+  }
+
   const now = dayjs();
 
   // Recursively search through the object to find all values that have the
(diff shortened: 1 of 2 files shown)


tests/api/serve.js
```diff
@@ -48,11 +54,11 @@ const processDates = (obj) => {
         obj[key] = now.format();
       }
 
+      // If there's a duration component, smoosh it on to the end of our
```
eric-gade, line 57: Any chance to smoosh I'll take

eric-gade (approved):
This is working well for me. One comment though: I think ultimately we _will_ have to align daily forecast period `startTime` and `endTime` values to the nearest hour. It will make e2e testing the alignment of alerts from being croissant level flaky

eric-gade:
@greg-does-weather I actually went ahead and tried to implement the hourly-alignment that I suggested above. I've pushed it to a separate branch/commit, which you can find here:
https://github.com/weather-gov/weather.gov/commit/177eb54469878f579f22b44b4aa056c995135fab

Issue #795, opened 2024-02-16
Apply styling


partly-igor:
Igor to provide assets

The high fidelity designs need to be implemented, applying styles to the components. 

- [x] Update uswds color settings
- [x] Update uswds typography settings
- [x] Match layout and style updates to designs for 
  - [x] Alerts tab
  - [x] Current tab
  - [x] Today tab
  - [x] Daily tab

Issue #792, opened 2024-02-16
Updating layout of Hourly to table


colinmurphy01:
![Screenshot of hourly table layout](https://github.com/weather-gov/weather.gov/assets/150970973/b23de56a-879b-4e91-9f23-4170d398e50a)

[Artboard](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=3765%3A26081&mode=design&t=e8AR2Vxv9mFLeRgJ-1)

[Protoype](https://www.figma.com/proto/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?page-id=3765%3A26060&type=design&node-id=3765-26081&viewport=654%2C935%2C0.41&t=SpafpQ9QMYUhEIXW-1&scaling=scale-down&mode=design) - shows scroll, but not arrow click functionality. Let's talk about how to implement. 

[Annotated Component](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=3765%3A26373&mode=design&t=e8AR2Vxv9mFLeRgJ-1)

What's included

- [x] Layout of hourly data should change from individual divs/items into a combined table
- [x] Includes time, temp, ~feels like~, condition, wind speed, ~gusts~, humidity, dew point
- [x] Columns are hours
- [x] Rows are weather data/values
- [x] If hazard is present, it will be visible in table across the hours it is in effect
- [x] Add right/left buttons to move table left and right (pair with igor on how much it would jump/scroll to)
- [x] First column is frozen as user scrolls right to left

Not included
- Gusts
- Apparent Temperature

Does adding multiple variables add significant complexity? Would it be easier if we used placeholder/dummy data for ay of these for testing? Would simplifying the variables down make it easier for now?

Separate:
- #795

colinmurphy01:
Current API doesn't have gusts and apparent temperature, so we'd leave them out.

eric-gade:
Closed by #847

Issue #841, opened 2024-02-27
Refactor unit tests


greg-does-weather:
### Description

Following #840, we need to add unit tests to the now unit-sized processing functions in the WeatherDataService.

PR #847, opened 2024-02-28
Hourly forecast table component

eric-gade:
## What does this PR do? 🛠️
This PR implements the changes outlined in #792, and places the hourly forecast data into a table with dynamic scrolling features.
  
See the [(1)](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=3765-26060&mode=design) and [(2)](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=3765-26060&mode=design) figmas for the visuals.

### Styling comes later ###
There is a separate ticket (#795 ) that will deal with finishing the actual styling of this component.

### New component: `hourly-table` ###
Mainly due to the addition of the dynamic scrolling buttons, we have wrapped the table in a custom element called `hourly-table`. This component handles the corresponding events, and also the calculation of how much to scroll in either direction when a scroll button is clicked. This calculation is based on the width of a td element in the last row of the table.
  
There are some assumptions about numbers in the scrolling calculations that will need to be adjusted in #795 when styling is revisited for this component.
  
### Alerts in the table ###
Aligning the relevant alerts as table cells that span across the corresponding time columns required some hefty code changes and testing additions, which you will see clearly in the diff.
  
The architecture is such that the HourlyForecastBlock will now receive an array with two properties, one being a list of "alertInfo" arrays. Each of these arrays has the following information in it:
| key | description of value |
|----|---------------------|
| alert | A reference to the original (parsed) alert object used throughout the location templates |
| periodIndex | The index of the time period (each hour) that the alert should begin at. This represents the number of `td` elements that should precede the beginning of the alert `td` element. |
| duration | This integer roughly corresponds to the duration of the alert in hours, truncated as needed to fit either end of the overall hourly period being considered. This number also corresponds directly to the `colspan` of the alert `td`|
  
#### Alert display decisions ####
We had to make a few assumptions along the way, some of which might need to change in the future and so are worth noting here.
1. If an alert has no `ends` value, we calculate the "end time" of that alert to be the `expires` value. If that is also null, then we simply ignore the alert for the purposes of displaying it in the hourly table. We might want to, instead, use the last hourly period's end time as this "end time" instead, so that  the given alert covers the whole rest of the period.
2. We do not align these alert spans based on minutes after a given hour. If an alert starts at any point between the beginning of an hourly period and the beginning of the next hourly period, it will be displayed as starting from the first period's column, and not offset at all to indicate extra minutes. In other words, an alert that begins at 11:23 will have its cell span appear to begin right at 11:00
3. We round up alert minutes when calculating the duration. If an alert is 5 hours and 23 minutes, we calculate it as having 6 hours and a duration of 6.

Screen Shot 2024-02-28 at 1 57 42 PM


web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -190,6 +198,259 @@ public function getAlerts($grid, $point, $self = false)
...
+        $alertPeriods = [];
+
+        foreach ($relevantAlerts as $currentAlert) {
+            $onsetTime = self::turnToDate(
+                $currentAlert->onsetRaw,
+                $currentAlert->timezone,
+            );
+            $endTime = $this->getEndTimeForAlert($currentAlert);
+            if (!$endTime) {
+                continue; // pass to the next alert, ignoring this one
```
greg-does-weather, line 257: @shadkeene @coreypieper @colinmurphy01  Do we need to add a ticket to decide how to address alerts that doesn't have an expiration or end time?

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -190,6 +198,259 @@ public function getAlerts($grid, $point, $self = false)
...
+            }
+            $alertEncompassesPeriods = $this->dateTimeEncompasses(
```
greg-does-weather, line 259: I love this method name. Very descriptive!

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -190,6 +198,259 @@ public function getAlerts($grid, $point, $self = false)
...
+            $onsetIsWithinPeriod = $this->dateTimeIsWithin(
+                $alertOnset,
+                $periodStartTime,
+                $periodEndTime,
+            );
+            $endIsWithinPeriod = $this->dateTimeIsWithin(
+                $alertEnd,
+                $periodStartTime,
+                $periodEndTime,
+            );
```
greg-does-weather, line 410: Am I misreading, or is this not being used?

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -190,6 +198,259 @@ public function getAlerts($grid, $point, $self = false)
...
+     * If null, we try the `expires` field.
+     * If both are null, return false.
+     */
+    private function getEndTimeForAlert($alert)
+    {
+        // We need to determine the correct ends field.
+        // If there is no value for the ends, then we use
+        // expires. If there is in that case no value for expires,
+        // then we do nothing and continue to the next alert,
+        // ignoring this one.
```
greg-does-weather, line 445: @shadkeene @coreypieper @colinmurphy01 Another product question here, I think: should we have a ticket to determine what to do with alerts in the hourly forecast table when they don't have an `end` field but do have an `expires` field? My understanding is such an alert continues indefinitely but the expiration is there to prompt forecasters to update it; is that right?

web/modules/weather_data/src/Service/WeatherDataService.php
```diff
@@ -892,4 +902,12 @@ public function getDailyForecastFromGrid(
...
         ];
     }
+
+    /**
+     * Convert C to F
+     */
+    private function celciusToFahrenheit(float $degreesC)
+    {
+        return $degreesC * 1.8 + 32;
+    }
```
greg-does-weather, line 912: A potential future refactor here as well, in that we should be able to replace this with the UnitConversionTrait that handles reading the input units from the data.

So instead of:
```php
"dewpoint" => $this->celsiusToFahreneith($period->dewpoint->value)
```

you'd have:
```php
"dewpoint" => $this->getTemperatureScalar($period->dewpoint)
```

And that should work even if the API suddenly started serving Fahrenheit.

web/themes/new_weather_theme/templates/partials/alerts-in-hourly-table.html.twig
```diff
@@ -0,0 +1,20 @@
+{% set periodLength = periods | length %}
+{% for alertInfo in alertPeriods %}
+  {% set remainingCount = periodLength - alertInfo.duration - alertInfo.periodIndex %}
+  <tr data-row-name="alert">
+    <th scope="row" class="position-sticky left-0"><span class="usa-sr-only">{{ "Alert" | t }}</span></th>
+    {% if alertInfo.periodIndex > 0 %}      
+      {% for i in 1..alertInfo.periodIndex %}
+        <td></td>
+      {% endfor %}
```
greg-does-weather, line 9: Could this be done with just a colspan? (Genuine question, I dunno if a colspan would screw something up.)

```suggestion
      <td colspan="{{ alertInfo.periodIndex }}"></td>
```

web/themes/new_weather_theme/templates/partials/alerts-in-hourly-table.html.twig
```diff
@@ -0,0 +1,20 @@
...
+    {% endif %}
+    <td colspan="{{alertInfo.duration}}" style="background-color:yellow;">
+      {{alertInfo.alert.event}}
+    </td>
+    {% if remainingCount > 0 %}
+      {% for i in 1..remainingCount%}
+        <td></td>
+      {% endfor %}
```
greg-does-weather, line 17: If the other colspan would work, I reckon this one could too?

```suggestion
      <td colspan="{{ remainingCount }}"></td>
```

greg-does-weather: This is making me think we should also create an accessibility testing ticket for this table. I'll run it through a screen reader later because I'm really curious what that experience is like.

greg-does-weather (approved):
Woo. No wonder this gave you a headache. Time math...
math

Anyway, I think this looks good. And what you've done with the unit tests is where #841 should eventually lead us, so I'm glad to have these as a reference!

There are some comments in here, but they're future debt/refactor issues and a couple of product questions. Merge at your leisure!

greg-does-weather:
Also clearly need to have a look at that "did you update the libraries file" thing. 😂

web/modules/weather_data/src/Service/WeatherAlertTrait.php
colinmurphy01: We should be using `eventEndingTime` (over `end`). That is the field in CAP and the field the current website relies on.

And yes, using #852 to track

web/modules/weather_data/src/Service/WeatherAlertTrait.php
colinmurphy01: yes, using #852 to track

web/modules/weather_data/src/Service/WeatherAlertTrait.php
greg-does-weather: To clarify, the API converts `eventEndingTime` from CAP into just `end`. We don't know why they rename it, but we've confirmed that's what happens.

web/modules/weather_data/src/Service/WeatherAlertTrait.php
```diff
@@ -190,6 +198,259 @@ public function getAlerts($grid, $point, $self = false)
...
+        $relevantAlerts = array_filter($alerts, function ($alert) use (
+            &$periods,
+            &$lastPeriodEndTime,
+        ) {
+            $onsetDateTime = self::turnToDate(
+                $alert->onsetRaw,
+                $alert->timezone,
+            );
+            return $onsetDateTime < $lastPeriodEndTime;
+        });
```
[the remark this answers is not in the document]
eric-gade: Yes, and now that I look at it, the check is incomplete. We want to also make sure that the alert end time is greater than the overall period start time, too. Fixed in https://github.com/weather-gov/weather.gov/pull/847/commits/f7f1d87aebaeef797a275ed431079ca75ca44429

web/modules/weather_data/src/Service/WeatherAlertTrait.php
eric-gade: Good catch! Nixed in https://github.com/weather-gov/weather.gov/pull/847/commits/f7f1d87aebaeef797a275ed431079ca75ca44429

web/modules/weather_data/src/Service/WeatherAlertTrait.php
colinmurphy01: Yes we chatted to them about it. It was originally built to support the OG weather.gov legacy site. It doesn't serve a purpose anymore.

We've been told to rely on `eventEndingTime`. If you all feel it's not a risk to rely on `end` knowing that, than proceed.

web/themes/new_weather_theme/templates/partials/alerts-in-hourly-table.html.twig
eric-gade: This is...so obvious....

web/modules/weather_data/src/Service/WeatherAlertTrait.php
greg-does-weather: We'll stick with `end` for now because it's a top-level property. We have to dig deeper into the data structure to get to `eventEndingTime` and the only difference between the two is that one of them is in UTC and the other includes a UTC offset, but they report the exact same time.

PR #849, opened 2024-02-28
Add more USWDS settings and style current tab

partly-igor:
## What does this PR do? 🛠️
This PR starts the styling work for #795. It adds additional USWDS settings:
- Primary, secondary, accent, and base colors
- Typography scales
- Monospace font token

Additionally it applies those styles to the location, alerts, and tabs components, as well as the current conditions tab content. It does not restyle the alerts, daily, or hourly tab content. That work will be completed in follow on PRs.

I'm unable to run linters and e2e so would appreciate the reviewer running that locally. 

screenshot of current tab at desktop width

screenshot of current at mobile width


eric-gade (approved):
Code LGTM 👍 
  
One issue I noticed: the alert accordions are displaying at half-width on certain desktop sizes:
Screen Shot 2024-02-28 at 5 52 00 PM
  
**EDIT**: Hah, I think that's the cause of the failing CI right now, actually

partly-igor:
Thank you! removed and it definitely looks less wonky, but still fails the test. I think the test doesn't quite match up with the design anymore – there is no story in the current conditions which is what the test was originally testing against. I am wondering if we should modify the test, but need to check that I am understanding the two widths it's comparing correctly.

greg-does-weather:
Yep, that test was for a visual bug in the conditions narrative, but the narrative is no longer shown visually so the test isn't useful. The test still passed because the narrative is still in the DOM and still has width, but... we don't actually care about the width since it's invisible. 😂

PR #850, opened 2024-02-28
Setup the api proxy to work in cloud.gov

loganmcdonald-noaa:
## What does this PR do? 🛠️
This deploys our api proxy endpoints for playing and stopping bundles to cloud.gov (https://weathergov-api-proxy.app.cloud.gov/play/testing). This allows us to use bundles creating locally to fake data in cloud.gov which is useful for showing off certain behavior during tests or to each other. Pretty neat! Thank you @greg-does-weather!

We've removed the endpoint for creating bundles so this can only be used to play or stop bundles that have been previously created. It will take coordination between design and engineering to create new bundles for desired scenarios. 

We had previously talked about making this an internal link between our two cloud.gov accounts but don't see a threat model for interaction with the api proxy at this point, and we would like to keep the UI. This is only enabled in weathergov-design and no other environments. 

Example of [many alerts](https://weathergov-design.app.cloud.gov/point/34.749/-92.275): 

Screenshot 2024-02-28 at 12 18 39 PM

### DIFF
--- a/manifests/manifest-design.yaml
+++ b/manifests/manifest-design.yaml
@@ -9,6 +9,7 @@ default_config: &defaults
   env:
     PHP_INI_SCAN_DIR: /home/vcap/app/php/etc/:/home/vcap/app/php/etc/php.ini.d/
     NEWRELIC_LICENSE: ((newrelic-license))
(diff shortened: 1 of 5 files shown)


greg-does-weather (approved):
cool beans

PR #857, opened 2024-03-01
Style alert details

partly-igor:
## What does this PR do? 🛠️
This PR add styling to the alert detail accordion to match the current design, as part of the work on #795. 

In order to keep the content in the Today and daily tab looking more presentable, I've added a container around that content but otherwise did not style it. These containers are temporary and will need to adjusted once styling is applied to the content of these two tabs. 

Before
Screenshot of alert before changes

After
Screenshot of alert before changes


greg-does-weather (approved):
Spiffy

PR #859, opened 2024-03-01
Get hourly forecast data from gridpoint endpoint

### LINKED ISSUE #853: Switch hourly forecast to /gridpoints endpoint instead of /gridpoints/[id]/forecast/hourly
### Description

In order to implement #793, we first need to switch out how we fetch the hourly forecast from the API. This will require somewhat complicated processing of the data to turn it into the format we expect (and is serving us well at the moment). We began a little bit of this work in #796.

Once this is done, then #793 will be ready to go.

greg-does-weather:
## What does this PR do? 🛠️

- updates how we fetch hourly forecast to use the `/gridpoints` endpoint so we can access the data properties there
- merges in data from the `/gridpoints/forecast/hourly` endpoint as well, since it's not all available in either place... :sigh:

closes #853


web/themes/new_weather_theme/templates/block/block--weathergov-hourly-forecast.html.twig
```diff
@@ -60,8 +60,8 @@
             {% for period in content.hours %}
               <td style="white-space:pre;">
                 <div class="display-flex flex-row flex-align-center">
-                  <span>{{period.windSpeed}}</span>
-                  <svg role="img" aria-hidden="true" class="width-2 height-2 margin-left-05 padding-2px" data-wind-direction="{{period.windDirection}}">
+                  <span>{{period.windSpeed}} {{ "mph" | t }}</span>
+                  <svg role="img" aria-hidden="true" class="width-2 height-2 margin-left-05 padding-2px" style="transform: rotate({{ period.windDirection.angle + 180 }}deg);">
```
greg-does-weather, line 64: The +180 is to account for wind direction being reported as *from*, but the arrow representing *to*.

web/themes/new_weather_theme/templates/block/block--weathergov-hourly-forecast.html.twig
```diff
@@ -60,8 +60,13 @@
...
                 <div class="display-flex flex-row flex-align-center">
-                  <span>{{period.windSpeed}}</span>
-                  <svg role="img" aria-hidden="true" class="width-2 height-2 margin-left-05 padding-2px" data-wind-direction="{{period.windDirection}}">
+                  <span>{{period.windSpeed}} {{ "mph" | t }}</span>
+                  {#
+                    The +180 here is to account for the fact that wind direction
+                    is reported as a FROM direction, but our arrow points in the
+                    TO direction. So we just need to spin it 'round.
+                  #}
+                  <svg role="img" aria-hidden="true" class="width-2 height-2 margin-left-05 padding-2px" style="transform: rotate({{ period.windDirection.angle + 180 }}deg);">
```
eric-gade, line 69: I'm cool with having it inline like this for readability, but the stylesheet was designed to read from the `data-wind-direction` attribute and apply the correct rotation already ([see here](https://github.com/weather-gov/weather.gov/blob/main/web/themes/new_weather_theme/assets/sass/_uswds-theme-custom-styles.scss#L123-L151))

eric-gade (approved):
LGTM 👍 
  
Only one non-blocking comment about wind direction styling. If we do it inline we should remove the attribute styles from the sass sheet
  
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security
- [x]data from external sources is cleaned and clearly marked

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/themes/new_weather_theme/templates/block/block--weathergov-hourly-forecast.html.twig
greg-does-weather: Understood. This code uses the actual angle rather than just the 8 cardinal/ordinal directions. I don't feel strongly about which we should use, though.

greg-does-weather:
I did mean to, bot, thank you!

Issue #838, opened 2024-02-27
Disable the >=90% code coverage requirement


greg-does-weather:
### Description

This is a necessary first step for our testing refactoring plan. [See the testing proposal document](https://docs.google.com/document/d/1l2SgLRqfuBDpp8P_cY1VgloSJnm8yQMmYr2T2rZgEGE/edit#heading=h.i40i4ikjxrlr) for more.

This is only temporary. The >=90% code coverage requirement will be reinstated when we get our end-to-end tests instrumented for coverage.

Here's the sequence of these issues:
![image](https://github.com/weather-gov/weather.gov/assets/142943695/517e9f5c-578b-4544-a9b3-bf1c4e3fc508)

### Acceptance criteria

- [ ] Pull requests are not blocked by low code coverage

Issue #842, opened 2024-02-27
Update end-to-end testing to capture backend code coverage


greg-does-weather:
### Description

In order to make our tests more robust and reliable, our end-to-end tests should compute code coverage. Some prototyping around this work has been started in the `mgwalker/change-testing` branch. This will help us identify the test gaps that end-to-end testing should cover.

### Acceptance criteria

- [ ] End-to-end tests report code coverage
- [ ] The CI/CD code coverage reporter step properly integrates the coverage information from end-to-end tests

colinmurphy01:
By the end of this sprint we'd close #843 and then have succeeding work clarified for closing additional gaps.

PR #861, opened 2024-03-02
838 / Remove 90% test coverage requirement from github

Kyle01:
## What does this PR do? 🛠️
* Disable the >=90% code coverage requirement. This is a PR to satisfy [issue 838](https://github.com/weather-gov/weather.gov/issues/838)
* In the future, PR can be reverted easily for ticket 844

* See below github test suite proving that 90% coverage test has been disabled 

Screenshot 2024-03-03 at 9 54 52 AM

### DIFF
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -162,6 +162,7 @@ jobs:
     name: PHP unit tests
     runs-on: ubuntu-latest
     needs: [should-test]
+    if: false
 
     steps:
       - name: checkout
--- a/.github/workflows/code-standards.yaml
+++ b/.github/workflows/code-standards.yaml
@@ -215,13 +215,13 @@ jobs:
 
     steps:
       - name: get coverage output
-        if: needs.should-test.outputs.yes == 'true'
+        if: false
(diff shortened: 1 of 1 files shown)


greg-does-weather:
Hi @Kyle01, and welcome to the repo. 🙂 Thanks for your contribution! This is a nice, clean solution. I think what we're going to do is change our repo setting so the code coverage test is not required for merging, but we'll keep running it so we've got an in-our-face reminder that we need to address it.

If you want to make a similar PR that disables unit tests, we'll need to do that as part of #842. It'll probably linger open for a few days, but we'd incorporate it when we get there!

Kyle01:
@greg-does-weather Thanks for the quick reply and consideration. That makes sense to me, I'll be on the lookout for other things I can contribute to in the future. I'm closing the PR.

greg-does-weather:
Our timing is spectacular because I was just coming back to accept this PR. I realized what I proposed earlier doesn't actually work – we really do need to disable these tests, along with the unit tests. 🤦‍♂️ And this is a nice solution that keeps the tests showing up in our checks list.

greg-does-weather (approved):
Thank you.

PR #870, opened 2024-03-07
Refactor WeatherDataService

greg-does-weather:
### Traits

- Renames `WeatherAlertTrait` to just `AlertTrait`, since the "weather" part is heavily implied 😛
- Moves daily forecast and current conditions into traits

### Utilities

- Renames `AlertPriority` to `AlertUtility` to better reflect its purpose
- Adds `DateTimeUtility` for... date time stuff
  - `filterToAfter` takes an array, a `DateTimeImmutable`, and (optionally) the name of the field to take the data from (this defaults to `startTime`). Returns the items from the array that are before the date. Resets the array index because I am not a monster.
  - `filterToBefore` does the same, but for before instead of after. Probably didn't need to say that.
  - `stringToDate` takes a string representation of a datetime and, optionally, a timezone, and returns a `DateTimeImmutable` for that string, optionally with the timezone set
- Moves `UnitConversionTrait` into a `UnitConversion` class with static methods, matching the other utility classes
  - Adds a method for length conversions

### Data layer
- makes stashed i-vars static because for some reason our services are constructed twice on each request – once for the preloader and then again for the blocks. As a result, anything we stash off in preloading is lost when the blocks fetch data. Making the stashed i-vars static protects against this.


web/modules/weather_data/src/Service/DataLayer.php
```diff
@@ -107,49 +107,49 @@ private function getFromWeatherAPI($url, $attempt = 1, $delay = 75)
         }
     }
 
-    private $i_alertsState = false;
+    private static $i_alertsState = false;
```
eric-gade, line 110: I've been trying to look around to double check, but are the classes themselves initialized per-request?

eric-gade (approved):
👍   

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_data/src/Service/DataLayer.php
greg-does-weather: This one and WeatherDataService are constructed twice per request. It is not at all clear to me why that is, but once I saw it was happening, I figured static i-vars was a way around it. It seems to be working.

PR #875, opened 2024-03-11
Fix a bug with county-zone alerts

greg-does-weather:
## What does this PR do? 🛠️

Fixes a bug where we filtered out alerts whose `affectedZones` property includes our point's county zone. County zones should be treated equivalently to forecast and fire zones.


eric-gade (approved):
👍 

## Code Review Checklist
## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Reliability


## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #882, opened 2024-03-13
Fix wind speed display when the value is zero.

greg-does-weather:
## What does this PR do? 🛠️

When the wind speed is 0, we display `N/A`. That is incorrect. PHP thinks `null` and `zero` are equal. This PR adds strict equality checking to fix that bug.

[reference](https://gsa-tts.slack.com/archives/C05DU11D4J2/p1710347151677179)


eric-gade (approved):
Looks good and working on my end 👍

PR #887, opened 2024-03-14
Hardcode last update timestamp to MDT for R3

### LINKED ISSUE #883: Turn off current conditions timestamp for testing
we are using javascript to update the current conditions timestamp to the browser’s local time

- [ ] Turn it off in the `/testing` branch for R3 next week
- [ ] Turn it back on after R3 is completed?

greg-does-weather:
## What does this PR do? 🛠️

closes #883 

image


mgwalker:
These test failures are expected.
- JS lint test is failing because of the scripts for converting CSVs into test data
- end-to-end tests are failing because we're not actually doing the timezone conversion

eric-gade:
> These test failures are expected.
> 
>     * JS lint test is failing because of the scripts for converting CSVs into test data
> 
>     * end-to-end tests are failing because we're not actually doing the timezone conversion

Let's just add a comment to ignore linting in those loader files, and maybe skip the e2e cases that are failing here for now

eric-gade (approved):
👍

PR #889, opened 2024-03-14
Move test site from Billings, MT, to Red Lodge, MT

greg-does-weather:
## What does this PR do? 🛠️

Just what it says. Only for the R3 branch.


partly-igor (approved):
Thank you!

PR #890, opened 2024-03-14
Have the API proxy default to serving the R3 bundle when it starts up

greg-does-weather:
## What does this PR do? 🛠️

This way when we redeploy, or the proxy restarts for some reason, it will restart by serving the R3 bundle.


partly-igor (approved):
Smart!

PR #891, opened 2024-03-14
Add hardcoded screenshots for alerts in R3

greg-does-weather:
## What does this PR do? 🛠️

Screenshot 2024-03-14 at 14-16-31 Red Lodge MT Weather gov National Weather Service


greg-does-weather:
I guess the screenshot is funky because I was scrolled down a bit and the tabs were stickied to the top of my viewport.

PR #894, opened 2024-03-15
Add radar gif for r3

partly-igor:
## What does this PR do? 🛠️
Adds a a hardcoded gif of the radar near Red Lodge, MT / Billings, MT for R3 testing. 

Screenshot of the radar gif


greg-does-weather (approved):
do the dew

PR #895, opened 2024-03-15
Remove unnecessary padding in alert tab

partly-igor:
## What does this PR do? 🛠️
When working on #894 I realized there was some unnecessary padding classes that were cancelling each other out in the alert tab. Quick fix to simplify. Visually should look the same as before.

Screenshot of alerts tab showing alignment


greg-does-weather (approved):
shrink

PR #896, opened 2024-03-15
Fix up alert times

### LINKED ISSUE #809: Replace Alert List date/times with plain language
The current formula for what to display in the Alert List for any hazard product is: 
[event] in effect [onset] [time zone] - [ends] [time zone] 

This update will replace that formula with a series of new ones that are conditional based on the current time in relation to the "onset" time. 
[A full table displaying every option is available here](https://docs.google.com/document/d/1fSe3Qx3TPcDxBy-6KbbV3npN6PMfb5U5TU5J0bUpJ4o/edit#heading=h.3w5mxqgb80k)

This new formula would apply to any/all alerts, and should be tested with multiple alerts in a single place to see the full effect.

greg-does-weather:
## What does this PR do? 🛠️

Fixes up alert times in the banner list. @kmranjo and I huddled to tweak the spec a bit more for alerts that end later today.

closes #809 

image


partly-igor (approved):
This looks great!

PR #898, opened 2024-03-15
CRITICAL: Fixes a bug where all data requests fail

greg-does-weather:
## What does this PR do? 🛠️

Adds the logger trait to the data layer. The bug is currently active in staging, the result being that we can't see current conditions or any forecast data for any location.


partly-igor (approved):
👍 Looks like it's working, thanks for fixing!

PR #900, opened 2024-03-15
CRITICAL: Fix an issue with point lat/lon

greg-does-weather:
## What does this PR do? 🛠️

We removed some code that converted point arrays (`[lon, lat]`) into objects (`{ lat: y, lon: x }`), but not all of the code that uses these coordinates was updated. In dev, this does not present any errors, but in our cloud environments it causes SQL errors.


partly-igor (approved):
Bless

PR #915, opened 2024-03-18
R3 fix date 

partly-igor:
## What does this PR do? 🛠️
One of the R3 alerts has a small bug where the end date is before the start date. The correct end date should be 5 AM MT Wed, 2/26. 

image

Question – Do we need to redeploy the proxy only? Everything?

image


mgwalker (approved):

mgwalker:
Proxy only

greg-does-weather:
I agree with me

greg-does-weather (approved):

PR #922, opened 2024-03-18
Support pinning API proxy dates/times specific times during a day

greg-does-weather:
## What does this PR do? 🛠️

Adds a new datetime token:

```
date:today 04:00:-3
```

This will be converted to 4am of the current day in a timezone that is 3 hours behind UTC. These can be combined with the existing modifiers as well, to create specific times of day on future days. For example, 3pm tomorrow:

```
date:today 15:00-5 +1 day
```

Also updates all of the daily forecast test data so that every period begins and ends at 6am/6pm or vice versa, accordingly.


eric-gade (approved):
Code looks good 👍 
  
Played around with this for a little bit so that I could understand the alerts etc. I've noticed a few things about the updated stamps at the testing "many alerts" example:
1. Some of the alerts still end right at the `now` time, ie when I've loaded the page, so we don't see them represented in the hourly details tables. I'm surprised, however, that they are still showing at all on the page in that case. Might still be some logical error I'm not seeing in the hourly detail / alert alignment code.
2. The "Today" card in the daily tab is actually set to tomorrow relative to the day you load the page. (I'm loading it the night of the 19th, but it shows "today" as tomorrow the 20th). I assume that's to get the full 6am-6pm coverage. Perhaps we should set the bundles to be more realistic -- ie the "today" card's hours will be from `now` to 6am the following day.
3. One problem with (2) above, is that the bundled data can have the two "today" periods in the json file (daytime / nighttime), but if I'm looking at night I would only expect to see one of those. Not crucial, but something that messes up time for sure
  
None of that is blocking, just noting it

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code


## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #924, opened 2024-03-19
Round precip values in the table to hundredths of an inch

### LINKED ISSUE #903: Round precipitation amounts to hundredth of an inch
Currently in beta.weather.gov, 6-hourly precipitation amounts are rounded to nearest tenth.  That means that anything 0.04 or less is rounded down to zero.  0.04 inches will cause puddles and is measurable (anything 0.01 or greater is considered measurable).  This can be confusing if there is rain in the forecast but zero in the precipitation forecast.  Also, forecasters predict precipitation in hundredths.  This ticket would mean that every time we show precipitation amounts in beta.weather.gov, they are rounded to the nearest hundredth. 

User story: I want to see any precipitation that is a hundredth or more because even a few hundredths can cause impacts to construction and other activities.

Capture

greg-does-weather:
## What does this PR do? 🛠️

closes #903

Screenshot 2024-03-19 at 9 04 07 AM

### DIFF
--- a/web/modules/weather_data/src/Service/HourlyForecastTrait.php
+++ b/web/modules/weather_data/src/Service/HourlyForecastTrait.php
@@ -273,7 +273,7 @@ trait HourlyForecastTrait
                 $periods[] = (object) [
                     "start" => $start->format("g A"),
                     "end" => $end->format("g A"),
-                    "value" => round($value, 1),
+                    "value" => round($value, 2),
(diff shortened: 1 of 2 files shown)


eric-gade (approved):
Code is small and straightforward, and I tested in a few rainy places live (thanks, Seattle and friends) 👍

Issue #837, opened 2024-02-27
The whole site crashes if we reference an icon that doesn't exist


greg-does-weather:
### Description

Our Twig template for current conditions pulls in icons by injecting them directly into the HTML, which causes a Twig exception if the icon file does not exist. We should, uh, not crash because an icon is missing.

[bundle_65de6cb1725a9.zip](https://github.com/weather-gov/weather.gov/files/14426676/bundle_65de6cb1725a9.zip)

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

colinmurphy01:
This feels pretttttty `high`?

moving to backlog

colinmurphy01:
Also related #764

PR #927, opened 2024-03-19
Fix icon crashing and test for icon errors

eric-gade:
## What does this PR do? 🛠️
This PR addresses #837, which initially would crash the whole site if an unknown icon was returned by the API, but in more recent versions of the code would crash specific block rendering.
  
This PR also accounts for three different "invalid" icon values:
1. A `null` value;
2. An arbitrary string that is not a valid URL of any kind;
3. A valid icon URL, whose final portion has no corresponding entry in our lookup table

Most of the heavy lifting was already done, we just needed to record/edit a bundle with examples of invalid icons and write e2e tests for them.
  
Additionally, I have updated current conditions to use the spritesheet rather than loading the file directly inline, which is how other templates render icons.


tests/api/data/testing/points/38.886,-77.094.json
```diff
@@ -0,0 +1,58 @@
+{
+  "@bundle": {
+    "name": "Invalid icons (should not crash site)"
```
greg-does-weather, line 3: :lolsob:

greg-does-weather (approved):
Nice tests!

Issue #901, opened 2024-03-15
Address lat/long inconsistencies


colinmurphy01:

colinmurphy01:
Eric and Greg to sync up on how we want to address this

PR #928, opened 2024-03-20
Define the variable used for cache lookups

greg-does-weather:
## What does this PR do? 🛠️

The `$wktPoints` variable is never defined before attempting to use it as a cache key. This does not appear to cause any issues, but it produces a PHP warning and prevents caching from working correctly.

Introduced in #901


eric-gade (approved):
Small change so no big checklist, this is good to go 👍

greg-does-weather:
Ha, this was also covered by #911

Issue #827, opened 2024-02-26
Hourly alert table cells should have normal click handling


eric-gade:
### Description
With #792, we are adding another set of Alert links, this time as spanned table cells in the new Hourly Forecast table component.
  
We need these links to have the same behavior as the above-the-fold alert links. Specifically, clicking a link here should:
1. Activate the Alerts tab in the tabbed navigator
2. Expand the corresponding Alert accordion
3. Scroll so that the expanded Alert is in view

### Acceptance criteria
- [x] When clicked, an hourly table alert link switches the tabbed navigator to the Alerts tab
- [x] When clicked, an hourly table alert link expands the corresponding alert accordion
- [x] When clicked, an hourly table alert link scrolls to the corresponding alert so that it is in view

Issue #878, opened 2024-03-13
Accessibility: Update hourly forecast tables


greg-does-weather:
### Description

The hourly forecast tables are currently pretty rough in assistive devices. Navigating through them on a screen reader, I'm not sure a table is the best layout. Some potential alternatives might be:

- ~instead of a row for each piece of information, collect it all into a single cell (using a card, perhaps?) so when a user lands in the cell, the reader announces the column header (the hour) followed by the hourly forecast details~
- ~hide the table from screen readers and implement an sr-only ordered list of the hourly forecast. This seems more appropriate to me than the first approach since the first approach essentially breaks the semantics of tables~
- ~reimplement the table as a flexbox, which could semantically be a list and thus 🪄 magically better for a11y~

See PR #877

### Acceptance criteria

Coming out of R3 a more ideal solution might be:

- Adding caption to the table
- Treating the times as column heads

Will confirm in Des/Eng sync today

- [ ] Implementation has been modified to work for all users
- [ ] Changes have been tested for accessibility

colinmurphy01:
@greg-does-weather is a more detailed, manual accessibly review needed or is this ready for `eng`?

colinmurphy01:
Will rescope after R3.

colinmurphy01:
@sarahmhgov have we identified most of the accessibility issues in the current experience.

Do we need to do a manual review?

colinmurphy01:
Coming out of R3 a more ideal solution might be:

- Adding caption to the table
- Treating the times as column heads

Will confirm in Des/Eng sync today

colinmurphy01:
@igorkorenfeld Did we confirm this last week? I think so, but don't recall.

colinmurphy01:
Something to explore - at celle level, my headings are X/Y.

This will allow a screen reader to announce both headers.

Test with WCAG/508 lead

greg-does-weather:
As far as I can tell, aside from adding a caption, there's not much we can do to improve the screen reader experience here. The row and column headings are already identified as such, and in my (admittedly naive) testing, VoiceOver reads them out.

colinmurphy01:
@igorkorenfeld any thoughts on what we should write as the caption?

partly-igor:
How about "Detailed hourly weather data for <date>. The rows are weather variables, the columns are hours"

greg-does-weather:
"for"? Is that supposed to be the day? "Today," "Thursday," etc.?

[...]

Issue #862, opened 2024-03-03
For anchor links on alerts, use more distinctive IDs than just list index


greg-does-weather:
### Description

Currently, if a user copies a link to an alert on the location page and shares that with a friend, if another alert is issued with higher priority before the second user opens the page, they will be shown a different alert than the one they were sent. Instead of using list index as part of the DOM ID, we should see if there's a more distinct ID associated with the alert we can use that will remain valid and correct for as long as the alert is active.

### Acceptance criteria

- [ ] A link to a specific alert within the location page remains valid for the lifetime of the alert, regardless of whether new, higher-priority alerts are issued in the interim.

PR #929, opened 2024-03-20
Eg 827 hourly alert links

eric-gade:
## What does this PR do? 🛠️
This PR implements the behavior in #827. Specifically, alert spans inside hourly-table element tables should be links, and when clicked they should have the same behavior as the above-the-fold alert links at the top of the page.

Have created e2e tests to capture the expected behavior properly.
  
Note also that we updated how alert ids (for the purposes of the twig templates / DOM) are created.


tests/e2e/cypress/e2e/hourly-forecast-table.cy.js
```diff
@@ -30,4 +30,29 @@ describe("Hourly forecast table tests", () => {
         .should("equal", "5");
     });
   });
+
+  describe("Alert span clicking behavior", () => {
```
greg-does-weather, line 34: These tests are lovely.

web/modules/weather_data/src/Service/AlertTrait.php
```diff
@@ -227,6 +227,13 @@ public function getAlerts($grid, $point, $self = false, $now = false)
 
         $alerts = AlertUtility::sort($alerts);
 
+        // The API's alert ids are just too long for use
+        // in the DOM. Let's loop through and simply
+        // assign an index value
+        for ($i = 0; $i < count($alerts); $i++) {
+            $alerts[$i]->alertId = $i + 1;
+        }
+
```
greg-does-weather, line 236: Agreed, those IDs are bonkers, and this is probably smarter than using the Twig index property. 😅 

We'll revisit this in #862 as well. One quirk of using indices is that the links in the table can be different colors: visited anchors will be black, unvisited ones will be blue. (cc: @igorkorenfeld on whether that's the desired behavior). But the unintended effect is that a completely different alert will have the exact same link in the future, so it can be misleading to users about what they have or haven't seen.

greg-does-weather (approved):
Something to add to #878 is how to handle the changing context when a user clicks a local link. Do screen readers just know to do that? Do we need to announce it somehow? But it's definitely out of scope of this PR.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [x] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_data/src/Service/AlertTrait.php
eric-gade: That is a really good point. Maybe we can get a smaller id by doing something like a hash of the timestamp, the alert type, and the first applicable forecast/county zone code

PR #934, opened 2024-03-21
Add check for values in hourly table

partly-igor:
## What does this PR do? 🛠️
This PR adds a check before each value in the the table and renders an "N/A" if the value is (strictly) null. 

To test I manually updated the Little Rock testing files to put in null and 0 values, but did not commit them. I could create a new set of files (in a different grid point) if that's helpful 
 
Screenshot of table with the added N/A values


greg-does-weather (approved):
I'm not available

Issue #921, opened 2024-03-18
Hourly table arrows look smaller on mobile devices


colinmurphy01:
### Description

Hourly table arrows seem to be smaller on mobile devices. 

- Re-produceable on chrome and safari browsers on an iphone Xr and 14 Pro
- Re-produceable on Today and Daily tabs

![Image](https://github.com/weather-gov/weather.gov/assets/24654605/00461b0c-d426-4aff-9bed-274c73342653)

Not able to reproduce on dev tools (mobile)

![Image](https://github.com/weather-gov/weather.gov/assets/24654605/2918a2b0-403d-4b3d-aa67-a6d1b0ed5fbd)

### Acceptance criteria

- [ ] Arrows are consistent on mobile devices
- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

partly-igor:
Also shows up on iPhone 12. But doesn't seem to occur on Android devices 

Not replicable on responsive mode in Chrome or Safari on desktop.

PR #943, opened 2024-03-22
Adjust hourly table styles

partly-igor:
## What does this PR do? 🛠️
This PR address two bugs that were rendered on iPhones but not on other devices: 

1. Arrow sizes - This should fix the visual bug in #921 
2. Gap below "Condition" and "Wind speed" and double border on those headers (also see in the screenshot in #921)

Additionally it adds hover and active state colors to the arrow buttons, matching the design. 

Currently live deployed on https://weathergov-eric.app.cloud.gov/point/40.715/-74.007#daily to test on iphone

iphone screenshot of hourly table


web/themes/new_weather_theme/assets/sass/components/hourly-table.scss
```diff
@@ -15,7 +28,7 @@
   td, th {
     padding-left: 0;
     padding-right: units(3);
-    vertical-align: text-top;
+    vertical-align: top;
```
greg-does-weather, line 31: Does this fix both the arrow size *and* the padding? (I ask because the other changes all appear to be related to the scroll buttons.)

greg-does-weather (approved):
One question, and I trust your judgement on when this is ready to merge.

web/themes/new_weather_theme/assets/sass/components/hourly-table.scss
partly-igor: This line only fixes the wonky space and double border in the table cell on iPhone browsers. The arrow sizing is fixed by setting explicit padding on the arrow buttons. 

screenshot of bad spacing in table on iphone

greg-does-weather: Oops, I was looking at the wrong arrows. Got it!

PR #963, opened 2024-03-26
Updating geonames data/parsing to include US territories

### LINKED ISSUE #957: Wrong city names displayed for cities in U.S. Territories
### Description
Searching for and picking a number of cities in  U.S. territories returns point location pages with seemingly correct grid points and data, but the wrong city name. 
Examples: 
"Guam Intl Airport, Guam, GUM" returns a point location page with "[Eastport, ME](http://localhost:8080/point/13.484/144.797)" as the city name, 13.484/144.797 as the grid point
"San Juan, PRI" -> [North Key Largo, FL](http://localhost:8080/point/18.465/-66.117) : 18.465,-66.117
"Cruz Bay, St. John, VRI" -> [North Key Largo, FL](http://localhost:8080/point/18.332/-64.794) : 18.332, -64.794, 

### Acceptance criteria
- [ ] Searching for location that are located in U.S. territories returns the correct name

eric-gade:
## What does this PR do? 🛠️
Issue #957 identified cases where point location pages in Puerto Rico and in Guam were not giving the correct place names. This is due to the fact that we had filtered the geonames dataset to the `US` country code. However, in that dataset places within US territories use their own ISO-3166-alpha-2 country codes, rather than `US`.
  
This PR implements the following changes:
* Normal spatial data loading makes sure we load all geonames rows with either a `US` or corresponding US territory row
* Added a script that trims the full cities500 file to the US places and its territories
* Updated the compressed version of the trimmed cities500 dataset

To properly run this, do the following:
1. Remove `spatial-data/us.cities500.txt` from your local directory, if it exists. It will be replaced by the unzipped version by the loading script
2. Run `make zap` to re-initialize the docker containers and database
  
## Closes 
#957


partly-igor:
Guam is looking great (searching for Guam Intl Airport seems to return the correct name),
but San Juan, PRI and Frederiksted, St Croix, VIR are still showing "North Key Largo, FL" as the place name.

partly-igor:
Relatedly, Rota, MNP returns "Mongmong-Toto-Maite Village, GU" which is the same name returned for "Guam Intl Airport, Guam, GUM", but should be different.

eric-gade:
@partly-igor @greg-does-weather I'm closing this PR for now. Further exploration on our end has revealed that this issue is more complicated than I'd hoped, and will require more work. I will log what I've found so far in the originating ticket, #957

greg-does-weather:
Don't abandon this PR yet. I played with a bit this evening before logging out. You're mostly there. We basically need to handle a few special cases and we're there.

The Northern Mariana Islands case is still returning a result from Guam, but it's at least in the right part of the ocean now. We're getting close.

greg-does-weather (approved):

PR #981, opened 2024-03-29
Cache a database dump in CI/CD to speed up runs

greg-does-weather:
## What does this PR do? 🛠️

- caches a dump of our fully-populated database (keyed on config and content hashes)
- if no cache, builds the database
- in subsequent testing steps, rather than building the database, restore it from cache

This takes our build times from around 8 minutes to around 4 minutes.

### DIFF
--- a/.github/actions/setup-site/action.yml
+++ b/.github/actions/setup-site/action.yml
@@ -1,0 +1,51 @@
+runs:
+  using: composite
+  steps:
+    - name: setup image cacheing
+      if: needs.should-test.outputs.yes == 'true'
+      uses: actions/cache@v4
+      with:
+        key: drupal-image-${{ hashFiles('composer.lock','web/sites/example.settings.dev.php','Dockerfile.dev') }}
+        path: /tmp/image.tar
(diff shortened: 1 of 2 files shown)


.github/actions/setup-site/action.yml
```diff
@@ -0,0 +1,51 @@
...
+          --password=drupal \
+          --user=root \
+          weathergov < weathergov.sql
+
+        echo "UPDATE users SET uid=0 WHERE uid=2;" | mysql \
+          --host=127.0.0.1 \
+          --port=3306 \
+          --password=drupal \
+          --user=root \
+          weathergov
```
greg-does-weather, line 51: User 0 is the anonymous user. The uid is preserved correctly in the dumped data, but MySQL will interpret that as a null and give this user an autoincremented ID. Without an anonymous user, we can't actually view any of the pages so all of our tests fail. Easy enough to get around it: change the uid back to 0.

loganmcdonald-noaa (approved):
LGTM!

PR #983, opened 2024-03-29
For visual cases, use separate degree and F characters

### LINKED ISSUE #930: Fahrenheit symbol font varies between operating systems
### Description
We currently use the single character "℉" (U + 2109) to represent Fahrenheit throughout the site. However our primary font, Nunito Sans does not have that character (nor do most of the fonts in our font stack). As a result the character renders in one of the system default fonts. This is particularly noticeable on windows machines where it is rendered in Times New Roman, and which appears inconsistent with the rest of the design. 

We should replace the single character with two, the degree symbol "°" and the "F". This will create more consistent rendering across browsers and operating systems, given that Nunito Sans does have the "F" character. 

We should consider adding an aria-label Fahrenheit wherever this symbol is used so that the screen reader reads the unit name rather than "F"

<details>
<summary> Screenshots </summary>
 
Screenshot from Windows

Screenshot from mac

</details> 

### Acceptance criteria
- [ ] All single character "℉" have been replaced with two characters "°F" 
- [ ] Compare the rendering of the font on different operating systems. 

greg-does-weather:
## What does this PR do? 🛠️

- Where we display degrees, visually use two characters so they will be in our design system fonts, but hide those from screen readers
- For screen readers, preserve the single character and mark it sr-only

closes #930

### DIFF
--- a/web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
+++ b/web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
@@ -33,9 +33,8 @@
         {% for period in hours %}
         <td>
           {% if period.temperature is not same as (null) %}
-          {{ period.temperature }}
+          {{ period.temperature }}<span aria-hidden="true" class="font-sans-3xs text-ttop padding-left-1px">&deg;F</span>
           <span class="usa-sr-only">℉</span>
-          <span aria-hidden="true" class="font-sans-3xs text-ttop padding-left-1px">&deg;F</span>
           {% else %}
           <span class="text-base font-mono-xs"> {{ "N/A" | t }} </span>
           {% endif %}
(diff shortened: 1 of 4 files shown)


web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
```diff
@@ -33,7 +33,9 @@
         {% for period in hours %}
         <td>
           {% if period.temperature is not same as (null) %}
-          {{ period.temperature }}<span class="font-sans-3xs text-ttop padding-left-1px">℉</span>
+          {{ period.temperature }}
+          <span class="usa-sr-only">℉</span>
```
partly-igor, line 37: Could we remove the space between the number and the degree symbol? I believe it's because the `span` elements are on their own line?

**PR** 
Screenshot of the gap


**Goal**
image

partly-igor (approved):
Just one visual change to make, but  other than that, both visually and with a screen reader it was looking good!

web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
greg-does-weather: Oops! I knew something seemed off but I couldn't identify it. Yes, I'll make that fix!

PR #985, opened 2024-03-29
Add chance of precipitation row to hourly table

### LINKED ISSUE #973: Add chance of precipitation as its own row on the hourly table
### Description
In R3 we heard users express confusion about the "% chance" line under the weather condition, particuarly when the condition was not rain (e.g. "Windy, 23 % chance").  We should add "Chance of precipitation" as its own line row of data, below the condition row. 

### Acceptance Criteria
- [ ] "% chance"is removed from the condition row
- [ ] Chance of precipitation is a row in the table below the "Condition" row

partly-igor:
## What does this PR do? 🛠️
This PR adds chance of precipitation as its own row in the hourly table, and removes it from the condition row. Every hour will now display a percent chance. 

Given the longer word, I have reduced the right-padding on the heading column a touch and allowed the text in that column to wrap.

Closes #973 

Desktop screenshot of precipitation row

Mobile sized screenshot of precipitation row


web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
```diff
@@ -60,15 +60,26 @@
...
         </td>
         {% endfor %}
       </tr>
+      <tr data-row-name="chance-precipitation">
+        <th scope="row" class="position-sticky left-0 font-family-mono text-primary-darker bg-primary-lightest font-mono-xs text-uppercase z-400">
+          {{"Chance of precipitation" | t}}
+        </th>
+        {% for period in hours %}
+        <td> 
+          {% if period.probabilityOfPrecipitation is not same as (null) %}
```
greg-does-weather, line 75: This adds back the `0%` chance, right?

greg-does-weather (approved):
raining

web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
partly-igor: That's right, my understanding is that `is not same as` is `!==` under the hood

greg-does-weather: Yep. Okay, cool. I was asking because the old code didn't show precip chances below 1%, but if we're going to have a row dedicated to precip chance, it makes sense (to me, anyway) to always show it.

partly-igor: That's my thinking as well – that 0% chance is actually meaningful data

PR #986, opened 2024-03-29
List out existing block-based e2e tests

greg-does-weather:
## What does this PR do? 🛠️

Longer term, it might be useful to document the behaviors we're capturing in our tests to help the whole team see where we might need additional testing. This is a first pass at doing that, by documenting the block-based e2e tests that we've implemented.

### DIFF
--- a/docs/dev/tests.md
+++ b/docs/dev/tests.md
@@ -1,0 +1,58 @@
+## Data behavior tests
+
+> [!NOTE]  
+> Data structure tests refer to data manipulation done by weather.gov in order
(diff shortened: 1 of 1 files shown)


loganmcdonald-noaa (approved):
Is there a way we could automate this when we update tests? Either way, very useful.

greg-does-weather:
I don't think so, which makes maintenance questionable. Maybe we could do something with [phpdoc](https://phpdoc.org/), but that looks like overkill for what we need.

PR #1002, opened 2024-04-02
Change alert ordering

### LINKED ISSUE #940: Order alerts chronologically
### Description
  
We current display alerts (both the list and the details) in order of severity according to a priority list of hazards. In R3 we learned that this causes confusions for users because it means that alerts may not be listed chronologically. For example, as seen in the screenshot below, an alert like a "Winter Storm Warning" that is happening _tomorrow_ will get listed before a lower priority alert like "High Wind Warning" that is happening _today_. We should update our sorting logic so that alerts are first sorted chronologically (earliest to latest) and then by priority order (high to low). 
  
### Acceptance criteria
- [ ] Active Alerts (past onset time) are listed according to priority list, first
- [ ] Future Alerts appear in chronological order, second
- [ ] Alerts with same onset time are listed according to the priority list 
- [ ] Alert accordions appear in the same order as the alert links at the top of the page
  
Screen Shot 2024-03-22 at 11 08 39 AM

greg-does-weather:
## What does this PR do? 🛠️

- Sort currently-active alerts by priority
- Sort future alerts by onset time
- If multiple alerts have the same onset time, sort them by priority within their onset time

closes #940

- Incidentally updates the alert structure tests to use JSON schema validation, so we don't have to care as much about *actual* values, only the form of the values. (Both of those e2e tests needed to be updated for this PR anyway because the ordering of alerts was different after the change, and I figured rather than keep fighting these super fragile timing tests, I'd just... try something new.)

### DIFF
--- a/web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
+++ b/web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
@@ -18,9 +18,7 @@ use PHPUnit\Framework\TestCase;
  */
 final class WeatherDataLegacyIconMapping extends TestCase
 {
-    /**
-     * @group unit
-     */
+    #[Group(unit)]
     public function testIconMapping(): void
     {
         $path = realpath(
--- a/web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
+++ b/web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
@@ -18,7 +18,9 @@ use PHPUnit\Framework\TestCase;
  */
 final class WeatherDataLegacyIconMapping extends TestCase
 {
-    #[Group(unit)]
+    /**
+     * @group unit
+     */
     public function testIconMapping(): void
     {
         $path = realpath(
--- a/composer.json
+++ b/composer.json
@@ -113,6 +113,7 @@
     "require-dev": {
         "drupal/coder": "^8.3",
         "drupal/devel": "^5.1.2",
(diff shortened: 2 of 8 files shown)


web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
```diff
@@ -18,9 +18,7 @@ use PHPUnit\Framework\TestCase;
  */
 final class WeatherDataLegacyIconMapping extends TestCase
 {
-    /**
-     * @group unit
-     */
+    #[Group(unit)]
```
eric-gade, line 21: Was trying to find docs about this syntax and couldn't, but if it's all good please ignore

eric-gade (approved):
LGTM 👍 
  
Love the JSON schema stuff. No-brainer!  

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_data/src/Service/Test/LegacyIconMapping.php.test
greg-does-weather: Oh oops, nope, that needs to come back out. I got off course trying to use a newer version of phpunit and forgot to unwind it all.

PR #1013, opened 2024-04-04
Adds the new content for taxonomies with region hierarchy

loganmcdonald-noaa:
## What does this PR do? 🛠️
Replacing taxonomy hierarchy with states with region only via content. 

Also adds some documentation about how to copy exports locally from a cloud environment if single-content-sync isn't doing it for you.


greg-does-weather (approved):
Excellent! The new exported taxonomy terms are exactly right.

I took the old `content-bulk-export*.zip` and split it into three files:
- `beta-banner.zip` contains the beta banner block that shows up at the top
- `footer-menus.zip` contains the agency identification and footer menu items
- `initial-blank-pages.zip` contains placeholder pages for things like "about" or "home"

It turns out that zip file also contained WFO taxonomy terms with some overlapping UUIDs, so importing them both resulted in some duplication and also some weird stuff with hierarchy. Deleting the taxonomy terms from the bulk export cleared that up for me.

PR #1025, opened 2024-04-08
Use wind direction arrows instead of cardinal direction names

### LINKED ISSUE #909: Replace cardinal wind direction in current conditions with arrow
On current conditions, wind direction is communicate via cardinal direction (e.g. NE, SW).  
image

We'd like the current conditions component to use an arrow instead, similar to how we display it in the hourly table.
image

Reasoning it's believed that arrows communicate the wind more clearly considering it's pointing where the wind is going, whereas the cardinal direction notes where the wind originates.

Accessibility will be tracked in #1007

### LINKED ISSUE #1007: Accessibility: Wind direction is not read by screen reader
### Description
Currently in the hourly table the wind direction is not read aloud by the screen reader.

Image

Related ticket: #909 

### Suggestion
- upon selection screen reader will also read aloud "from [origin] to [destination]"

### Acceptance criteria
- [ ] Wind direction is read aloud

greg-does-weather:
## What does this PR do? 🛠️

For current conditions, replaces the wind direction with the arrow. For screen readers, the arrow is hidden and the text "from the {wind direction}" is used instead.

Also adds the new wind direction icon to the sprite sheet.

- closes #909
- closes #1007

- pulled the wind direction into a partial so we can use it in the hourly tables as well as the current conditions

image


web/themes/new_weather_theme/templates/partials/wind-direction.html.twig
```diff
@@ -0,0 +1,9 @@
+{#
+The +180 here is to account for the fact that wind direction
+is reported as a FROM direction, but our arrow points in the
+TO direction. So we just need to spin it 'round.
+#}
+<svg role="img" aria-hidden="true" data-wind-direction class="width-2 height-2 margin-left-05 padding-2px" style="transform: rotate({{ windDirection.angle + 180 }}deg);">
```
partly-igor, line 6: In general looking great but on the current conditions the icon looks a little too small. I imagine it's the combination of padding and the width on the icon:

Compare the table with current conditions:  
wind icon on the table

wind icon on current

Playing around with the dev tools I found that setting the class on the svg to `width-105 height-105 text-middle` (removing the padding and margin) got the icon pretty close to the right size and centered. But I imagine this would cause an issue in the table?

The other thing is that I could only get the icon to more closely be centered with the text if I added a `translate: -5% -12%;` to the `transform`. But those are some magic numbers. I am not sure why the rotated svg isn't quite centering with the text without that.

greg-does-weather: On it!

greg-does-weather: There was an old `.item` class on the arrow's parent in the current conditions. That was meant for the old icons that we're not using anymore (wind and humidity), and it was further reducing the size. I yoinked that out and wrapped the wind speed in a flex container with center alignment for the second problem.

greg-does-weather: image

image

eric-gade (approved):
👍 
  
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #1029, opened 2024-04-09
Fix the timestamp in the screen reader narrative for current conditions

### LINKED ISSUE #1016: Accessibility: Screenreader reads off GMT+0000 within Current Conditions
### Description
What it says on the tin.

Image

### Acceptance criteria

- [ ] Screen reader description includes plain language timezone (EDT)

greg-does-weather:
## What does this PR do? 🛠️

The timestamp in the screen reader narrative section of the current conditions component did not get updated when we switched how we do time localization. This PR fixes that.

> [!NOTE]  
> VoiceOver treats inline elements (like `span` or `time`) as independent blocks when reading them, so users must navigate to them individually. We cannot get around this. In Safari, adding `role="text"` to the container element will force VoiceOver to treat it as a single block of text, but that's a nonstandard role and is ignored by all other browsers.
>
> I only have access to VoiceOver, so I cannot test this behavior with other screen readers, unfortunately, but the internet seems to think it should work as expected with them. VoiceOver is the oddball.

- closes #1016

### DIFF
--- a/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
+++ b/web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
@@ -14,15 +14,13 @@
       {# First row is temperature #}
       <div class="display-flex">
         <div class="current-conditions-temp display-flex flex-row flex-align-start margin-right-2 text-primary-darker">
-          {{ content.temperature }}<span aria-hidden="true" class="font-body-xs display-inline-block left-full margin-top-05 margin-left-neg-2px">&deg;F</span>
-          <span class="usa-sr-only">℉</span>
+          {{ content.temperature }}<span class="font-body-xs display-inline-block left-full margin-top-05 margin-left-neg-2px">&deg;F</span>
         </div>
         <div class="margin-top-05 position-relative">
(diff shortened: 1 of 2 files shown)


web/themes/new_weather_theme/templates/block/block--weathergov-current-conditions.html.twig
```diff
@@ -14,15 +14,13 @@
...
+          {{ content.temperature }}<span class="font-body-xs display-inline-block left-full margin-top-05 margin-left-neg-2px">&deg;F</span>
         </div>
         <div class="margin-top-05 position-relative">
           <div class="font-mono-2xs font-family-mono text-base text-uppercase">{{ "Feels like" | t }}</div>
           <div class="text-primary-darker">
-            <p class="margin-top-2px font-body-md">{{ content.feels_like }}
-              <span aria-hidden="true" class="font-body-3xs position-absolute margin-top">&deg;F</span>
-              <span class="usa-sr-only">℉</span>
+            <p class="margin-top-2px font-body-md">
+              {{ content.feels_like }}<span class="font-body-3xs position-absolute margin-top">&deg;F</span>
```
greg-does-weather, line 23: We don't need screen reader-specific handling here because these entire blocks are hidden from screen readers.

PR #1042, opened 2024-04-12
Use ISO8601 timestamps instead of Unix epoch

### LINKED ISSUE #1014: BUG: Current observation reports observation time as epoch timestamp instead of ISO8601
### Description

Just what it says on the tin. The upshot is that we put a semantically invalid value into our `<time>` element. Our timestamp localization script expects the epoch timestamp, but in another PR (tk) I have updated it out of necessity to also work with ISO8601 timestamps.

However, epoch timestamps are not valid for the `<time>` element. ISO8601 timestamps are. So... let's make the switch.

### Acceptance criteria

- [ ] the current observations block returns the timestamp as an ISO8601 string
- [ ] the current observations e2e test captures the bug

greg-does-weather:
## What does this PR do? 🛠️

When I originally created the current conditions block, I opted to put the Unix epoch timestamp in the Twig template because Javascript Dates natively understand that (plus or minus 3 orders of magnitude...). Then we switched to using the `<time>` element, whose `datetime` property has a limited set of valid formats, none of which are Unix epoch.

This PR updates the current conditions and weather story blocks to use ISO8601 timestamps instead of Unix epoch, making our `<time>` elements spec-compliant. It also fixes the relevant tests.

There should be no user-visible changes.

- closes #1014

### DIFF
--- a/web/modules/weather_blocks/src/Plugin/Block/Test/WeatherStoryBlock.php.test
+++ b/web/modules/weather_blocks/src/Plugin/Block/Test/WeatherStoryBlock.php.test
@@ -85,7 +85,7 @@ final class WeatherStoryBlockTest extends Base
             ],
             "updated" => [
                 "formatted" => "Jun 24, 1977, 7:16 AM",
-                "utc" => "235984600",
+                "utc" => "1977-06-24T07:16:40+00:00",
             ],
         ];
 
@@ -112,7 +112,7 @@ final class WeatherStoryBlockTest extends Base
             "image" => null,
             "updated" => [
                 "formatted" => "Jun 24, 1977, 7:16 AM",
-                "utc" => "235984600",
+                "utc" => "1977-06-24T07:16:40+00:00",
             ],
         ];
 
--- a/tests/e2e/cypress/e2e/dates-are-localized.cy.js
+++ b/tests/e2e/cypress/e2e/dates-are-localized.cy.js
@@ -53,9 +53,7 @@ describe("main script", () => {
             const format =
               timestamp.getAttribute("data-date-format") || "basic";
 
-            const expected = formatters
-              .get(format)
-              .format(dayjs.unix(utc).toDate());
+            const expected = formatters.get(format).format(dayjs(utc).toDate());
             expect(actual).to.equal(expected);
           }
         });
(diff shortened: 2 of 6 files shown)


web/modules/weather_blocks/src/Plugin/Block/Test/WeatherStoryBlock.php.test
```diff
@@ -112,7 +112,7 @@ final class WeatherStoryBlockTest extends Base
             "image" => null,
             "updated" => [
                 "formatted" => "Jun 24, 1977, 7:16 AM",
-                "utc" => "235984600",
+                "utc" => "1977-06-24T07:16:40+00:00",
```
eric-gade, line 115: what's the backstory on this date? [Here](https://en.wikipedia.org/wiki/June_1977) it says the National Primary Drinking Water Regulations first went into effect, so I'm rolling with that for now.

eric-gade (approved):
This looks good 👍 but I noticed one issue: when using VO, the robot will read the time as given, plus the GMT offset. This is usually not the same time as is displayed on the page. In other words, the timestamp reads funny because it's not specifically localized to the location's timezone and I guess the screenreader doesn't do anything fancy with these timestamps.
  
For example, VO read the following as "1:54 PM GMT plus zero zero zero zero," but it's 9:54 AM in the local time:
  
Screen Shot 2024-04-15 at 10 27 33 AM


## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_blocks/src/Plugin/Block/Test/WeatherStoryBlock.php.test
greg-does-weather: Let's go with that. I don't actually know. On the eAPD project, I put comments next to all the test dates to explain why I'd chosen them, but this one may just be totally arbitrary.

greg-does-weather:
Oh geez. That's technically the correct behavior, I think. I'll dig into this some more.

I *think* we've discussed ditching user-local times in favor of location-local times, in which case we would send ISO strings with the correct offset for the location and it might work out. Though, honestly, the offset is not a useful thing for almost anyone so it'd be great if the screen reader was smarter than that. :sigh:

There are other valid formats for the `datetime` property too, so I'll look into those as well.

PR #1053, opened 2024-04-15
Fix reading order of current conditions

### LINKED ISSUE #1008: Accessibility: Reading Order on Current Tab
### Current
Upon loading a location page and navigating the tabs. as you tab through the tabs and the focus within the Current tab. The screen reader goes to the observation station info first and then reads the descriptive forecast, while the focus box remains on the observation link.

Image

### Expected
Upon focusing within the Current tab, screen reader should:
1. Read the summarized version of current conditions first
2. Read Observation Station and the link second
3. Read Last update time last. 

### Acceptance criteria
- [ ] Order has been updated to reflect expected behavior

greg-does-weather:
## What does this PR do? 🛠️

Moves the narrative of the current conditions, which is used by screen readers, before the "observation station" section, so it is in the same order as for sighted users.

closes #1008


eric-gade (approved):
Quick approve since it's small 👍

PR #1058, opened 2024-04-15
Update hourly table behavior and add appropriate tests

### LINKED ISSUE #999: Add e2e tests: hourly forecast table
### Description

We need to make some more assertions about the hours that are shown.

### Acceptance criteria

- Tests have been implemented for:
  - [x] Hours start at the top of the next hour
  - Hours go up to 6am the next day
    - [x] If before midnight on day 1, hours end at 6am on day 2
    - [x] if after midnight on day 1, hours end at 6am on day 3

### LINKED ISSUE #1000: Update hourly forecast table in "today" tab
### Description

See behavior described in #999.

### Acceptance criteria

- [ ] I mean, basically if tests for #999 are implemented and passing, this must also be done.

greg-does-weather:
## What does this PR do? 🛠️

For the current day, change which hours we show:
- show through 6 AM the next day, if the current local time is before midnight
  > Eg:
  > If it is currently 8pm on a Thursday, the hours will range from 9pm Thursday through 6am Friday. In this case, we are adhering to NWS day rules, and "now" is Thursday.
- show through 6 AM the third day, if the current local time is after midnight
  > Eg:
  > If it is currently 2am on a Thursday, the hours will range from 3am Thursday through 6am Friday. In this case, we are effectively treating "now" as Thursday, even though by NWS rules it's still Wednesday.

---

- closes #999 
- closes #1000

### DIFF
--- a/web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
+++ b/web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\weather_blocks\Plugin\Block;
 
-use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Logger\LoggerChannelTrait;
 
 /**
@@ -21,54 +20,22 @@ class HourlyForecastBlock extends WeatherBlockBase
     /**
      * {@inheritdoc}
      */
-    public function blockForm($form, FormStateInterface $form_state)
-    {
-        $form = parent::blockForm($form, $form_state);
-        $config = $this->getConfiguration();
-        $max = $config["max_items"] ?? "12";
-
-        $form["max_items"] = [
-            "#type" => "textfield",
-            "#title" => "Maximum items to display",
-            "#default_value" => $max,
-        ];
-
-        return $form;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function blockSubmit($form, FormStateInterface $form_state)
-    {
-        parent::blockSubmit($form, $form_state);
-
-        $this->setConfigurationValue(
-            "max_items",
-            $form_state->getValue("max_items"),
-        );
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function build()
+    public function build($now = false)
     {
         $location = $this->getLocation();
 
         if ($location->grid) {
             $grid = $location->grid;
 
-            $config = $this->getConfiguration();
-            $max = $config["max_items"] ?? "12";
-
             try {
                 $data = $this->weatherData->getHourlyForecastFromGrid(
                     $grid->wfo,
                     $grid->x,
                     $grid->y,
+                    $now,
                 );
-                $data = array_slice($data, 0, $max);
+
+                $data = $this->weatherData->filterHoursToSingleDay($data);
 
                 // Also retrieve any alerts that overlap with
                 // the given hourly periods
--- a/web/modules/weather_blocks/src/Plugin/Block/Test/EndToEnd/DailyForecast/HourlyData.php.test
+++ b/web/modules/weather_blocks/src/Plugin/Block/Test/EndToEnd/DailyForecast/HourlyData.php.test
@@ -40,29 +40,90 @@ final class DailyForecastWithHourlyDataTest extends EndToEndBase
     }
 
     /**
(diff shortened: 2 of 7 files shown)


web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
```diff
@@ -21,54 +20,22 @@ class HourlyForecastBlock extends WeatherBlockBase
...
-            "#title" => "Maximum items to display",
-            "#default_value" => $max,
-        ];
-
-        return $form;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
```
eric-gade, line 41: I guess part of the reason we even had this configurable was for the testing pages, right? This is just me thinking aloud that we don't really use the testing pages so maybe should revisit similar code elsewhere?

web/modules/weather_blocks/src/Plugin/Block/Test/EndToEnd/DailyForecast/HourlyData.php.test
```diff
@@ -40,29 +40,90 @@ final class DailyForecastWithHourlyDataTest extends EndToEndBase
...
-        $start = new \DateTimeImmutable();
-        $start = $start->setTimezone(new \DateTimeZone("America/Chicago"));
-
-        $hour = (int) $start->format("H");
-
-        $end = $start->setTime(6, 0, 0, 0);
-        if ($hour >= 6) {
-            $end = $end->modify("+1 day");
-        }
+        // We will look at the weather at 2am point-local time
```
eric-gade, line 49: Looks to me like 9:15 PM point local time, right? Or am I missing some context where the actual point-local time is something else?

web/modules/weather_data/src/Service/HourlyForecastTrait.php
```diff
@@ -55,6 +55,32 @@ protected function getForecastPropertyAsHourlyPeriods($all, $propertyName)
...
+    {
+        $times = array_column($hours, "time");
+        $firstTime = explode(" ", $times[0]);
+
+        $first6am = array_search("6 AM", $times);
+        if ($first6am === 0) {
+            $first6am = array_search("6 AM", array_slice($times, 1));
+        }
+
+        if ($firstTime[1] === "AM" && $firstTime[0] < 6) {
```
eric-gade, line 68: Will PHP implicitly convert the string "6" to an int? If not we might want to do it explicitly here so the comparison doesn't somehow blow up right?

eric-gade (approved):
Looks good 👍 
  
I also tested here in the morning by checking Honolulu, which gives the three periods. All the times looked correct for all the days there.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_data/src/Service/HourlyForecastTrait.php
greg-does-weather: It does implicitly cast for the comparison, but you're right, we should be explicit anyway. Change incoming!

web/modules/weather_blocks/src/Plugin/Block/Test/EndToEnd/DailyForecast/HourlyData.php.test
greg-does-weather: Nope, I think that's a copy-pasta mistake.

web/modules/weather_blocks/src/Plugin/Block/HourlyForecastBlock.php
greg-does-weather: Originally we thought we'd configure the number of hours to display via the Drupal UI. While we were using Drupal's block layout, we technically could, but once we stopped using that, this config became meaningless. It was just cruft sitting around, so I took out the cruft.

We're also using block config for testing pages, where we can configure the blocks with specific WFO grid points. I think there ***could*** be a use there for state and county pages, and since we've already implemented it, we might as well leave it there for now. Once we get to those, then we'll know whether block configuration is actually useful for us.

Issue #1028, opened 2024-04-09
Document all our modules


loganmcdonald-noaa:
We should start a markdown file in documentation where we document all the modules we've chosen to use, kind of like an ADR collection for modules specifically. We can also use the document to show criteria for choosing a module. 

See how we can leverage this documentation for CM controls.

PR #1073, opened 2024-04-17
Document our contributed modules

loganmcdonald-noaa:
## What does this PR do? 🛠️
#1028 

Not sure how useful the Github action actually is, and it won't catch dev-only modules. Those aren't the most important, but still nice to have information on.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,6 @@
         "drupal/core-composer-scaffold": "^10.2",
         "drupal/core-project-message": "^10.2",
         "drupal/core-recommended": "^10.2",
-        "drupal/devel": "^5.1.2",
         "drupal/key_asymmetric": "^1.1",
         "drupal/new_relic_rpm": "^2.1",
         "drupal/s3fs": "^3.1",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -22,7 +22,6 @@
         "drupal/core-composer-scaffold": "^10.2",
         "drupal/core-project-message": "^10.2",
         "drupal/core-recommended": "^10.2",
-        "drupal/devel": "^5.1.2",
```
loganmcdonald-noaa, line 25: I've always been confused why we have both here and in dev-only 🤔 . Now that I'm thinking about it, it might not belong in dev, it might belong here but excluded via config_split?

greg-does-weather: I'm genuinely not sure how this stuff got into our composer file in the first place. I think it was from a scaffold and we just never changed it. That said, what you've said here makes sense. The only reason I can think of for only having `drupal/devel` in the dev-only section is so it's not even installed in prod, and thus cannot possibly be exploited. But... I dunno how much of a concern that is?

docs/dev/contributed-modules.md
```diff
@@ -0,0 +1,118 @@
...
+
+_Added in January 2024_
+
+This module allows us to be compliant with AC-02(5) "Inactivity Logout", which requires us to set a 90 minute timeout period on user sessions. This is the most used module providing this functionality. 
+
+### [`config_split`](https://www.drupal.org/project/config_split)
+
+_Added in January 2024_
+
+We were looking for a way to disable certain modules locally (namely, samlauth and autologout) which only pertain to our cloud environments. We also wanted to set different configurations for autologout timing in beta than other cloud environments, fine tuning our settings for our production environment. This also allows us to restrict certain modules to dev-use only like `devel`. We could also use the `--no-dev` flag, but since the composer install in cloud.gov is not something we directly control, this flag is not used. `config_split` is the primary module for splitting configuration across environments. It required a complete reorganization of our configuration files.
```
greg-does-weather, line 38: Oh. So cloud.gov installs it anyway. Well then, seems like config-split makes the most sense then. 😂

greg-does-weather (approved):
right on

PR #1088, opened 2024-04-18
Redirect URLs with uppercase letters to lowercase instead

greg-does-weather:
## What does this PR do? 🛠️

This is a not-ideal solution to case-insensitive URLs. Technically URLs are case-sensitive according to specification, but we don't have to treat them that way. Drupal doesn't! What Drupal *does* do, however, is use the case-sensitive version of the URL path to determine which Twig file to use and it expects the Twig filename to exactly match the case-sensitive URL path.

This PR works around that by issues an HTTP redirect if the URL path has any uppercase characters.

The primary drawback here is that if a user comes to a URL that has uppercased characters, we force them to make an additional network request before we show them anything. This may be a minor concern – we will never *generate* uppercase URLs, so someone will likely have had to manually type it that way, which seems like it ought to be an infrequent problem. (Though I can imagine an EM in a chaotic environment needing to get people's attention quickly might make the URL all caps.)

A better solution would be if we can get Drupal to lowercase what Twig file it expects, but this PR is here as a backup.

### DIFF
--- a/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
+++ b/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
@@ -10,6 +10,7 @@ use Drupal\weather_data\Service\SpatialUtility;
 use Drupal\weather_data\Service\WeatherDataService;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 /**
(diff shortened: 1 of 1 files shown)


web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
```diff
@@ -54,6 +61,11 @@ public function noop()
 
     public function serveLocationPage($lat, $lon)
     {
+        $path = $this->request->getCurrentRequest()->getPathInfo();
+        if ($path !== strtolower($path)) {
+            return new RedirectResponse(strtolower($path));
+        }
+
```
loganmcdonald-noaa, line 68: If we go this route, maybe a comment explaining why we're doing it

loganmcdonald-noaa (approved):

Issue #1019, opened 2024-04-08
Accessibility: Combo Box Accessibility Iterations


colinmurphy01:
### Description
After reviewing search component for WCAG compliance, the custom build component seemed to do the best and will be easier to customized with additional features. During the convo, the following suggestions were made:

- Maybe think about adding a field label on combo box (@kmranjo something like "type zip or city"?)
- Keep browser focus and screen reader focus on same elements - item in list 
- As user is typing, focus goes back to text field being entered
- Consider: testing for more unique breaking use cases

### Acceptance Criteria

- [ ] Field label is visible and read off
- [ ] Upon selection and enter, location page loads

kmranjo:
For the field label, I would go with "Enter city or zip code"

PR #1094, opened 2024-04-19
Combobox a11y fixes and improvements

eric-gade:
## What does this PR do? 🛠️
This PR primarily deals with #1019.

The following is addressed:
1. The ACs from the ticket (immediate navigation on selection, adding new label mapped to component)
2. Resolves a lingering issue where selecting list items was not working in Safari and on some mobile browsers
3. Adds hover styles to list items in the listbox


web/themes/new_weather_theme/assets/js/components/combo-box.js
```diff
@@ -339,6 +339,11 @@ class ComboBox extends HTMLElement {
                 li.addEventListener("focus", (e) => {
                     this.cacheLocationGeodata(e.target.dataset.value);
                 });
+                li.addEventListener("mousedown", event => {
+                    // Stop the input from losing focus by
+                    // blocking normal browser behavior here
+                    event.preventDefault();
+                })
```
greg-does-weather, line 346: Oooh, is that what was happening? Fascinating!

web/themes/new_weather_theme/assets/sass/components/location-search.scss
```diff
@@ -1,11 +1,18 @@
...
+@use "uswds-core" as *;
 @forward "uswds-theme";
 @forward "uswds";
 
 .wx-combo-box {
   @extend .usa-combo-box;
 }
 
+.wx-combo-box li[role="option"] {
+  &:hover {
```
greg-does-weather, line 11:
```suggestion
  &:not(.wx-combo-box__list-option--focused):hover {
```

This change preserves the styling of the focused element when it is hovered. Otherwise, this makes the focused element unreadable when it's hovered:

image

Though perhaps a better solution would be to target the text color if the hovered item is focused. 🤷‍♂️ In any case, this is a super minor, easy change, so whatever makes the best sense.

greg-does-weather (approved):

Issue #1099, opened 2024-04-22
RADAR not loading consistantly


colinmurphy01:

colinmurphy01:
@eric-gade is this done?

eric-gade:
Yup! Closed by #1101

PR #1101, opened 2024-04-22
Radar fixes: round 2

eric-gade:
## What does this PR do? 🛠️
This PR addresses #1099. We have tried several times to fix radar loading, but keep coming across inconsistent behavior in our live environments.

There are two key changes in this PR.
  
First, we are swapping the `async` property on the radar wrapper script tag with `defer`. This allows it to take advantage of the `DOMContentLoaded` event, after which we can attempt to load the radar based on whatever the status is of the actual NWS radar library.
  
Second, we wrap initialization of the radar component in _both_ the `DOMContentLoaded` event handler _and_, if needed, the tab switch event. This is because sometimes the page loads with the current tab displayed first.

### DIFF
--- a/web/themes/new_weather_theme/assets/js/radar.js
+++ b/web/themes/new_weather_theme/assets/js/radar.js
@@ -96,12 +96,14 @@ const setupRadar = () => {
   );
 };
 
-document.addEventListener("wx:tab-switched", (event) => {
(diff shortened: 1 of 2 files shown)


greg-does-weather (approved):
Nice!

PR #1102, opened 2024-04-23
Refactor spatial loading scripts

### LINKED ISSUE #1098: Clean up and modify spatial-loading scripts
### Description

Currently, our script for loading spatial data are kind of messy, so they can be tricky to update.

Also, they always empty and completely re-populate spatial data. We should modify them so they don't do that. I propose we add a spatial metadata table with some kind of versioning information that we can use to determine whether to run updates on a per-table basis. Something like this perhaps:

```
table weathergov_geo_metadata
```

table|version|
---|---|
x|y|

Then in the code, we might put two pieces of information alongside each data source: a version, and what kind of update needs to be done (e.g., empty and rebuild, only add, only remove, etc.).

### Acceptance criteria

greg-does-weather:
## What does this PR do? 🛠️

This PR refactors our spatial loading script into multiple files so it's easier to read, and adds metadata to each data source (`version`, `table`, for now). The script uses this information to determine which data sources need to be reloaded so we do not have to reload everything every time.

closes #1098


.eslintrc.js
```diff
@@ -14,38 +14,36 @@ module.exports = {
...
         "class-methods-use-this": 0,
       },
     },
     {
-      files: [
-        "tests/a11y/**/*.js",
-        "tests/e2e/**/*.js",
-        "tests/load-times/**/*.js",
-      ],
+      files: ["tests/**/*.js", "spatial-data/**/*.js"],
```
greg-does-weather, line 24: I'm not sure why these two were given separate overrides. 🤷‍♂️

PR #1103, opened 2024-04-23
Load forecast and fire weather zones

greg-does-weather:
## What does this PR do? 🛠️

Adds forecast and fire weather zones to our spatial database


loganmcdonald-noaa (approved):
works for me! I can run this in all environments when it's merged - is there any sequencing to take into consideration? can we run this just fine in beta?

greg-does-weather:
I think it should be fine to just run it. The first time, it might reload all spatial data because the metadata table added in #1102 won't exist yet.

We should get Shad or Corey's thumbs-up in Slack before we do it, since it could interrupt the site for a few minutes.

PR #1106, opened 2024-04-24
Hazard maps

greg-does-weather:
## What does this PR do? 🛠️

Adds maps to alert accordions.

- adds Leaflet with ESRI plugins so we can use the same base map that we use for radar
- adds alert utility method to generate a single combined polygon for any given alert
  - if the alert already has a polygon, just use that
  - if the alert covers one or more zones, fetch their geometries from our database (available since #804), union them into a single geometry, and do geometric simplification to 0.03º resolution
  - if the alert does not have a geometry or any associated zones, that's probably an API bug. but...
    - if the alert has county SAME codes, do the same thing as with zones, except fetch the county geometries first instead of zone geometries
- updates the data layer alert methods to return alerts with full geometries
- updates the alert list twig template to add alert geometries as a data property
- adds new Javascript to load the hazard map
  - hazard maps are two layers: ESRI base map and a GeoJSON polygon
  - hazard maps also get an icon placed at the page's lat/lon
  - hazard maps only attempt to display when:
    - the Leaflet and ESRI libraries have loaded
    - the alert tab is open
    - a specific alert's accordion is open

I leaned heavily on @eric-gade's radar loading fixes. High-five for already having solved that!

image


web/modules/weather_data/src/Service/SpatialUtility.php
```diff
@@ -45,4 +45,39 @@ public static function geometryArrayToWKT($geometry)
...
+        // then we are ready to swap them.
+        if (count($point) === 2 && count($numbers) === 2) {
+            return [$point[1], $point[0]];
+        } else {
+            // If every element of the incoming array is itself an array, then
+            // we're ready to recurse on those arrays.
+            $arrays = array_filter($point, "is_array");
+            if (count($point) === count($arrays)) {
+                for ($i = 0; $i < count($point); $i += 1) {
+                    $point[$i] = SpatialUtility::swapLatLon($point[$i]);
```
eric-gade, line 73: it just recurred to me that you've been _scheme_-ing to do this for a while...

web/themes/new_weather_theme/assets/js/components/alertMap.js
```diff
@@ -0,0 +1,118 @@
...
+
+        if (expandedNow) {
+          setupMap(alert);
+          observer.disconnect();
+        }
+      });
+      mutationObserver.observe(button, { attributes: true });
+    }
+  }
+};
```
eric-gade, line 78: Would be nice if these accordions dispatched some kind of custom event. Anyway, to my reading this seems functionally equivalent to adding a click handler to the accordion, doing all the stuff, and then removing the handler when it's done the first time. Unless, of course, something besides a click opens the toggle...

web/themes/new_weather_theme/assets/js/components/alertMap.js
```diff
@@ -0,0 +1,118 @@
...
+    // keep doing this until we're done.
+    document
+      .querySelector("[data-wx-leaflet]")
+      .addEventListener("load", checkForLeaflet);
+    document
+      .querySelector("[data-wx-leaflet-esri]")
+      .addEventListener("load", checkForLeaflet);
+    document
+      .querySelector("[data-wx-leaflet-esri-vector]")
+      .addEventListener("load", checkForLeaflet);
```
eric-gade, line 114: Nice, never knew about / seen this before

web/themes/new_weather_theme/new_weather_theme.libraries.yml
```diff
@@ -69,6 +69,23 @@ radar:
...
+          crossorigin: /
+  js:
+    https://unpkg.com/leaflet@1.9.4/dist/leaflet.js:
+      data-wx-leaflet: true
+    https://unpkg.com/esri-leaflet@3.0.12/dist/esri-leaflet.js:
+      data-wx-leaflet-esri: true
+    https://unpkg.com/esri-leaflet-vector@4.2.3/dist/esri-leaflet-vector.js:
+      data-wx-leaflet-esri-vector: true
+    assets/js/components/alertMap.js:
+
```
eric-gade, line 88: It's not strictly necessary, but you _could_ add the `defer` attribute here I think

eric-gade (approved):
Merge it! 👍   

the lat/lon standards issue == jokes  

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/themes/new_weather_theme/new_weather_theme.libraries.yml
greg-does-weather: `defer` unblocks page rendering, right? That's probably a good thing to do!

Issue #1108, opened 2024-04-25
Bug: s3 configuration being overridden


loganmcdonald-noaa:
We were seeing a bug in beta where images on weather stories were disappearing. Upon further investigation it looks like our file storage and s3fs is misconfigured when we re-deploy, overriding previous settings and thus removing images.

loganmcdonald-noaa:
Something that I still don't completely understand is why/how this is working when the settings are not correct (people were still able to _upload_ and then uploads were disappearing).

PR #1109, opened 2024-04-25
add partial splits for two config values we'd like to change

loganmcdonald-noaa:
## What does this PR do? 🛠️
Proposal for fixing #1108. Control the system file setting and image setting for weather stories. 

Have this working in Greg's environment but should confirm in others, also upload an image with these settings and confirm it stays >6 hours.


greg-does-weather (commented):
Gonna withhold approval until after today's release so this doesn't sneak into beta before we can get good testing in staging.

greg-does-weather (approved):

PR #1114, opened 2024-04-26
Update docs related to how we merge

greg-does-weather:
## What does this PR do? 🛠️

We have decided to disable merge queues because they make simple fixes take FOREVER, and they don't really benefit us very often. Instead, we've enabled a requirement that PRs be up-to-date before they can be merged, which gains us the same peace of mind without requiring all of our PRs to be built twice.

Also we haven't used squash merging in a long time.


greg-does-weather:
Builds will skip correctly after #1110 is merged, It hink

loganmcdonald-noaa (approved):

PR #1119, opened 2024-04-29
Copy WFO Promo preview styling from display templates

### LINKED ISSUE #1093: Preview of WFO promo doesn't appear to work
**Expected behavior**
1. Author the WFO promo
2. Click the Preview button
3. See the content as it would be displayed to a front end user when published 

**Current behavior**
Preview is not rendering as a styled component 

Image

greg-does-weather:
## What does this PR do? 🛠️

This PR moves the templates we use for displaying WFO promos into node and field templates. It also updates the WFO promo block to return a node ID. The WFO promo template finally is updated to use a Twig Tweak method to render the node. This way, we are using the same templates for previewing and displaying the WFO promo content, so that the two places will look the same.

closes #1093


loganmcdonald-noaa:
Does this mean we'll have to maintain the templates in two places?

greg-does-weather:
Yeah, for now at least. The way data gets into the two things is pretty different, so they're not interchangeable. We might be able to refactor how the `WFOPromo` block returns data so we can reuse the same templates but I'm not sure.

greg-does-weather:
Okay, now using the same templates in both places. That wasn't bad at all.

web/config/sync/block.block.new_weather_theme_pagetitle.yml
```diff
@@ -21,4 +21,4 @@ visibility:
   request_path:
     id: request_path
     negate: true
-    pages: '<front>'
+    pages: "<front>\r\n/point/*\r\n/node/*"
```
greg-does-weather, line 24: We don't want the built-in page title for our previews because they just duplicate the content itself. There might be an accessibility concern here, though: without an H1 anywhere, there's no hierarchy. We could add an H1 in our templates somewhere to mitigate this, but I'm not sure where.

web/modules/weather_blocks/src/Plugin/Block/WFOPromoBlock.php
```diff
@@ -28,39 +28,10 @@ public function build()
...
-                    $social["youtube"] = $youtube;
-                }
-
                 return [
-                    "name" => $wfoName,
-                    "code" => $grid->wfo,
-                    "shortIntro" => $shortIntro,
-                    "phone" => $phone,
-                    "social" => count($social) ? $social : false,
+                    "node" => $promo->id(),
```
greg-does-weather, line 34: Instead of pulling all that content out ourselves, just pass the node ID into the template. We'll use a Twig tweak helper to render the node.

web/themes/new_weather_theme/templates/block/block--weathergov-wfo-promo.html.twig
```diff
@@ -1,48 +1 @@
...
-          </a>
-        </p>
-        {% endif %}
-
-      {% endif %}
-    </div>
-  </div>
-</div>
-
+{{ drupal_entity("node", content.node) }}
```
greg-does-weather, line 1: Here's were we tell Twig to render the node. That'll pass us off to the `node--wfo-promo.html.twig` template.

web/themes/new_weather_theme/templates/content/node--wfo-promo.html.twig
```diff
@@ -0,0 +1,102 @@
...
+ * @see template_preprocess_node()
+ *
+ */
+#}
+
+<div class="grid-container padding-x-2">
+  <div class="grid-row">
+    <div class="grid-col tablet-lg:grid-offset-2 tablet-lg:grid-col-8">
+      <h2 class="text-normal text-primary-darker"> Your Local Forecast Office </h2>
+      <h3 class="margin-bottom-1">{{ content.field_wfo }}</h3>
```
greg-does-weather, line 75: Instead of using `{{ content }}` to have Twig render everything itself, you can instead use `{{ content.field_machine_name }}` to render a specific field wherever you want it.

Fields can have their own templates, even tied to specific content types.

web/themes/new_weather_theme/templates/block/block--weathergov-wfo-promo.html.twig
eric-gade: nice, this has turned out to be a sweet plugin

web/themes/new_weather_theme/templates/content/node--wfo-promo.html.twig
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @see template_preprocess_node()
+ *
+ */
+#}
+
+<div class="grid-container padding-x-2">
+  <div class="grid-row">
+    <div class="grid-col tablet-lg:grid-offset-2 tablet-lg:grid-col-8">
+      <h2 class="text-normal text-primary-darker"> Your Local Forecast Office </h2>
```
eric-gade, line 74: Prob want to run through translator
```suggestion
      <h2 class="text-normal text-primary-darker"> {{ "Your Local Forecast Office" | t }} </h2>
```

web/config/sync/block.block.new_weather_theme_pagetitle.yml
eric-gade: We could consider an overall h1 called "Preview" at the top of all preview pages

eric-gade (approved):
Code LGTM 👍 
  
Not sure if we've done another styling pass or not, but the bullets are being listed with the social links. Is that what we want? Example:
Screen Shot 2024-05-01 at 9 55 49 AM


## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/config/sync/block.block.new_weather_theme_pagetitle.yml
greg-does-weather: Yesss, smart. I'll add that to the headings overhaul PR.

greg-does-weather:
cc: @igorkorenfeld What do you think about the styling on the social links? Do we want to tweak those now, or do another pass later?

partly-igor:
@greg-does-weather it looks pretty close otherwise, I leant toward resolving in this one. The curious thing is that it does `usa-list--unstyled` and I am not sure why it isn't working here.

partly-igor:
Oh, my guess it that each of those social media items needs a `class="usa-list usa-list--unstyled"` on it's individual `ul`

PR #1125, opened 2024-04-29
When navigating directly to a location page, save that location into recent results

### LINKED ISSUE #1107: Recent Locations: Add all recent locations to list
### Currently

- Visit beta
- Click use my location , it loads Shaker Heights
- I click the combo box to look up Washington DC
  - Shaker doesn’t come up as a recent location
- I load washington DC
- I click the combo box to look up Chicago
  - Washington DC shows up, Shaker Heights doesn’t show up

### Acceptance criteria

- [ ] Add any location page you visit to the list of recent locations, regardless of how you got there (via my location, search, or landed on the page)

greg-does-weather:
## What does this PR do? 🛠️

- adds a `data-place` attribute to the search combo box
- when the combo box loads, if its `data-place` attribute is set, saves the current page to recent results
- closes #1107

### DIFF
--- a/web/themes/new_weather_theme/assets/js/components/combo-box.js
+++ b/web/themes/new_weather_theme/assets/js/components/combo-box.js
@@ -186,6 +186,13 @@ class ComboBox extends HTMLElement {
     this.input.addEventListener("focus", () => {
       this.updateSearch("");
     });
+
+    if (this.dataset.place) {
+      this.saveSearchResult({
+        text: this.dataset.place,
+        url: window.location.href,
+      });
+    }
   }
 
   disconnectedCallback() {
--- a/web/themes/new_weather_theme/templates/block/block--weathergov-location-search.html.twig
+++ b/web/themes/new_weather_theme/templates/block/block--weathergov-location-search.html.twig
@@ -1,6 +1,7 @@
 {{ attach_library("new_weather_theme/location-search") }}
 {{ attach_library("new_weather_theme/location-combo-box") }}
 {{ attach_library("new_weather_theme/browser-location-button") }}
+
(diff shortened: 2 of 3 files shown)


web/themes/new_weather_theme/assets/js/components/combo-box.js
```diff
@@ -186,6 +186,13 @@ class ComboBox extends HTMLElement {
     this.input.addEventListener("focus", () => {
       this.updateSearch("");
     });
+
+    if (this.dataset.place) {
```
eric-gade, line 190: ezgif-7-0ab5d96eaf

eric-gade (approved):
Working 👍 
  
Only one issue: if I type a location into the search and then navigate to it, it will save that location into storage twice (sometimes with slightly different names). This example was after just typing buffalo and selecting the first option, when on the front page:
Screen Shot 2024-05-01 at 10 11 33 AM
Screen Shot 2024-05-01 at 10 11 55 AM
  
I suspect this is because of a difference between Arc place names and our own GIS place names?
  
That said, since we are filtering recent results I'm not sure how much of a problem it actually is, though it might be slightly confusing.


## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

greg-does-weather:
That's a big ol' bug where it's putting the same place in twice. It's supposed to be filtering on URL, but in one case it's using the whole URL and in the other it's only the path. Will fix that!

PR #1131, opened 2024-04-29
Add landing page design

partly-igor:
## What does this PR do? 🛠️
This PR adds the implementation for the landing page design for beta.weather.gov with content and text that were agreed to in our working sessions and discussion over the last sprint. 

### Front page
The template logic relies on the `is_front` flag of the page template. In order for the `is_front` to change to `true` for the current front page, I had to "touch" the configuration for the Front Page (under Basic Configuration). I temporarily switched it to something else and then back to `/home`. We will likely need do that on the beta site or otherwise figure out why `is_front` for `/home` is initially `false` in the current setup. 

### Beta banner text
I decided not to export the local content as part of this PR because there is test data I've put in, and it's easier to just re-enter the beta banner content on beta.

### Utilities
I've added a few more custom, responsive utility classes. For these I've kept the pattern of separating the breakpoint size prefix with a double underscore (e.g. `tablet__`) to hint that these are different from the standard USWDS classes, which use a colon as the delimiter. Not sure if this is a good idea or if we should just use the colon for consistency. 

More broadly we may want to move more of these custom utilities into a separate scss file. 

Screenshot of the landing page on mobile

Screenshot of the landing page on desktop


greg-does-weather (approved):
This looks great. Like, really great.

One note for the future: unless the location search on the front page needs to be drastically different from elsewhere, we'll probably want to update the block so we can use it in both places rather than maintain two sets of templates. I don't see any point in dealing with that right now, but it was the only thing that made me pause.

partly-igor:
Thank you! 

That makes a lot of sense, the primary difference is the sizing and color of the text, so 
we would just need to account for that. Maybe we can include options for that or use a "mode" class to help us with it.

colinmurphy01:
It's not showing up on staging, so I guess that's the config bug you were talking about @igorkorenfeld

Issue #1059, opened 2024-04-15
Accessibility: Combobox ARIA description 


partly-igor:
### Description

WAVE is reporting an error for the location search combobox:
> An aria-labelledby or aria-describedby reference exists, but the target for the reference does not exist.

Currently the combobox has `aria-describedby="combo-box-1--input--description"` , but there is no element with and `id` of `combo-box-1--input--description`

### Acceptance criteria

- [x] There is an element that matches the `id` of the `aria-describedby` attribute of the location search combobox, which properly describes the combobox. 
- [x] Changes have been tested for accessibility

eric-gade:
Starting work on this, and I'm wondering why our a11y tests didn't pick it up

eric-gade:
Closed by #1141

PR #1141, opened 2024-05-01
Combo box aria description

eric-gade:
## What does this PR do? 🛠️
Deals with #1059 where there wasn't a matching element with aria-descrption text.

I _think_ this was pretty straightforward? But am slightly suspicious.

Screen Shot 2024-05-01 at 12 52 33 PM

### DIFF
--- a/web/themes/new_weather_theme/assets/js/components/combo-box.js
+++ b/web/themes/new_weather_theme/assets/js/components/combo-box.js
@@ -226,7 +226,6 @@ class ComboBox extends HTMLElement {
     input.setAttribute("autocorrect", "off");
     input.setAttribute("spellcheck", "false");
     input.setAttribute("aria-activedescendant", "");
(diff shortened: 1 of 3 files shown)


partly-igor:
It seems like this might be creating a duplicate text description that's read after the screen reader focus has moved away from the combo box

Screenshot of voiceover text moved after the combo box

partly-igor:
Duplicate in the sense that it's similar to "Enter city or zip code"

eric-gade:
@partly-igor I'm wondering: do we actually need an `aria-describedby` attribute on the input at all? Seems like the label above the component does the trick on its own, plus most screenreaders will have general guidance for the combobox role anyway

partly-igor:
I think you are right we may not need it all.

greg-does-weather (approved):

Issue #1075, opened 2024-04-17
Change alert accordions to be open by default


partly-igor:
In R3 we found that users were missing the second alert accordion, which was collapsed and sat between the top alert accordion and the safety information. We'd like to try having all accordions expanded by default, rather than only the first one being expanded. 

AC
- [ ] All alerts accordions on the alerts tab are expanded
- [ ] Alerts accordions are still collapsable

eric-gade:
Closed by #1142

PR #1142, opened 2024-05-01
All alerts open by default

eric-gade:
## What does this PR do? 🛠️
Addresses #1075 and makes all alert accordions open by default.

I've also added e2e tests for this


mgwalker (approved):
Oh thank goodness.

greg-does-weather (approved):
I agree with me.

PR #1143, opened 2024-05-01
Change how we determine when to load the front page layout

greg-does-weather:
## What does this PR do? 🛠️

There's an old, old, old, old, old [Drupal issue](https://www.drupal.org/project/drupal/issues/1503146) where the `is_front` variable doesn't work if your site settings refer to the front page by a URL alias, which is exactly what we do because we don't really want it tied to a particular content ID since that's inconsistent across environments.

This PR changes the check from `is_front` to instead look at the page path. If the page is `/home` (which is our alias to the front page), then it loads the front page layout.


partly-igor (commented):
Looks great!

partly-igor (approved):

Issue #972, opened 2024-03-27
Adjust Hourly Table Scroll


partly-igor:
### Description
In R3 we heard users express some frustration with the scroll interaction on the hourly table. As an initial step we propose adjusting the amount the table scrolls on button press. Engineering and design will need to pair to figure out the right amount, and  this will need to be dependent on the screen size (i.e. based on how many hours are displayed in the viewport), 

### Acceptance Criteria
- [ ] Clicking on the left and right scroll buttons will scroll more than 1 hour
- [ ] Design and engineering pair to determine how much more

colinmurphy01:
Related question: what's the ideal minimum amount of hours a user wants to view?

eric-gade:
Closed by #1159

PR #1159, opened 2024-05-06
Hourly table scrolling

eric-gade:
## What does this PR do? 🛠️
This PR implements the changes in #972, which called for the hourly table scroll buttons to scroll more than just one column at a time.

Instead of scrolling one column, each direction will attempt to scroll one full "page" (ie width of the table) view in the relevant direction. It will then attempt to align on the nearest intersecting column on the left, so that the whole column is visible.

### DIFF
--- a/web/themes/new_weather_theme/assets/js/components/HourlyTable.js
+++ b/web/themes/new_weather_theme/assets/js/components/HourlyTable.js
@@ -42,6 +42,9 @@ class HourlyTable extends HTMLElement {
   }
 
   clickScrollRight(wrapper) {
+    const containerWidth = wrapper.getBoundingClientRect().width;
+    const firstColumnWidth = this.querySelector("table tr:last-child th").getBoundingClientRect().width;
+    const visibleWidth = containerWidth - firstColumnWidth;
     const cellsToRight = Array.from(
       this.querySelectorAll("table tr:last-child td"),
     ).filter((cell) => {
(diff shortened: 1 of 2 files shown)


partly-igor:
Testing out the scroll it looks good on desktop and tablet, but on mobile sizes < 343px wide the right scroll button stops working after the first click.

greg-does-weather:
I can imagine what's happening, but I'm impressed you were testing at such a narrow viewport!

partly-igor:
I have been using 320px as my smallest viewport when designing so decided to give it a look

eric-gade:
@partly-igor give it a shot now. I _think_ I have the right fix...

partly-igor:
Nice, looks to be working at those viewport sizes now as well

web/themes/new_weather_theme/assets/js/components/HourlyTable.js
```diff
@@ -42,32 +42,52 @@ class HourlyTable extends HTMLElement {
...
+    const visibleWidth = containerWidth;
+    const rightSide = wrapper.scrollLeft + visibleWidth;
 
-    if (cellsToRight.length) {
+    const nextCol = Array.from(this.querySelectorAll("table tr:last-child td"))
+          .find(el => {
+            const left = el.offsetLeft;
+            const right = el.offsetLeft + el.offsetWidth;
+            return left <= rightSide && right >= rightSide;
+          });
```
greg-does-weather, line 55: Do I understand this correctly? The "next column" is the one whose left side is inside (or on the border of) the current viewable area and whose right side is outside (or on the border of) the current viewable area?

web/themes/new_weather_theme/assets/js/components/HourlyTable.js
```diff
@@ -42,32 +42,52 @@ class HourlyTable extends HTMLElement {
...
+            const left = el.offsetLeft;
+            const right = el.offsetLeft + el.offsetWidth;
+            return left <= rightSide && right >= rightSide;
+          });   
+
+    if (nextCol) {
+      let nextLeftPos = nextCol.offsetLeft - firstColumnWidth - 16;
+      if(Math.floor(nextLeftPos) <= Math.floor(wrapper.scrollLeft)){
+        nextLeftPos = nextCol.nextElementSibling.offsetLeft - firstColumnWidth - 16;
+      }
```
greg-does-weather, line 61: And if scrolling to this position would actually cause us to scroll back to the left, instead move to the next element and use that as our scroll basis?

greg-does-weather (approved):
Just a couple questions to make sure I understand the logic, but it's definitely working.

web/themes/new_weather_theme/assets/js/components/HourlyTable.js
eric-gade: Yes, that's correct

PR #1168, opened 2024-05-06
Fix mis-labeling of today in rare case where it is currently after midnight but API data for "today" still begins before midnight

### LINKED ISSUE #1151: 7-day tab skipping days
### Description

I was looking at a Texas forecast spot last night on my phone, and I noticed that it skipped days from May 2 (current day) to May 4.  I wanted to bring this up in case we were able to fix this.  I entered the location again a few minutes later and it showed the correct days (didn't skip).  I've seen this behavior a few times, but I've only noticed it over the past week.
image

### Acceptance criteria

No days will be skipped when looking at a 7-day forecast.

greg-does-weather:
## What does this PR do? 🛠️

When querying the API for daily forecast data, it returns data in 12-hour periods, except for the first period which begins at the start of the current hour and goes up to the next 6:00. Thus, the first period can be 12 or fewer hours.

However, when the hour changes, the API does not immediately begin returning updated hourly forecast periods. This becomes quirky at midnight. We build the day name and overnight/day/night period descriptions based on the start time of a given forecast period. If the actual time is just after midnight and the API is still returning a first daily forecast period that begins at 11pm the previous day, we will mis-label that day as the previous day and we'll mis-label the overnight period as just "night."

This PR fixes that issue. Instead of relying strictly on the forecast period start time, we now use the actual time to determine the day name. We also use a combination of the current time and the forecast period ending time to determine whether the period is "overnight."

- fixes #1151


web/modules/weather_data/src/Service/DailyForecastTrait.php
```diff
@@ -59,10 +71,17 @@ private function formatDailyPeriodForToday($period, $timezone = null)
...
+        $overnightEnd = $now->setTime(6, 0);
+
+        // This is an overnight period if the current time is after midnight and
+        // the period ends on or before 6am of the same day.
+        //
+        // If now is before midnight, this must either be a day or night period.
+        //
+        // If now is after midnight and this period ends after 6am, then it must
+        // also be a day or night period.
+        $isOvernightPeriod = $now >= $midnight && $endTime <= $overnightEnd;
```
greg-does-weather, line 84: I need to think through this some more. I believe as-written, `$now` will by definition always be after `$midnight`. 🤔 Maybe it should be:

```php
$isOvernightPeriod = intval($now->format("G")) <= 6 && $endTime <= $overnightEnd;
```

That is, if it is currently between midnight and 6am and this period ends at or before 6am, then this period must be overnight.

greg-does-weather: So it's not strictly necessary because both the API and our own code should have filtered out the potential overnight period once the actual time reaches 6am, but there's no harm in a little extra defensive coding.

eric-gade (approved):
Confused myself about timezones at first (since we initially pulled it from the startTime in the forecast period), but I get it now!
## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

greg-does-weather:
Are you *really* programming if you're not confused about timezones?

PR #1171, opened 2024-05-07
Log errors to stdout

### LINKED ISSUE #871: Add error logging
### Description

After discussion among the engineering team, we decided that we don't really need database error logging, per se. We haven't seen any database errors so far. But we also can't be sure because we have very poor error logging in general, so the first step should be to log our errors. Once we have that data, we may be able to determine whether we need more specific error handling.

We currently "eat" most of the exceptions we encounter. A small number are logged to Drupal Watchdog, but most are just caught and ignored. This mostly serves us fine, but when things go wrong, it can be hard to figure out _why_. We should instead catch and log these errors.

We will install the [log_stdout module](https://www.drupal.org/project/log_stdout) that redirects Watchdog logs to the standard out and error handles, where they will be captured by cloud.gov logging. This will also incidentally make them easier to access in local development.

We will also update our error handling so that caught errors are logged to Watchdog.

### Acceptance criteria

- [ ] Watchdog logs are sent to stdout and/or stderr
- [ ] stdout and stderr logs are captured by cloud.gov logging
- [ ] all caught errors are logged to Watchdog

---

Original description:

### ~Description~

~We have currently not experienced any database errors, but connection disruptions or (heaven forbid) data corruption could eventually cause them to happen. We should spend some time thinking about what we want to do in those situations. Currently any errors would bubble up to the blocks, which will quietly eat them and produce the more user-friendly error boxes, but we might want to think about fallbacks (e.g., if a database query fails, then perhaps we fallback to the API; or maybe we put in some wait-and-retry logic).~

### ~Acceptance criteria~

- ~We have determined how we should handle database errors~
- ~We have implemented database error handling~

greg-does-weather:
## What does this PR do? 🛠️

- adds [`log_stdout`](https://www.drupal.org/project/log_stdout) module
- updates our block error handlers to log their errors
- adds error handling to the base block
- fixes some PHP deprecation warnings that I can see now that logs are sent to stdout

closes #871


greg-does-weather:
Deployed to my environment, and it sure looks like our application logs are showing up in cloud.gov now

loganmcdonald-noaa (approved):
Awesome!

PR #1172, opened 2024-05-07
update drupal core and other modules

loganmcdonald-noaa:
Drupal core was updated May 1st to 10.2.6, so updating that. Also updating some packages.


greg-does-weather (approved):
OH BOY!

PR #1173, opened 2024-05-07
Add and configure metatag module

loganmcdonald-noaa:
## What does this PR do? 🛠️
We want to be able to manage our metatags and the best module to do it is metatag! 

Also updates our favicon to be NWS not NOAA. 

🚧 🚧 🚧 

Before merging we should: 
- [x] finalize copy with product/design and decide whether to replace logo with actual banner
- [x] do a walkthrough with design of how to use metatag
- [x] ensure there are no errors in drupal related to serving public files

After merging we should: 
- [ ] Add new metatag setting for the frontpage node specifically, changing the title to `beta.weather.gov | National Weather Service` for the basic tag page title, export this new content

New banner: 
Screenshot 2024-05-08 at 3 52 42 PM

Facebook/OpenGraph:
Screenshot 2024-05-08 at 3 53 49 PM

Twitter:
Screenshot 2024-05-08 at 3 53 54 PM

Slack:
Screenshot 2024-05-08 at 3 53 06 PM


loganmcdonald-noaa:
Ok only one problem here remains that is super tricky to figure out. We want the frontpage to have a different metatag for title than everything else. It should be "beta.weather.gov | National Weather Service" whereas every other node for basic pages should be "[current-page:title] | beta.weather.gov". Metatag let's us have a different configuration for the frontpage but it identifies it by node and we are currently using an alias (/home) because the node is variable to change by environment depending on what has been created in the past content-wise.

loganmcdonald-noaa:
This is working now! Solved by enabling per-node meta tags. This means for the basic pages we have (seven of them) we'll want to do a re-import of that content.

web/config/sync/core.entity_form_display.node.page.default.yml
```diff
@@ -31,6 +33,14 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
+  field_meta_tags:
+    type: metatag_firehose
```
greg-does-weather, line 37: 😳

greg-does-weather (approved):
What is involved in this?

> This is working now! Solved by enabling per-node meta tags. This means for the basic pages we have (seven of them) we'll want to do a re-import of that content.

Will we need to delete the existing basic pages and import from the new zip file?

That aside...
I'm gonna ship it

PR #1174, opened 2024-05-07
Filter out ended/expired alerts

### LINKED ISSUE #1157: Expired or concluded alerts should be filtered out
### Description

We currently do not filter out alerts that may be expired. The API has never sent us such an alert, but we should be defensive about it anyway. We have a few places in code where we check for concluded alerts (the alert duration text, for example) that can serve as guides for how to build the filters.

Testing should be pretty straightforward: duplicate an existing alert and hardcode its time to the past. Then make sure it's not there.

### Acceptance criteria

- [ ] Changes have been implemented to make the tests pass

greg-does-weather:
## What does this PR do? 🛠️

If an alert's end time – or expired time, if there is no end time – has past, filter it out.

fixes #1157


tests/api/data/e2e/alerts/active__status=actual&area=AL.json
```diff
@@ -47,7 +47,7 @@
         "effective": "2024-01-08T13:48:00-06:00",
         "onset": "date:now -1 hour",
         "expires": "date:now +1 hour",
-        "ends": "date:now +0 hour",
+        "ends": "date:now -1 minute",
```
greg-does-weather, line 50: This was working entirely by accident. In our tests, we compared it to "now," and the only reason the alert ending time was after "now" is because there were a few milliseconds between when we initiated the API request and when our proxy parsed these tokens into a date. This change makes it explicit and removes the narrowly-timed, almost race condition.

PR #1179, opened 2024-05-08
add an init command for setting isolation level

loganmcdonald-noaa:
See https://www.drupal.org/node/3264101


greg-does-weather:
I'll deploy this to my environment just to double-check that everything works correctly.

greg-does-weather:
Well, the site works, but I can't log into it anymore to see if the database message is gone. 😂 As long as the site's working, we can deal with status messages on the admin side later, if necessary.

greg-does-weather (approved):

PR #1184, opened 2024-05-09
Style adjustment to hourly details 

partly-igor:
## What does this PR do? 🛠️
This PR makes two adjustments to the hourly details sections: 

1. We discussed the empty "Feels like" row with our POs a a couple of weeks ago looking like an error, and that seeing that the apparent temperature matches the temperature may be helpful in itself. We decided to fill in the whole row for now.

2. There were a couple of alignment/spacing inconsistencies within the table that I touched up. 

Screenshot of before and after of the alignment adjustment of the hourly details area


web/modules/weather_data/src/Service/HourlyForecastTrait.php
```diff
@@ -254,7 +254,7 @@ public function getHourlyForecastFromGrid(
 
             return [
                 "apparentTemperature" =>
-                    abs($apparentTemperature - $temperature) >= 5
+                    abs($apparentTemperature - $temperature) >= 0
```
partly-igor, line 257: I kept in the ternary if we want to edit this later, but if that's not a great practice happy to adjust it so it's just the variable `$apparentTemperature`

greg-does-weather (approved):
rock and roll

PR #1185, opened 2024-05-09
Fix how we id alerts

greg-does-weather:
## What does this PR do? 🛠️

Include more of the OID in the alert ID until we sort out what's going on with CAP.

@loganmcdonald-noaa This should unblock your PR.


loganmcdonald-noaa (approved):
Thank you!

PR #1225, opened 2024-05-20
Split tests into composite actions

greg-does-weather:
## What does this PR do? 🛠️

Splits several tests into composite actions:
- accessibility tests
- browser tests (Playwright)
- Javascript component tests
- Javascript linting
- PHP linting
- Style linting

The reasoning is that there are fewer `if` conditions related to the "should test" check. The step where the composite action is `used` can have the conditional and the composite action doesn't need any in its steps. However, composite actions have to specify a shell, so every `if` property is replaced with a `shell` property.

I think that's okay, though, because two things become true:
1. We are less likely to run tests that should be skipped since there are fewer conditionals. The only way to identify when this is happening is to see it and manually fix it.
2. GitHub will fail to run composite actions where the shell is unset, so it'll let us know right away that something is wrong.


eric-gade (approved):
I think I do like this better. Easier for me to reason about, anyway 👍

loganmcdonald-noaa (approved):
works for me, makes sense

Issue #1198, opened 2024-05-13
I want to see all unpublished content


kmranjo:
### Current state 
When viewing a content dashboard (/admin/content), users with the System Admin role are able to view all content regardless of publication status. However, all other users are unable to view any content where the publication status is unpublished. 

### Desired change
All users regardless of role should be able to view all content that they have permission to view, regardless of publication status.

loganmcdonald-noaa:
From the docs: 
```
In Drupal 7, users are not allowed to view any unpublished content by default, not even content created by themselves. By enabling 'View own unpublished content', you grant users access to their own unpublished content.

The only way to allow users access to all unpublished content in Drupal 7 (without using a contributed module), is to enable the permission 'Bypass content access control'.
Use this permission with extreme caution as it has far more consequences than just allowing access to unpublished content. In combination with Organic Groups for example, this permission allows access to private group content regardless of whether the user is a group member.

View Unpublished Content with Views Module
The Views module allows access to view data from unpublished nodes. It has a separate filter for published / unpublished status to block access if wanted.
```

Right now content editors and admins have the permission to view their own unpublished content so it is confusing that is not showing up but we will need to think about if we want to enable a module for permissions to view all unpublished content. I do not want to turn on bypassing action control.

loganmcdonald-noaa:
Confirmed that users can see content that is unpublished _if they authored it_, but as of now they cannot view unpublished content _they have not authored_. Unassigning myself as I think @eric-gade will handle this in his work on the unpublish/archive state! Correct me if I'm wrong and something else needs to be changed here.

eric-gade:
From what I can tell, once we enable the Content Moderation module we will have fine-grained control over the permissions that will cover this case.

Issue #1118, opened 2024-04-29
Unable to unpublish content


shadkeene:
### Description

I created a weather story, and when trying to unpublish (following the process in first screen grab below), I got the following error (second screen grab image below).

![image](https://github.com/weather-gov/weather.gov/assets/139174561/ff0902f4-8d1e-45f7-8171-615c72265ef0)

![image](https://github.com/weather-gov/weather.gov/assets/139174561/9e48f9e5-0c52-451f-a56d-9eaa0947c3c9)

### Acceptance criteria

Content editors and content admins should be able to unpublish content so it is not deleted from the system, but it does not show up any longer on beta.weather.gov.

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

colinmurphy01:
This might be related to V2 refinement from testing (save, preview, publish, unpublish)

loganmcdonald-noaa:
So weirdly the content editors _can_ unpublish content but only when editing that content in the view, not from this bulk action UI. Will try to figure out why this is happening, but yes with the changing flows this might go away.

PR #1254, opened 2024-05-24
Remove Publish/Unpublish bulk action, and update Actions UI

eric-gade:
## What does this PR do? 🛠️
This PR in part addresses #1118, which highlighted a permissions error when using "Bulk Actions" to set a given piece of content to Unpublished (and, as it turns out, Published).

### Drupal "published"/"unpublished" vs our moderation states ###
As we have pointed out elsewhere, we are using the Content Moderation core module in order to define our own moderation states. However, Drupal has its own binary notion of "published" or "unpublished" -- which map to "publicly viewable" or "not publicly viewable" respectively. Our moderation states can and do map to these values when needed, but they are not equivalent. The upshot here is that **users should never be able to update the Drupal Published / Unpublished status directly**. Instead, they should change the moderation states, which will handle the "viewability" of a piece of content for them.

### Patch ###
In order to properly link the Content Moderation transitions to the Bulk Actions menu, we need to add [the patch linked here](https://www.drupal.org/project/drupal/issues/2797583#comment-14814660). The Content Moderation module was only recently added to core, and I think this patch addresses a known issue when trying to use state transitions from within bulk actions.
  
### Actions UI Changes ###
In order to edit, delete, or add actions to the "Bulk Actions" menu that appears in the content administration view, we need to enable the core module Actions UI. I have done that in this PR. Additionally, I have _removed_ the bulk actions for Publish/Unpublish (which were direct Drupal versions of the status) and added an "Archive" action, which simply runs the Archive moderation state transition on the selected pieces of content.
  
## Issue #1198 Seeing unpublished content ##
I believe the new moderation permissions here also resolve issue #1198. But we will probably want @kmranjo to test this in a live environment just to be sure before closing that one.

Screenshot 2024-05-24 at 5 32 25 PM
Screenshot 2024-05-24 at 5 32 16 PM
Screenshot 2024-05-24 at 5 32 01 PM


web/config/local/devel.settings.yml
greg-does-weather, line 1: This one keeps wanting to be deleted on me, too, but I never know why. I guess if it wants to do it so badly, we might as well let it, huh? 😛

web/config/sync/system.action.node_publish_action.yml
greg-does-weather, line 1: I assume these two config yamls remove the publish/unpublish bulk actions, but how does the archive action get added? I can confirm locally that publish and unpublish are gone, but I don't see "archive":

image

web/config/local/devel.settings.yml
loganmcdonald-noaa: I think we need to be running a second import as part of our build steps after the local split is declared as active... but yes, annoying.

eric-gade:
@greg-does-weather Ok I have exported the config again, this time I hope with the archive action in bulk actions (see https://github.com/weather-gov/weather.gov/pull/1254/commits/3712d8edca656fc17126ee37592dbfe0ae7ce455)

web/config/sync/system.action.node_publish_action.yml
greg-does-weather: Resolved!

greg-does-weather (approved):

PR #1262, opened 2024-05-29
Add radar legend

### LINKED ISSUE #1204: Add legend to the radar
### Description
Our Radar component needs a legend so that viewers can better understand what the colors on the radar map mean.  

[Radar legend mockups](https://www.figma.com/proto/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?page-id=5558%3A149380&type=design&node-id=5774-222769&viewport=8%2C765%2C0.26&t=64delBci36CtJ2U5-1&scaling=min-zoom&starting-point-node-id=5774%3A222769)

[Radar legend documentation](https://www.figma.com/file/0dSABt9zCqoknaKua6DHmP/NOAA---Weather.gov?type=design&node-id=5774%3A223161&mode=design&t=fwAylhpoqYdUlCyS-1)

The legend should be added below the Radar and where where the Radar component appears – currently on the "Current" tab and on the "Alerts" tab when it appears.  

Noted in the documentation, I created an individual gradient element for each row of the table, rather than creating one single gradient. This way if the height of the row changes, the correct colors range is still contained within that row. 

### Acceptance Criteria
- [ ] Legend is added below the Radar on the "Current" tab
- [ ] Legend is added below the Radar on the "Alerts" tab

greg-does-weather:
## What does this PR do? 🛠️

- closes #1204

- the original issue mentions that the radar component appears on the alert tab, but it does not – only on the "current" tab
- the design spec calls for all the table rows to be equal height based on the tallest row, which is impossible without Javascript; for a first pass, I opted to do this purely in CSS, but if we want to equal-height rows, I can add the JS

<details>
<summary>
Desktop
</summary>

</details>

<details>
<summary>Mobile</summary>

</details>


web/themes/new_weather_theme/assets/sass/components/_radar.scss
```diff
@@ -17,3 +17,45 @@
...
+      // but it checks out.
+      min-height: 1px;
+    }
+
+    .wx-radar-legend__intensity {
+      // Use padding so that it goes inside
+      // the border, effectively pushing
+      // text content left and down from the
+      // gradient.
+      padding: 0.5rem 1rem;
```
partly-igor, line 49: We've been using USWDS units for most sizing/spacing so think it'd be worth replacing these two values with `units(1)` and `units(2)` respectively.

web/themes/new_weather_theme/assets/sass/components/_radar.scss
```diff
@@ -17,3 +17,45 @@
...
+
+      // Setup the border so we can put a
+      // gradient in it. We could try using
+      // attr() and define the background
+      // image here, but attr() is experimental
+      // for properties other than content
+      // so that doesn't seem safe yet.
+      border-left: 5px solid;
```
partly-igor, line 57: Border should be a touch thicker,  `units(1)`

web/themes/new_weather_theme/templates/partials/radar.html.twig
```diff
@@ -2,14 +2,17 @@
...
-    <h3 class="visual-h2 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3 margin-bottom-2">Radar</h3>
+    <h3 class="visual-h2 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3 margin-bottom-2">{{ "Radar" | t }}</h3>
   </div>
 
   <div class="grid-container padding-0 bg-white shadow-1 grid-col-12 tablet-lg:grid-offset-2 tablet-lg:grid-col-8">
     <div class="usa-sr-only">
-      <p>Listed below is a visual representation of radar data for
-      {{ weather.place }}. For more details on specific weather conditions,
-      view hourly data within the <a href="" data-daily-tab-link>daily tab.</a></p>
+      <p>
```
partly-igor, line 10: Thanks for thinking to include this. Rather than the paragraph here I am thinking this should be be a (sr-only) `caption` within the table itself

web/themes/new_weather_theme/assets/sass/components/_radar.scss
```diff
@@ -17,3 +17,45 @@
...
+
+  .wx-radar-legend__row {
+    th {
+      vertical-align: top;
+    }
+
+    td {
+      padding: 0;
+      vertical-align: top;
+
```
partly-igor, line 36: The top and bottom borders of the cells should be `color("base-light")` (currently they are the USWDS default)

partly-igor (changes requested):
It's looking great! Just a couple of visual changes I noticed.

For placement, in our original designs we did have Radar on both tabs so that we could always have a map even if for some reason we couldn't get one for the one inside the alert details. That is a separate issue and may need another discussion though so no need to deal with it here. 

Sounds good about that consistent sizing of the rows. I think let's keep this for now and revisit if we need to.

web/themes/new_weather_theme/templates/partials/radar.html.twig
partly-igor: oh  misinterpreted those one. You can ignore

web/themes/new_weather_theme/templates/partials/radar.html.twig
```diff
@@ -23,5 +26,60 @@
...
+      dbz: ">65",
+      description: "Extremely heavy precipitation including water-coated hail",
+      gradient: "180deg, #F174FD 0%, #F875FF 30.26%, #AA0BFA 34%, #5B06D3 98.5%"
+    }]
+    %}
+
+    <div class="padding-x-2 padding-bottom-1 margin-top-4 wx-radar-legend">
+      <h4 class="visual-h3 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3 margin-bottom-2">{{ "Radar legend" | t }}</h4>
+
+      <table class="usa-table usa-table--borderless margin-top-0 wx-radar-legend" tabindex="0">
```
partly-igor, line 60: I was wrong in my original comment, we should leave that radar description as is, but perhaps include a SR only caption here with the table description:
"Legend which describes the intensity levels that correspond to color and dBz values"

web/themes/new_weather_theme/assets/sass/components/_radar.scss
mgwalker: Oops! Thanks, good catch. I'd meant to go back and replace as much as I could with USWDS functions, but I got too eager... 😛

web/themes/new_weather_theme/assets/sass/components/_radar.scss
greg-does-weather: 👍 I moved the border color and vertical alignment into a `th, td` selector instead of duplicating them, which is why this comment doesn't appear as outdated.

partly-igor (approved):
Looks great, thanks for making the updates!

Issue #1202, opened 2024-05-13
Update timezones within Drupal


kmranjo:
### Current state 
The default time zone options for any configuration within Drupal reflect a long list of locations, but not the standard time zones that users are used to seeing. 

### Acceptance criteria
Default time zone options within Drupal should reflect the most frequently used options: 
- Coordinated Universal Time (UTC)  (aka Zulu time)
- Alaska (UTC-8)
- Aleutian (UTC-9) 
- Arizona Mountain Standard (UTC-7)
- Central (UTC-5)
- Eastern (UTC-4)
- Hawaii (UTC-10)
- Mountain (UTC-6)
- Pacific (UTC-7)
- Samoa (UTC-11)

Source: [time.gov](https://www.time.gov/)

eric-gade:
@kmranjo Do we want to display the UTC offsets in the dropdown list? I ask because depending on the time of year, these are going to be different. For example, Alaska is currently in Alaska Daylight Time (UTC-8) but later on will be in Alaska Standard Time (UTC-9)

kmranjo:
Dangit, I totally forgot about daylight savings (as I usually do). I think
if we CAN include the UTC offsets then thats helpful to the user, but it's
a nice-to-have and not necessary. I think generally they'll know what time
zone they need to be selecting based on where they are publishing for/from,
so the specific UTC offset is maybe just a reminder when we do hit DST.


*Katrina Ranjo* (she/her)

Content Design, 18F

Technology Transformation Services

PR #1264, opened 2024-05-30
CMS Timezone List

eric-gade:
## What does this PR do? 🛠️
The PR addresses #1202, which calls for using a much shorter canonical list of available timezones when creating or editing a user in the CMS. The new timezones are limited to those covering the US and its territories.

The available timezones are computed from a standard list of [tz database names](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), with one place representative of each timezone. We also only display the current variant (standard vs daylight) that makes sense based on the time of year, if needed.

Screenshot 2024-05-31 at 8 23 53 AM

### DIFF
--- a/web/modules/weather_cms/src/Timezones.php
+++ b/web/modules/weather_cms/src/Timezones.php
@@ -82,7 +82,7 @@ class Timezones
      */
     public static function getTimezoneIdsToLabels()
     {
-        $sorted_timezones = $result = [];
+        $result = [];
         foreach (self::$US_TIMEZONE_IDS as $timezone_id) {
             $timezone = new \DateTimeZone($timezone_id);
             $label = self::getLabelForTimezone($timezone);
--- a/web/modules/weather_cms/src/Timezones.php
+++ b/web/modules/weather_cms/src/Timezones.php
@@ -8,6 +8,7 @@ namespace Drupal\weather_cms\Timezones;
  * to longer names we have take from NIST at
  * time.gov
  */
+global $ABBREVS_TO_LONGNAME;
 $ABBREVS_TO_LONGNAME = [
     "UTC" => "Universal Coordinated Time (UTC)",
     "AKDT" => "Alaska Daylight Time",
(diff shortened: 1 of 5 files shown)


kmranjo (approved):
From chaos to a list I can actually read! Yay!

web/modules/weather_cms/src/Timezones.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Return an associative array that maps
+     * PHP timezone ids to custom labels that
+     * can be used in a user interface
+     */
+    public static function getTimezoneIdsToLabels()
+    {
+        $sorted_timezones = $result = [];
```
greg-does-weather, line 85:
```suggestion
        $result = [];
```
It looks like `$sorted_timezones` isn't used?

web/modules/weather_cms/src/Timezones.php
```diff
@@ -0,0 +1,94 @@
...
+        "Pacific/Honolulu",
+    ];
+
+    /**
+     * For a given DateTimeZone instance, provide a label
+     * string that can be used in a user interface.
+     */
+    public static function getLabelForTimezone(\DateTimeZone $timezone)
+    {
+        $abbrev = (new \DateTime("now", $timezone))->format("T");
```
greg-does-weather, line 59: Ooh, clever.

I wonder if this means on the day before DST ends, all the labels for the future will still reference their respective daylight savings time zone names even though DST will be ending. I don't think there's a general solution to that, though, because whether or not the DST name is correct will depend on when the user is trying to schedule something, not on, like, the natural order of the universe or something. Still, will be interesting to see. Anyway, clever!

web/modules/weather_cms/src/Timezones.php
```diff
@@ -0,0 +1,94 @@
...
+        $longname = self::$ABBREVS_TO_LONGNAME[$abbrev];
+        $utc = new \DateTime("now", new \DateTimeZone("UTC"));
+        if ($abbrev == "UTC") {
+            return $longname;
+        }
+        $offset = "";
+        $offsetInt = $timezone->getOffset($utc) / 60 / 60;
+        $prefix = "+";
+        if ($offsetInt < 0) {
+            $prefix = "";
```
greg-does-weather, line 72: Should this be `-` instead of empty string for UTC offsets to the east of the prime meridian?

greg-does-weather (approved):
Two minor questions, but this looks good! I like how you figure out whether it's currently DST. Very clever.

PR #1296, opened 2024-06-06
Parse location text from alert descriptions, if available

### LINKED ISSUE #1115: Address redundancy in "severe" Watch products
I believe the formatted "areas affected" is repeating the unformatted "counties" list above.  Can we eliminate this redundancy?  I understand these types of watches may be formatted differently.  Also, it's not completely clear that something like "Lamar, TX" in the formatted list is a county vs. a city.
image

greg-does-weather:
## What does this PR do? 🛠️

This PR adds logic that attempts to parse out county region and county name information from alert descriptions. If this extra location information is present, we use it in place of the `areaDesc` property.

We observed that this "extra location information" had a consistent format in all of the examples we saw, and it looks like this:

```
IN [statename] THIS [warning/watch/etc] INCLUDES ## COUNTIES

IN [region of state]

county 1    county 2    county 3
county 4    county 5    county  6

IN [other region]

county 7    county8

THIS INCLUDES THE CITIES OF city 1, city 2, city 3
city 4, city 5, city 6
```

This PR adds a utility function that executes a series of regular expressions to identify whether this block exists and then to split it into a list of regions (e.g., "Northern Nebraska"), a list of counties for each region, and a list of cities. It then returns the more-detailed location information along with a modified version of the alert description that has the location information removed.

The alert trait uses the new utility method. If the utility method returns false, then the alert trait continues working as before. Otherwise, it updates the alert's description and sets the location information based on what's returned from the utility.

- closes #1115

image


web/themes/new_weather_theme/templates/block/block--weathergov-local-alerts.html.twig
```diff
@@ -56,12 +56,55 @@
             {{ alert.instruction | nl2br }}
           </p>
           {% endif %}
-          <h4 class="wx-visual-h3 text-normal text-primary-dark">{{ "Areas impacted" |t }}</h4>
-          <ul class="usa-list {%- if alert.areaDesc | length > 7 %} wx-col-2 {%- endif -%}">
```
partly-igor, line 60: For the alerts that do have region information, I think it still makes sense to add a unifying/top heading ("Areas impacted")  or this section of the hazard details, and then each of the "Counties in" would be the next heading level down (i.e. `h5.wx-visual-h4`). That way the section as a whole is described and grouped.

eric-gade (approved):
🥳 Woohoo! This is top-notch parsing!
  
is-that-code

  
I took my time with this one and double-read everything. I couldn't see any glaring issues. I was only able to view the example bundle (couldn't find any alerts that fit the description at the time I wrote this). Everything looks good!

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation
- [x] any new modules added documented in modules.m

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #1297, opened 2024-06-06
Add a dark color background to hero as fallback

partly-igor:
## What does this PR do? 🛠️
This PR adds an additional color background to the hero image so that the search box is visible before the image loads and as a fallback if it doesn't. 

Normal loading

No image


eric-gade (approved):
Good thinking! 👍

PR #1310, opened 2024-06-10
Add current satellite loop

### LINKED ISSUE #1240: Prototype: Satellite
- Using GOES, WFO WCA images https://www.star.nesdis.noaa.gov/GOES/wfo_index.php
- Place under radar on `Current` tab

greg-does-weather:
## What does this PR do? 🛠️

- adds WFO grid information to the page-level metadata (available as `weather.grid` in Twig templates)
- fetches WFO satellite metadata from [NESDIS](https://cdn.star.nesdis.noaa.gov/WFO/catalogs/WFO_02_mpx_catalog.json) (the WFO is part of the URL)
- builds correct GIF URL based on satellite metadata
- adds the most recent satellite loop GIF for the location's WFO as an image in the today tab

Deployed in my environment: https://weathergov-greg.app.cloud.gov/point/36.166/-86.779?#current

- closes #1240 

image


web/themes/new_weather_theme/templates/layout/page--point.html.twig
```diff
@@ -56,6 +56,30 @@
         <div class="wx-tab-container wx-focus-offset-205" id="current" role="tabpanel" aria-labelledby="current-conditions-tab-button" tabindex="0">
           {{ drupal_block("weathergov_current_conditions") }}
           {% include '@new_weather_theme/partials/radar.html.twig' with { 'point': weather.point  } %}
+
+
+          {% if weather.satellite %}
+          <div class="grid-container margin-top-2 padding-x-0 tablet:padding-x-2" wx-outer-radar-container>
+            <div class="grid-row tablet-lg:grid-col-8 tablet-lg:grid-offset-2">
+              <h3 class="visual-h2 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3 margin-bottom-2">{{ "Satellite" | t }}</h3>
```
partly-igor, line 64:
```suggestion
              <h3 class="wx-visual-h2 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3 margin-bottom-2">{{ "Satellite" | t }}</h3>
```

partly-igor (changes requested):
It's looking good, a few design observations:

1. The "Satellite" heading isn't showing up a "visually h2" I think because it's missing the `wx` prefix
2. The satellite gif should take up 100% width of the container. Currently it varies depending on the browser width 
3. The most complicated issue (perhaps meriting its own PR) is that we need to allow users to stop the gif for general usability but also for accessibility standards.  I've put together documentation in [figma](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=10139-201677&t=8HrpYwJcoEAoOJGI-1) for how we could do this (along with the notes above). I've put the icons for the new buttons ([play](https://drive.google.com/file/d/1tJdKOqvK-967f1pLwdChF4DmDuQSLGyn/view?usp=drive_link), [pause](https://drive.google.com/file/d/1QcJwxVE-HiYCzjUCOhXCVlBpHDypn6Q9/view?usp=drive_link)) on the drive

greg-does-weather:
I think you're right that stopping animation is a separate PR because there are a few things to consider with it. I'll write out my thinking here so I don't have to try to just keep it in my head.

There's no way to pause a GIF in the browser, and there are basically three options to support pause-like functionality:
1. When the image should be paused, replace it with a static image. When it should be playing, replace it with the GIF. This is an easy approach, but if the user "pauses" the GIF in the middle of the loop, instead of staying at and resuming from the middle, they'll instead see a flash as the image changes to the static image, and when they "unpause" the GIF, it will start playing from the beginning, not where they stopped it.
2. Replace the GIF with a series of images and animate them ourselves. This is harder and I'd advocate against it, but it does allow us to preserve the right tracking location within the animation. I think we already have access to the individual image frames via the metadata JSON, so at least we don't have to figure that out. Another argument against this approach is it requires the user's device to download each of the images individually, which could result in a lot more data going across the wire than downloading a single GIF.
3. We could talk to NESDIS about whether they could add a WebM (or similarly patent unencumbered encoding) video file in addition to the GIF. Oftentimes video files can be smaller than GIFs, and because they are videos, we can take advantage of browser-native video controls, including pausing and resuming. For example, the GOES16 GIF for MPX right now is about 7.7 MB. Converting it to WebM reduces that to 1.3 MB. (I don't want to make too many assumptions about how NESDIS is creating images, but ffmpeg can convert GIFs to WebM.)

greg-does-weather:
For the width of the GIF, the source image is 600 pixels wide. Should we stretch the image to be 100% of the container? (Easy fix, just wanted to confirm!)

partly-igor:
Yeah that makes sense – I do think we should pursue one of these options, and it sounds like option 3 would be the best path if we are able to get the video format. I am curious if converting the gif to WebM on our server is a possibility? 

For the width – yeah, with a GIF I tend to be more forgiving about going a little bigger than the source size, and I think it'll look better than the variable border. I do want to make sure the aspect ratio stays the same (i.e stretch proportionally) though.

greg-does-weather:
Yeah, fully agreed. If we can get video, that's the best case. I'll open a ticket to dive deeper into the pause/play issue.

For now, I've updated this PR to make the satellite's minimum width 100%.

web/themes/new_weather_theme/templates/partials/satellite.html.twig
```diff
@@ -0,0 +1,22 @@
...
+      <p>
+        {{ "Listed below is an animated loop of the most recent
+            satellite imagery for @place."
+            | t({"@place": weather.place })
+        }}
+      </p>
+    </div>
+
+    <div aria-hidden="true">
+      <img src="{{ weather.satellite.gif }}" class="wx-minw-full">
```
partly-igor, line 18: From  some light searching, it seems like even though it's aria-hidden we should have a blank `alt` attribute on this. 

I am wondering if there is a reason not to use alt text here instead of the sr-only div/paragraph above?

partly-igor (approved):
Changes look good, one small question about the alt text, but otherwise 👍 

Should we open a ticket for the animation controls?

PR #1313, opened 2024-06-11
Reorganize SASS files and classes

partly-igor:
## What does this PR do? 🛠️
This PR reorganizes some of our sass files, putting component specific classes into their own sass file partial. This is 1 of 2 PRs for #1247. The second PR will add the `wx` prefix to our custom utilities and components. 

## What does the reviewer need to know? 🤔
There should be no visual difference before and after this PR

PR
Screenshot of current tab for the PR

Screenshot of 7 days tab for the PR

Beta
Screenshot of current tab on the beta site

Screenshot of 7 days tab on the beta site

PR #1315, opened 2024-06-12
Add wx prefix to custom classes 

### LINKED ISSUE #1247: Organize CSS files

partly-igor:
## What does this PR do? 🛠️
This PR adds a `wx` prefix to the custom utility classes and components we use.  It add to the work started in #1313 and 
together with that PR closes #1247 

Two notes:
1. This PR is on top of #1313 and should be reviewed/merged second (or alternatively just review this one, though there are more files changed so the diffs for the prefix addition might be longer) 
1. Seeing in in practice here, it's a good moment to re-evaluate if we do want to go ahead with this approach of using the prefix for all custom classes.


web/modules/weather_blocks/src/Plugin/Block/Test/DailyForecastBlock.php.test
```diff
@@ -24,7 +24,7 @@ final class DailyForecastBlockTest extends Base
      * Test that the block returns the expected data if we're on a location route.
      * @group unit
      * @group block
-     * @group daily-forecast-block
+     * @group wx-daily-forecast-block
```
greg-does-weather, line 27:
```suggestion
     * @group daily-forecast-block
```
These test groups refer to the Drupal block names, not CSS classes. They're only used to target tests to particular components instead of having to run the whole test suite.

web/modules/weather_blocks/src/Plugin/Block/Test/DailyForecastBlock.php.test
```diff
@@ -45,7 +45,7 @@ final class DailyForecastBlockTest extends Base
      * Test that the block returns null if we're not on a grid route.
      * @group unit
      * @group block
-     * @group daily-forecast-block
+     * @group wx-daily-forecast-block
```
greg-does-weather, line 48:
```suggestion
     * @group daily-forecast-block
```

web/modules/weather_blocks/src/Plugin/Block/Test/HourlyForecastBlock.php.test
```diff
@@ -25,7 +25,7 @@ final class HourlyForecastBlockTest extends Base
      * Test that the block returns the expected data if we're on a location route.
      * @group unit
      * @group block
-     * @group hourly-forecast-block
+     * @group wx-hourly-forecast-block
```
greg-does-weather, line 28:
```suggestion
     * @group hourly-forecast-block
```

web/modules/weather_blocks/src/Plugin/Block/Test/HourlyForecastBlock.php.test
```diff
@@ -55,7 +55,7 @@ final class HourlyForecastBlockTest extends Base
      * Test that the block returns null if we're not on a grid route.
      * @group unit
      * @group block
-     * @group hourly-forecast-block
+     * @group wx-hourly-forecast-block
```
greg-does-weather, line 58:
```suggestion
     * @group hourly-forecast-block
```

greg-does-weather (approved):
This looks good to me! I made some suggested edits related to PHP test names, but it's not breaking or blocking either way.

partly-igor:
Thanks for catching those

PR #1317, opened 2024-06-13
Let's Control Time

eric-gade:
## What does this PR do? 🛠️
This PR implements a way for us to control time (but not space -- being lords of all creation will have to wait for a subsequent PR).
  
The gist is that setting the `WX_NOW_TIMESTAMP` to a valid ISO timestamp will set the "now" time to be that value in both the drupal and api-proxy containers. The upshot being that, in certain cases, we can control what "now" means in development environments.
  
Additionally, we provide in-code / rest methods for updating the now value, should such a need arise.

### Drupal Side ### 
The Drupal dev settings file will now check for an env variable `WX_NOW_TIMESTAMP` and, if set, will set the `$settings['wx_now_timestamp']` key to that value.
  
The `DateTimeUtility` has a static method for `now()` that will check the Drupal settings. If there is a timestamp there, it will create a DateTime object according to that timestamp. If not, it will check to see if the setter has been used to programmatically set a system-wide `now` timestamp, and use that. Otherwise, it uses the actual current time as "now".
  
**One issue to note:** Every time you run `make cc` or anything that reloads the Drupal app, you will need to re-pass the environment variable. This is because the settings file is re-read in these cases, and if it doesn't see the environment variable, it will simply use the current time as the "now" value.
  
### API Proxy Side ###
The API proxy application has similarly been updated, so the `config` object has getters and setters for its `now` property. The getter will:
1. Check the env variable `WX_NOW_TIMESTAMP` and, if present, parse a `dayjs`object from it, otherwise
2. Check to see if an internal variable has been set programmatically via the api itself (see below), and if present use that value, otherwise
3. Use the current time as now
  
You can progammatically set the time by sending a get request like so:
```
get("http://localhost:8081/set-now?t=<iso-timestamp-url-encoded>");
```
  
Additionally, the method and current "now" time will be displayed in the proxy's main index UI.


greg-does-weather (approved):
Fabuloso

Issue #1153, opened 2024-05-05
Rectify high temperature hourly vs. daily differences 


coreypieper:
Forgive me if this is on an existing ticket. Another example today. 
![IMG_4571](https://github.com/weather-gov/weather.gov/assets/4570239/515a7d00-72dd-4dcd-85a3-5949231c3a2f)

eric-gade:
One solution we had discussed in the recent past was to simply use all the hourly temperature data to figure out the high and low values for the corresponding time period. I don't remember offhand why we decided against this. But it is worth knowing that the daily forecasts day/night temps don't always correspond to the highest/lowest forecast temps in the hourly data.

greg-does-weather:
Closed by #1319

PR #1319, opened 2024-06-13
Fix high/low temp for 'today' period

eric-gade:
## What does this PR do? 🛠️
With #1153 we noticed that the 7-day's "today" forecast card, when it's after 6pm, still displayed both the high and low temperatures. We decided that after 6pm and through the "overnight" period, we should only show the low temperature.
  
This PR implements the change, along with a test.

It can be hard to visually test this, since we don't have full control of time ([ahem, ahem](https://github.com/weather-gov/weather.gov/pull/1317)). In the screenshot below, I was able to confirm this is working by checking the weather in Guam at 9:30AM New York time, which is 11:30PM Guam time, and therefore meets the criteria for only showing the low temp.

Shouts to @greg-does-weather for highlighting a simpler way to do this than I originally came up with

Screenshot 2024-06-13 at 9 53 21 AM


web/modules/weather_blocks/src/Plugin/Block/Test/EndToEnd/DailyForecast/DailyForecast.php.test
```diff
@@ -260,4 +260,24 @@ final class DailyForecastStructureTest extends EndToEndBase
...
+     * @group eric
+     */
+    public function testOnlyLowAfter6pmToday()
+    {
+        $now = \DateTimeImmutable::createFromFormat(
+            \DateTimeInterface::ISO8601_EXPANDED,
+            "2024-05-07T18:01:00-0500",
+        );
+
+        $this->onLocationRoute(32.778, -96.796);
```
greg-does-weather, line 277: I had to think about this for a minute, but I get it now. Based on how the `useOnlyLowForToday` flag gets sets below, I was thinking the only thing that mattered was how the data was returned from the API, but then I remembered that we subdivide the daily forecast periods into days based on the current time and the times of the periods.

https://github.com/weather-gov/weather.gov/pull/1319/files#diff-3cf5ed10dc8559478c4d8c2552ad2691ced84edeb99636aa49fdeebdacec014cR262

greg-does-weather (approved):
This is such a hacky problem. Nice solution!

PR #1342, opened 2024-06-20
Fix a bug with transforming GeoJSON to WKT

### LINKED ISSUE #1331: SpatialUtility doesn't handle Multipolygon geometries
### Description
Currently, our `SpatialUtility` helpers are able to transform API geometry data into spatial SQL for geometries whose `type` is `Polygon`. This SQL generation will fail, however, if the type is `Multipolygon`.
  
The `Polygon` type consists of coordinate pairs in an array structure like so:
```json
"type": "Polygon",
"coordinates": [
    [
            [-96.398, 28.344],
            [-96.414, 28.327],
            [-96.545, 28.264000000000003],
            [-96.7159999, 28.156000000000002],
            [-96.852, 28.066000000000003]
      ]
  ]
```
  
However, the `Multipolygon` type consists of an extra nested array representing sets of coordinate arrays (in other words, it's one dimension deeper):
```json
{
  "type": "MultiPolygon",
  "coordinates": [
    [
      [
        [-96.8, 28.22],
        [-96.81, 28.22],
        [-96.801, 28.264],
        [-96.798, 28.273],
      ],
      [
        [
          [-96.324, 28.642],
          [-96.362, 28.626],
          [-96.366, 28.628],
          [-96.356, 28.635],
        ]
      ],
    ]
  ]
}
```

### Example bundle ###
The attached bundle will throw a SQL error that is caught by the `DailyForecastBlock` and displayed to the user as a failure to load. The actual error occurs in `SpatialUtility` during the creation of the spatial SQL string, because any attempt to create such a string from a `Multipolygon` geometry type will respond with `POLYGON(1, 1)` because of the assumption of array dimensions.
  
Bundle: 
[bundle_6670bf653ab46.zip](https://github.com/user-attachments/files/15886928/bundle_6670bf653ab46.zip)

### Acceptance criteria
- [ ] `SpatialUtility` can generate correct spatial SQL for Multipolygon geometry types
- [ ] The daily forecast loads even if there are errors associated with alerts
- [ ] If there is an error getting alerts, we show the alerts tab with an error message on it

greg-does-weather:
## What does this PR do? 🛠️

Due to a bad assumption about what the API returns (I assumed it was always polygons or nothing), the process by which we converted an alert geometry into a SQL query would break on multipolygons. This PR implements a new helper method called `geoJSONtoSQL`, which takes the entire GeoJSON object (instead of just the list of points) and uses the GeoJSON geometry type as a switch for what kind of WKT text to create.

- fixes #1331

### DIFF
--- a/web/modules/weather_data/src/Service/AlertTrait.php
+++ b/web/modules/weather_data/src/Service/AlertTrait.php
@@ -81,13 +81,11 @@ trait AlertTrait
             // If there's a geometry for this alert, use that to determine
             // whether it's relevant for our location.
             if ($alert->geometry) {
-                $alertWKT = SpatialUtility::geometryArrayToWKT(
-                    $alert->geometry->coordinates[0],
-                );
+                $alertSQL = SpatialUtility::geoJSONtoSQL($alert->geometry);
 
                 $sql = "SELECT ST_INTERSECTS(
                     $gridWKT,
-                    $alertWKT
+                    $alertSQL
                 ) as yes";
 
                 $intersects = $this->dataLayer->databaseFetch($sql)->yes;
--- a/web/modules/weather_data/src/Service/SpatialUtility.php
+++ b/web/modules/weather_data/src/Service/SpatialUtility.php
@@ -39,11 +39,51 @@ class SpatialUtility
         return "ST_GEOMFROMTEXT('POLYGON(($wkt))')";
     }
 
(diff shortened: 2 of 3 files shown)


loganmcdonald-noaa (approved):
I read the code and to me trigonometry-challenged mind it makes sense :) nice test.

PR #1348, opened 2024-06-21
Updating relevant library versions for v1.0.0

eric-gade:
## What does this PR do? 🛠️
We failed to update some javascript library versions before the v.1.0.0 release. This means that browsers that were caching the JS files were loading the page incorrectly, due to new element class names etc.
  
I believe this PR fixes the issue by updating the versions on all relevant / changed js libraries


greg-does-weather (approved):
Let's give it a shot

PR #1364, opened 2024-06-25
Updates all our package-lock.jsons since we are getting out of sync errors

loganmcdonald-noaa:


greg-does-weather (approved):
I dunno how the lockfiles would have gotten out of sync, but if this fixes it, then okie dokie.

PR #1393, opened 2024-06-26
Add a button to expand and contract the radar

### LINKED ISSUE #1309: Add expand button to radar map
### Description
Through Touchpoints and individual conversations we've heard that users can get stuck scrolling the radar when they want to scroll the page on mobile devices. This is because the radar and the site both allow scrolling, and the map can takes up the entire screen at certain points. To resolve the issue we've decided to:
1. Change the starting height (and thereby the aspect ratio) 
2. Add a button to expand/shrink the radar map 

[Figma Prototype](https://www.figma.com/proto/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?page-id=8510%3A184114&node-id=8530-192779&viewport=882%2C468%2C0.2&t=ZshdVhXSZjHA3GZG-1&scaling=scale-down&starting-point-node-id=8530%3A192779)

[Documentation](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=8735-44604&t=W3iiiU3sNhTv1h0E-1)

Note that there is currently no USWDS icon for map zoom-in, only zoom-out. I've created the [icon](https://drive.google.com/file/d/1hz-pATlz9JTBiYzTcZseALS4KlkPpNSK/view?usp=drive_link) we should use in the expanded state (the collapse/zoom-in icon). 
### Acceptance Criteria
- [ ] On mobile the map takes up the full container width, and starts with an aspect ratio of 6/4 (width/height)
- [ ] On mobile the map  takes up the full container width and expands to an aspect ratio of 3/4 
- [ ] On tablet+ the starting ratio is 3/2. 
- [ ] On tablet+ the expanded ratio is 1/1
- [ ] The tablet+ width of map stays as is now – taking up the full container on tablet, and limited to 8 columns on tablet-lg. 
- [ ] The expand/collapse button follows the styling for the different states outline in the documentation 
- [ ] The expand/collapse button is sticky

greg-does-weather:
## What does this PR do? 🛠️

I had fun playing with CSS, except for the surprise discovery that inline elements (and inline-block elements as well) have a default vertical alignment of `baseline` which causes them to have an un-stylable small bottom margin. Still, good to know, I reckon. Otherwise, I don't think there's anything remarkable here, code-wise.

Curious if we might later want to add a transition to the aspect ratio change, but I'm not sure how nicely that would play with the window resize event. Maybe something to experiment with, if we're interested.

- closes #1309

https://github.com/weather-gov/weather.gov/assets/142943695/30c4cb79-cdff-4582-aa1e-205ddd2271b8


web/themes/new_weather_theme/assets/sass/components/radar.scss
```diff
@@ -1,12 +1,57 @@
 @use "uswds-core" as *;
 
 .wx-radar-container {
-  aspect-ratio: 3/4;
+  aspect-ratio: 3/2;
```
greg-does-weather, line 4: The design calls for a starting aspect ratio of `6/4` on mobile and `3/2` on tablet+, but AFAIK those are the same number. 😆

web/themes/new_weather_theme/assets/sass/components/radar.scss
```diff
@@ -1,12 +1,57 @@
...
+
+  &.wx-radar-container__expanded {
+    aspect-ratio: 3/4;
+  }
+
+  button.wx-radar-expand {
+    @include u-bg("transparent");
```
greg-does-weather, line 11: I think this is actually not strictly necessary anymore. I added this when I was trying to get the button to be the same height as its contents, before I learned about the baseline vertical alignment magic margin. But I suppose it doesn't hurt anything, either?

web/themes/new_weather_theme/templates/partials/radar.html.twig
```diff
@@ -21,10 +21,20 @@
...
-    <div class="grid-row wx-radar-container">
-       <wx-radar class="grid-col-12" lat="{{ point.lat }}" lon="{{ point.lon }}">
-         <div id="wx_radar_container" style="height: 300px;"></div>
+    <div class="grid-row wx-radar-container position-relative">
+      <wx-radar class="grid-col-12" lat="{{ point.lat }}" lon="{{ point.lon }}">
+        <div id="wx_radar_container" style="height: 300px;"></div>
       </wx-radar>
+
+      <div class="position-absolute top-0 right-0 height-full">
+        <div class="position-sticky top-0" style="top: 55px;">
```
greg-does-weather, line 30: `top` needs to account for the height of the stickied tab list. It's about 55 pixels, so use that as a default, though the very actually real value is found and set in Javascript.

web/themes/new_weather_theme/assets/js/radar.js
```diff
@@ -1,32 +1,52 @@
...
+    const descriptors = Array.from(
+      container.querySelectorAll(".wx-radar-expand__description"),
+    );
+    for (const descriptor of descriptors) {
+      if (descriptor.classList.contains("display-none")) {
+        descriptor.classList.remove("display-none");
+      } else {
+        descriptor.classList.add("display-none");
+      }
+    }
```
eric-gade, line 31: there is also [`classList.toggle()`](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle) which might do the same thing

web/themes/new_weather_theme/templates/partials/radar.html.twig
```diff
@@ -21,10 +21,22 @@
...
```
eric-gade, line 30: Just playing around with this, I'm wondering if there needs to be the outer `position-absolute` div at all. I'm thinking you could just have the inner one, have _it_ be positioned absolutely, and give it top and right values of like 5px or something. That might accomplish the same thing without all the worry of `sticky` and you won't need the inline style attribute at all -- you could get rid of that code in the js file
```suggestion
        <div class="position-absolute top-2px right-2px">
```

eric-gade: (`top-0` and `right-0` will also work, due to the positioning of the SVG within the containing div there is already some margin)

greg-does-weather:
@partly-igor I think 1, 2, and 3 are fixed. I similarly saw 4 for a while last night and it was related to which element was receiving the click event. I can't reproduce it anymore, but if you see it again, definitely let me know so we can dig into it.

partly-igor (approved):
Design edits looks great!

PR #1426, opened 2024-07-11
Update to latest CMI library

greg-does-weather:
## What does this PR do? 🛠️

- Updates our libraries to point to the latest deployed version of the CMI library
- Updates our local copy of the CMI CSS to the latest
- Updates how we initialize the CMI component to include two new required settings (set to null since we don't use them, but they are required by schema validation)
- Updates selector used to remove the menu from the DOM

If you pull it down locally, radar should work now.


eric-gade (approved):

greg-does-weather:
I suggested we hold briefly on this. CMI will hopefully have a hotfix release in the near term that reverts the `TILES` vs. `TILED` change. As expected, it was a result in an update to the geoserver, but the `TILES`/`TILED` change was a bug. CMI was updated to mitigate the bug, but when the geoserver is fixed, CMI will be updated again.

We can check with Kari to see if there's an ETA on the geoserver fix. Unless we want to get a temporary fix up on beta sooner, in which case we can merge it and cut a release.

cc: @shadkeene @coreypieper

PR #1427, opened 2024-07-11
ADR for basic internationalization (1343)

eric-gade:
## What does this PR do? 🛠️
This PR addresses #1343 by providing an ADR for how we will initially set up internationalization for static / templated text.

## What does the reviewer need to know? 🤔
This is just the ADR. A follow-on PR will be filed that includes the initial implementation for English.

Issue #1347, opened 2024-06-20
Implement lopcalization of static content in English


colinmurphy01:
Dependent on #1343

colinmurphy01:
Likely to linger into 4.3

Might need to break into different types of content:
- POC with the homepage first

greg-does-weather:
Closed by #1428

PR #1428, opened 2024-07-11
Initial i18n setup

eric-gade:
## What does this PR do? 🛠️
This PR implements the recommendations in #1347 and is a follow-on to  the ADR in #1427
  
### Basic settings ###
With the enabling of the Drupal core Language and Interface Translation modules (reflected in the exported config), we now have the capability to localize the site in multiple languages. Those languages need to be added in the "Languages" part of the administration settings. For now, we only have one language enabled (English).
  
(NOTE: We have enabled English-English translations for now, as a way of testing our system. However, if English is set as the default language we can also turn off the need to provide translations for it later. For now it is beneficial to keep on until we officially start working in a second language).
  
The name of the specific setting for translating static assets from templates or other config files is "Interface Translation," and the config updates in this PR should also account for the correct settings there. Note: we do _not_ want to update any individual phrase translations from within the CMS here, even though it offers that capability. We will strictly be using the `.po` files (see below) that are checked into this code repository.
  
### Translation files ###
Drupal uses the [gettext file format](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html) for specifying translations in each language. 
  
The convention is to name each file with the two-letter code for the given language, followed by the gettext file extension of `.po`. So for English, our translation file is named `en.po`.
  
Gettext uses the convention where the English phrase is itself the key to other translations. For example, a portion of a Spanish translation file might look something like:
```gettext
msgid "Hello"
msgstr "Hola"
```
  
Awkwardly, this means that our English translation file would have the same phrases as both the lookup and value. Fortunately, if we give a value of an empty string in these gettext files, the system will default to using the lookup key, which is already the English version of the phrase.
  
The initial translation files were generated first using the [`potx`](https://www.drupal.org/project/potx) tool, followed by removing some extraneous pairs by hand.
  
### New module `weather_i18n` ###
Following the example of the vote-gov team, our translation files as well as some i18n settings will be stored in a new custom module called `weather_i18n`. Translation files will be saved to `web/modules/weather_i18n/translations`. 
  
We are using a custom module in part to have a regular place to store the translation files, but also so that any edge cases we encounter down the road can be addressed by module hooks or similar (ths is what happened with vote-gov).
  
### Updating translations ###
If you edit a `.po` file, you will need to let Drupal know there are new translation updates, and also clear the cache. We have added a Makefile command `make ut` which will handle this for you.
  
## Future work ##
We need to address testing the consistency of these files in #1429 
  
## Try it ##
If you want to test this for yourself, try the following:
1. Pull down this branch, import the config, then clear cache or zap if needed
2. Open the english translation file and add a new English-to-English translation for "Day" (ie, edit the `msgstr` value where the `msgid` is "Day").
3. Run make ut
4. Load any forecast page, then look at the daily forecast cards. Instead of saying "Day" for the daytime segment, it should now have the translated string you edited.


Makefile
```diff
@@ -78,6 +78,12 @@ reset-site-database:
 shell: ## Get a shell inside the Drupal container
 	docker compose exec drupal bash
 
+ut: update-translations ## Update Drupal from local .po translation files
+update-translations:
+	docker compose exec drupal drush locale:clear-status
+	docker compose exec drupal drush locale:update
+	docker compose exec drupal drush cache:rebuild
```
loganmcdonald-noaa, line 85: will we want to add any of these steps to our post deploy script?

eric-gade: Oops! Yeah definitely, maybe that will help with my issues here...

web/modules/weather_i18n/translations/en.po
```diff
@@ -0,0 +1,366 @@
+# English translation of beta.weather.gov
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Vote.gov Utility\\n"
```
greg-does-weather, line 5:
```suggestion
"Project-Id-Version: weather.gov Utility\\n"
```

web/modules/weather_i18n/translations/en.po
```diff
@@ -0,0 +1,366 @@
...
+msgid "Your Local Forecast Office"
+msgstr ""
+
+#: wfo.labels.find-us
+msgid "Find us on"
+msgstr ""
+
+#: theme.text.submitted-by
+msgid "Submitted by @author_name on @date"
+msgstr ""
```
greg-does-weather, line 366:
```suggestion
msgstr ""

#: backend.login.sso-login-path
msgid "SSO Login Path"
msgstr ""

#: backend.login.login-text
msgid "Login Text"
msgstr ""

#: backend.login.sso-cancel-path
msgid "SSO Cancel Path"
msgstr ""

#: backend.login.cancel-text
msgid "Cancel Text
msgstr ""
```

These are using the `t()` function on the backend, for building up the SSO admin form. It's less critical that they be translated since it's in an admin form, but it'd still be good to do I guess?

greg-does-weather (changes requested):
Requesting one change, not covered in the comments below.

From the way the `populate-database` action was changed, I'm assuming that translations are loaded into the database from the PO files. If that's correct, then the `action.yml` files for `populate-database` and `setup-site` both need to be updated. They should add the PO files as inputs to the hash function used to create the cache key.

Eg.,

```yaml
- name: database dump cache
  uses: actions/cache@v4
  id: db-cache
  with:
    key: drupal-database-${{ hashFiles('web/config/**/*.yml', 'web/scs-export/**/*') }}
    path: weathergov.sql
```

would become

```yaml
```yaml
- name: database dump cache
  uses: actions/cache@v4
  id: db-cache
  with:
    key: drupal-database-${{ hashFiles('web/config/**/*.yml', 'web/scs-export/**/*', 'web/modules/weather_i18n/translations/*.po') }}
    path: weathergov.sql
```

web/modules/weather_i18n/translations/en.po
eric-gade: 😅

web/modules/weather_i18n/translations/en.po
eric-gade: How did you find these in particular? When I used the `potx` tool, I found thousands of instances of translated strings on the backend/cms, and specifically excluded them from this file because it's just too much, and we don't necessarily care about translating the CMS interface at this point.
  
That said, I'm not especially against including these few in the file for now

eric-gade:
> Requesting one change, not covered in the comments below.
Whoa, good catch!

web/modules/weather_i18n/translations/en.po
eric-gade: There is also this method ``{{ 'Some string' | t }}`` and in the core CMS templates (which we dont directly see unless we go looking) the `trans` filter is also used prodigiously. All of this makes it difficult to actually find what is being translated using normal string parsing -- I think using Twig modules themselves would make it easier. But the `potx` module _does_ use basic string parsing to get these all out

greg-does-weather (approved):

Issue #1355, opened 2024-06-24
Add Hazard on Map Legend and Update Colors


partly-igor:
### Description
As part of #1135 we  designed a simple legend for the hazard on a map that appears in our alert details and it is ready to be implemented. 

As part of this implementation we should update the colors of the hazard to better correspond with our updated alerts (#1322)

[Documented design](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?m=dev&node-id=8917-22283&t=OzxstgbVsSU9B3En-1) 

### Acceptance Criteria
- [x] Legend appears below the map
- [x] The legend color match the updated colors for warnings, watches, and advisories
- [x] The outlined area on the map matches the updated colors for warnings, watches, and advisories

PR #1430, opened 2024-07-11
add impacted area square for hazard map legend

loganmcdonald-noaa:
## What does this PR do? 🛠️
Finishes #1355 by adding the small legend square. 

Screenshot 2024-07-11 at 1 20 03 PM


eric-gade (approved):
LGTM 👍

partly-igor:
Colors and legend are looking great, the only issue I am noticing is that on mobile the map has gotten a little squished (see screenshots below). I am guessing that the legend is counting in the aspect-ratio calculation? We want the map itself to be 3:2 (width:height), in the PR it's closer to 2.2:1 

**PR**
Screenshot of hazard map on design

**beta**
Screenshot of hazard map on beta

web/themes/new_weather_theme/templates/block/block--weathergov-local-alerts.html.twig
```diff
@@ -105,20 +105,14 @@
                 <li class="">{{ areaName }}</li>
               {% endfor %}
             </ul>
-          {% endif %}
+            {% endif %}
 
-          <div class="grid-container margin-top-2 padding-x-0 tablet:padding-x-2">
-            <div class="grid-container padding-0 wx-isolation-isolate grid-col-12">
```
partly-igor, line 111: We'll want to move the `wx-isolation-isolate` class from here and put it on the `wx-alert-map` element

partly-igor (commented):
It's looking really good, the only thing I noticed  with the last update is that the map control floating/z-index issue was reverted because the isolation class was removed. After that I think 👍

PR #1431, opened 2024-07-12
Hourly table alert header a11y fix

partly-igor:
## What does this PR do? 🛠️
This PR updates the alert table heading in the hourly table to always display for screen readers. Currently it was only displaying for the first alert row, resulting in empty table headers which is flagged as an a11y error. 

There is no visual change to the table


eric-gade (approved):

eric-gade:
Man, I can't get the page load times test to cooperate. Could it really be that adding translations pushes us over the limit??

PR #1444, opened 2024-07-18
Fix CI databse cache / building issue

eric-gade:
## What does this PR do? 🛠️
I think we just forgot to update how the hash for the database build is formed .


loganmcdonald-noaa (approved):
if it passes, 👍

PR #1458, opened 2024-07-22
Translation consistency checking

### LINKED ISSUE #1429: Create tests for translation file consistency
### Description

We need to create CI and local tests that ensure the consistency of our translation files. In particular, we need to make sure that all strings that are meant to be translated in our custom templates have entries in the appropriate translation files, and that the translation files themselves do not contain entries that we no longer need.
  
One possible method might involve using the `potx` library and drush command, though I was not able to get that to work properly when working through #1428 

### Acceptance criteria
We have tests that:
- [ ] Ensure all calls to `t()` functions/filters have corresponding entries in all of our `.po` translation files
- [ ] Entries in all of our `.po` translation files have corresponding matches in our templates (ie, there are no "unused" entries)
- [ ] There is a CI action that runs these tests

eric-gade:
## What does this PR do? 🛠️
This PR tries hard to implement the requirements of https://github.com/weather-gov/weather.gov/issues/1429, specifically:
- Ensuring that we have translation terms in the `.po` files for every `t()` variant called throughout our templates/code
- Reporting on any potentially "stale" translations in the `.po` files that are _unused_ anywhere in code

closes #1429

You can run the consistency check with `make check-translations` or `make ct` for short.
  
The check will fail with an error status if there are any `t()` or ` | t` calls in the code/templates, along with reporting on which files/lines these calls appear.
  
It will also spit out a warning -- but not exit with failure -- if it believes there are "stale" translations in the translation file. I have decided to make this a warning rather than a requirement because we might want to keep certain translations around for as yet unspecified reasons. In particular, we have some in there now for the CMS that would not be found by the current setup.
### Implementation ###
The check is a custom script that can be found inside of `tests/translations`.
  
A series of regex checks across templates and php files is conducted, and then checked against the values found in the `.po` translation files for each language.
  
You can specify what template/code files to include or exclude in the `tests/translations/config.js` file.
  
## Fair warning ##
At this point I'm requesting a pull of this code because it could be useful when updating translations. Without having done true translations, it's unclear whether these checks will be "correct" in most cases. Therefore I have decided **not** to add these checks to the CI for now.
  
Also: there's lots of new code here. After banging my head against this for several days, it's still unclear to me whether this is all worth it, and I would not be offended if we scrapped it altogether for now.


greg-does-weather (approved):
LGTM!

PR #1463, opened 2024-07-24
Fixes a bug with how we load forecast and fire zones from shapefile

### LINKED ISSUE #1409: Trouble splitting counties in hazard map/Inland and Coastal Harris County TX example
We received some detailed feedback about a hazards that was showing up as partial county coverage (likely a forecast zone that is just part of a county), and beta.weather.gov hazard map didn't show any of the county as being affected by the hazard.  Harris County is shown here: 
image
Here's what the partial county looked like (look at the Houston area).
image

And here's what beta showed.
image

The person sharing the feedback thought it was odd because beta seems to be handling another "partial county" in that only "Northern Liberty" county was being displayed, and this was correct on beta.

greg-does-weather:
## What does this PR do? 🛠️

Previously, we'd encountered a situation where the zones shapefile could include multiple entries for a single zone. We mistakenly believed this was simply a duplication. However, zones are sometimes described by multiple polygons. We only kept the first polygon we encountered for a given zone, resulting in a couple dozen zones that had wildly incorrect geometries. Worse, it meant we could sometimes tell a user that there were no alerts for them because we had a misrepresentation of zones in our database. And it showed up visually (see #1409) where we drew the zone on the map, but what we drew was not the entire zone.

To fix this, our zones table is modified to expect a geometry collection for its `shape` column. When we import zones from the shapefile, we collect all geometries for a single zone and put them into the geometry collection for that zone's database row.

- closes #1409

After merging, we will need to run a spatial load on all of our environments.

### DIFF
--- a/.github/actions/populate-database/action.yml
+++ b/.github/actions/populate-database/action.yml
@@ -5,7 +5,7 @@ runs:
       uses: actions/cache@v4
       id: db-cache
       with:
-        key: drupal-database-${{ hashFiles('web/config/**/*.yml', 'web/scs-export/**/*', 'web/modules/weather_i18n/translations/*.po') }}
+        key: drupal-database-${{ hashFiles('web/config/**/*.yml', 'web/scs-export/**/*', 'web/modules/weather_i18n/translations/*.po', 'spatial/**/*.js') }}
         path: weathergov.sql
 
     - name: setup image cacheing
--- a/.github/actions/setup-site/action.yml
+++ b/.github/actions/setup-site/action.yml
@@ -11,7 +11,7 @@ runs:
       uses: actions/cache@v4
       id: db-cache
       with:
-        key: drupal-database-${{ hashFiles('web/config/**/*.yml', 'web/scs-export/**/*', 'web/modules/weather_i18n/translations/*.po') }}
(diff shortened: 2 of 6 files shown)


loganmcdonald-noaa:
> After merging, we will need to run a spatial load on all of our environments.

Can we test this in a cloud environment first, or have we already?

loganmcdonald-noaa (approved):
LGTM but let's test this in a cloud environment and document exactly what was run before merging? Let's do this in all envs (design, greg, eric) before merging and then doing in staging. Can we just do it for beta without deploying?

greg-does-weather:
We haven't, but I know someone who can. Hang on a sec!

greg-does-weather:
And yes, we can run these updates outside of a release/deploy.

greg-does-weather:
From this branch (or from `main` after the PR is merged), run this script to load spatial data in cloud environments:

```sh
cf login --sso -a api.fr.cloud.gov
./scripts/load-spatial-data.sh <space-name>
```

Where `space-name` is one of:
- `beta`
- `staging`
- `design`
- `eric`
- `greg`

The script handles targeting the right org/space so no worries there. It also handles the wonkiness associated with `beta` also being called `prod`. 😜 

I have already run it successfully on my environment and can confirm that it does fix #1409. A flood watch as displayed by the Houston WFO:

image

And the same one in my environment:

image

Issue #1469, opened 2024-07-26
[WFO-CMS-Data] Update WFO Promo content type with additional fields


eric-gade:
### Description
This is the first ticket coming out of #1422.
  
The goal is to revamp the WFO Promo content type to include the additional fields as specified by the [Figma file](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=9942-4943&t=vX6pzHVfmYJxkin4-0), as well as the [discussion here](https://github.com/weather-gov/weather.gov/issues/1422#issuecomment-2250633599). When that is complete, editing a (formerly) "WFO Promo" will mean entering all of the relevant data about the WFO as will appear on the WFO page. Pieces of this information (such as the contact info) can also appear on other pages, too.

### Questions and Considerations
* This ticket will involve renaming the WFO Promo content type, which will inevitably lead to some code conflicts that will need to be resolved
* We have already enabled uniqueness (per WFO) validation on this content type. That needs to be preserved, so that we do not end up with multiple nodes for a given WFO
* We will want to ensure that versioning is off for this content type. It should be treated as data entry.
* If possible, we want to prevent a published node of this content type from directly having a link to the node, ie we do not want the plain node view for this content type of be publicly accessible
  
Additionally, whoever works on this issue will want to keep in touch with the design team about the requirements for each of the fields. For example, what types of formatting should be allowed in the "About" field for the WFO, etc.

### Acceptance Criteria
- [ ] WFO Promo content type is renamed to something more descriptive of WFO data generally (something like "WFO Information")
- [ ] The fields specified [here](https://github.com/weather-gov/weather.gov/issues/1422#issuecomment-2250633599) are added to the content type
- [ ] The content type has the uniqueness validation check (cannot make more than one per WFO)
- [ ] Versioning is turned off for the content type
- [ ] Attempt to prevent the plain node view for published instances of this content type from being publicly accessible (we will address how this information is displayed in later tickets)

eric-gade:
I have started work on this ticket. In the course of doing so, I've decided to alter the task slightly:
* Keep the WFO Promo content type as-is for now
* Create a new content type called WFO Information
* Create a subsequent ticket to remove the WFO Promo content type (and all associated nodes) as a later part of the larger work

greg-does-weather:
Closed by #1485

PR #1485, opened 2024-07-29
WFO Information content type

eric-gade:
## What does this PR do? 🛠️
This PR implements #1469, the creation of a new content-type called WFO Information. This content type will eventually supersede the WFO Promo type.
  
The changes are mostly config changes.

### WFO Promo type lives! (for now)
In order to make the transition smoothly, we decided to preserve the existing WFO Promo content type for the moment, and to create a completely new content type called WFO Information. (Removal will be done in https://github.com/weather-gov/weather.gov/issues/1486)
### `Phone number` module
We will now use the [Phone number module](https://www.drupal.org/project/phone_number) to give us an actual typed phone number field across the system, with validation and configuration options. This field will now be used on the WFO Information content type.
### `Address` module
We will also be using the [Address](https://www.drupal.org/project/address) module, which gives us a typed field for physical addresses with configuration options and validation on entry. This field will be used by the WFO Information content type.


greg-does-weather (approved):
Rock and roll

kmranjo (approved):

PR #1493, opened 2024-07-30
Add wind cardinal direction and update icon

### LINKED ISSUE #1445: Add wind cardinal direction and update icon
### Description
As described in #1400 we have heard feedback that users would like to see the cardinal directional for wind speed, in addition to an icon. This is also a good opportunity to update the wind icon to be more arrow-like again. 

[Figma documentation ](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=9848-407691&t=AA0iHvSxZoZJTw8n-1)

[Updated wind icon](https://drive.google.com/file/d/1FS8EhclM-4pS--gyfrdrKXFAkhGs-Iol/view?usp=drive_link)

### Acceptance Criteria
- [ ] Text cardinal direction is added wherever we display wind speed (i.e. current conditions, hourly table)
- [ ] Wind direction icon is updated
- [ ] Spacing between the wind speed and wind direction text and icon matches the design

greg-does-weather:
## What does this PR do? 🛠️

This PR updates the wind direction icon and adds cardinal and ordinal (N, E, W, S, NE, NW, SE, SW) to wind direction. It also updates the spacing to match the design.

- closes #1445

From current observations:
image

From an hourly table:
image


greg-does-weather:
@partly-igor Here are some screenshots from my first pass at implementation. The rotation looks okay to my not-very-discriminating eyes. But this is why I'm not a designer. 😂 

With the addition of the two-character cardinal directions, the wind becomes two lines in the current observations in a lot of cases. Should we do something about that?

web/themes/new_weather_theme/templates/partials/wind.html.twig
```diff
@@ -6,13 +6,14 @@
...
     <span>{{wind.speed}} {{ "mph" | t }}</span>
+    <span class="padding-x-05">{{ wind.direction.short }}</span>
     {#
     The +180 here is to account for the fact that wind direction
     is reported as a FROM direction, but our arrow points in the
     TO direction. So we just need to spin it 'round.
     #}
-    <svg role="img" aria-hidden="true" data-wind-direction class="width-2 height-2 margin-left-05 padding-2px" style="transform: rotate({{ wind.direction.angle + 180 }}deg);">
-      <use xlink:href="{{ "/" ~ directory ~ "/assets/images/spritesheet.svg#wind_arrow_s" }}"></use>
+    <svg role="img" aria-hidden="true" data-wind-direction class="width-2 height-2 padding-2px" style="transform: rotate({{ wind.direction.angle + 180 }}deg);">
```
partly-igor, line 15: It looks like putting padding on the actual svg element makes the vector smaller because the padding counts as part of the size, and makes it a little harder to align. So I suggest we warp the svg in a span that we can then add padding and alignment classes to:

```suggestion
      <span class="display-flex flex-align-center padding-bottom-2px">
    <svg role="img" aria-hidden="true" data-wind-direction class="width-2 height-2" style="transform: rotate({{ wind.direction.angle + 180 }}deg);">
```

partly-igor (commented):
Good catch on the line breaks. I think the value `td`s have 2 units of right-padding by default. We could set that to 0 and that should give enough space for the text. 

Suggesting an edit to the icon markup that should get us pretty consistent alignment and spacing, and get the arrow to be the right size:

Screenshot of wind icon in table with suggested changes applied

web/themes/new_weather_theme/templates/partials/observation-table.html.twig
```diff
@@ -5,7 +5,7 @@
             <th scope="row"  class="font-family-mono text-base font-mono-xs text-uppercase z-400">
                 {{ "Wind" | t }}
             </th>
-            <td>
+            <td class="padding-left-0">
```
partly-igor, line 8: Woops, I think you applied `padding-left` instead of `padding-right` in this commit

web/themes/new_weather_theme/assets/sass/components/current-conditions.scss
```diff
@@ -9,10 +9,13 @@
       fill: color("primary-dark");
     }
   }
+
+  & table td {
+    padding-left: 0;
```
partly-igor, line 14: Same here, should be `padding-right`

web/themes/new_weather_theme/templates/partials/wind.html.twig
```diff
@@ -6,14 +6,17 @@
...
-    {#
-    The +180 here is to account for the fact that wind direction
-    is reported as a FROM direction, but our arrow points in the
-    TO direction. So we just need to spin it 'round.
-    #}
-    <svg role="img" aria-hidden="true" data-wind-direction class="width-2 height-2 margin-left-05 padding-2px" style="transform: rotate({{ wind.direction.angle + 180 }}deg);">
-      <use xlink:href="{{ "/" ~ directory ~ "/assets/images/spritesheet.svg#wind_arrow_s" }}"></use>
-    </svg>
+    <span class="padding-x-05">{{ wind.direction.short }}</span>
+    <span class="display-flex flex-align-center">
```
partly-igor, line 10: Just missing the `margin-bottom-2px` on the wrapper span

partly-igor (commented):
One last missing class for the arrow alignment, otherwise looks good from design perspective now. 🥳  Thank you!

igorkorenfeld (approved):
Design wise looks great!

partly-igor (approved):
Looks great!

Issue #496, opened 2023-12-04
As a site user, I want page URLs and titles to be meaningful


greg-does-weather:
### Description

As a site user, I want page URLs and titles to be meaningful so I can tell what they are at a glance and so I can share them with other people easily.

### Questions and Considerations

- What information do users need?

### Acceptance Criteria
- [ ]

colinmurphy01:
@sarahmhgov @partly-igor @kmranjo Would love a gut check prioritization for MVP (high/medium/low)?

kmranjo:
I think that depends - is this ticket only for meaningful URLs, or would this also cover some guardrails that we want to put around URLs to begin with? For instance, making all URLs lower case doesn't change their meaning, but is a good best practice. I think some basic guardrails to create usable URLs that are less likely to create problems would be medium to high in prioritization for MVP.

greg-does-weather:
Definitely want those guardrails! I think right now, at least some of our URLs are case-sensitive like this is 1990s GeoCities or something. So help thinking through that would be good, if nothing else. That would probably just be something we could huddle on for like 15-20 minutes and then go implement.

With that little thought added, I'll otherwise defer to y'all on the priority of meaningful URLs. 😄

greg-does-weather:
Our URLs will be case-insensitive when #559 is merged. We also add some checking on the values we receive to make sure they make sense (lat/lon are numeric, WFO is 3 letters, etc.) in there. With that change, the last part of the local page URLs also goes away, so instead of:

```
https://beta.weather.gov/local/OHX/50/50/Minneapolis
```

we will instead have:

```
https://beta.weather.gov/local/OHX/50/50
```

We were originally using the value in the URL to display the place name, but that was never the long-term solution and we've moved on from that now, so we no longer need it in the URL. However, URLs that contain it will still work. We just don't use that information for anything. IMO, we should eventually make those 404, but I was being very conservative in case anyone had bookmarked an existing URL.

Eric also brought up whether it'd be okay to ditch the WFO-based URLs entirely and instead just stop at the lat/lon URLs:

```
https://beta.weather.gov/point/44.971/-93.268
```

That would eliminate one API call from our process, and it could be helpful with place names.

colinmurphy01:
That's an interesting idea. Can you document it in a debt ticket so that we can discuss it in a separate thread.

colinmurphy01:
This should be part of our Share epic which we'll scope in `milestone 2`

[...]

PR #1514, opened 2024-08-01
Support redirecting from URLs with place names

greg-does-weather:
## What does this PR do? 🛠️

Adds a new `/place/{state}/{city}` route. We look the state/city up in our places database and then redirect to the lat/lon URL.

- related to #496


loganmcdonald-noaa (approved):

greg-does-weather:
I think we're holding this for now. Is that right, @colinmurphy01?

colinmurphy01:
Yea, this was an idea. I'd love for @kmranjo to weigh in on the strategy here a little more and if we need a second opinon, we can bring in my supe.

greg-does-weather:
One thing I'll note is that by doing this, we could conceivably add all of the ~22,000 places in our database to our sitemap so search engines can find them. We cannot realistically add all of the grid points to a sitemap.

kmranjo:
Having readable place names in the URL can certainly help with SEO, but I don't think its make or break if there are drawbacks. I'm also not sure how we would do places that aren't as simple as state/city. Things like counties, mountains, etc where lat/long points might work better?

greg-does-weather:
This conversation really intrigues me because I don't see any downside to adding this. Whether it serves SEO value and whether/how we extend it to support other place types, it only seems like a net good. The code to support it is tiny and easy to maintain.

I'm not gonna make a stink about it or anything, to be clear. I'm fine with letting it ripen, I'm just not clear on how it's unripe. 😂

colinmurphy01:
Appreciate the nudge.

In summary:
- it'll improve SEO
- it's easy to maintain and expand if needed
- low-no risk in implementing now

If that's the case, let's deploy it. Any objections?

CC @coreypieper @shadkeene

PR #1516, opened 2024-08-01
Add empty alt text to satellite image

greg-does-weather:
## What does this PR do? 🛠️

#1310 added satellite imagery to the "current" tab. The satellite imagery should be hidden from assistive technology because it has the `aria-hidden` attribute; however, we should also set the `alt` attribute for technologies to fall back on if they don't support `aria-hidden`. This PR does that.

This PR also puts the "current" tab back into our accessibility testing flow. It fell out by accident when the tab was renamed from "today" to "current."


partly-igor (approved):
Excellent!

Issue #1470, opened 2024-07-26
[WFO-CMS-Data] Initial WFO page template and routing


eric-gade:
### Description
**NOTE**: #1469 is a prerequisite to this work
  
This issue stems from #1422, which sought to create tickets for WFO data entry and display.
  
The purpose of this ticket is to spin up the initial scaffolding for a WFO page template and the corresponding Drupal routing. The primary task is to simply display WFO data on a custom template (styling and formatting does not matter yet), and to be able to access that page at a predictable URL.

### Questions and Considerations
* What should the URL be for WFO pages? We can change this later, and it might be good for now to start with `/office/<code>`, where `<code>` is the three-letter WFO code, like `OKX` for New York.
* We will __not__ be including any dynamic parts of the page yet, such as local expertise links or the WFO map

### Acceptance Criteria
- [ ] We have a predictable URL which Drupal can successfully route to a (bare bones) WFO page
- [ ] The WFO page template can load field data from the WFO Information content type that was modified in #1469 
- [ ] Do __not__ include any dynamically generated portions of the page, such as the WFO map or the local expertise links

greg-does-weather:
Closed by #1517

PR #1517, opened 2024-08-01
WFO Info routing and initial template structure

eric-gade:
## What does this PR do? 🛠️
This PR implements #1470. Specifically, it:
* Creates a new route in which WFO information nodes can be accessed via their WFO code
* Creates an initial override template structure for this node type, along with hooks that massage the data into a usable form.

### New Routing ###
We have added a new route and handler class in the `weather_routes` module that lets us access a WFO Information node by its WFO code, ie `/offices/OKX` will route to the single WFO Information node that is linked to the OKX office (note that the code can be upper or lowercase).
  
Any incorrect WFO codes -- or references to WFOs for which a node has not yet been created -- will redirect to the 404 page.
  
### Drupal hook for modifying template data ###
Because we want complete control over how WFO Information nodes, their fields, and the constituent data fields on all these entities are rendered in the template, we have to do some modifications to the node's render array. That happens in [this hook inside the theme file](https://github.com/weather-gov/weather.gov/blob/eg-1470-wfo-info-routing/web/themes/new_weather_theme/new_weather_theme.theme#L34).
  
The only thing to know about this is that if we add/remove/modify any of the fields, we will likely need to update the hook.
  
### Template structure ###
The templates for the wfo-information node are overrides, and located at `new_weather_theme/templates/wfo-info/`. Right now there are only two: the override for the node itself, and an override for the image formatter. We went with a field template override for the image because munging its data in the hook is way more complex than other field types.
  
Aside from the image field, the example template will create a `<dl>` list for the variables and their current values for each of the field types available for WFO Information nodes. In other words, for now, the page will simply display what pieces of information are available.

Screenshot 2024-08-02 at 9 44 44 AM
  
localhost_8080_offices_OKX


kmranjo (approved):

greg-does-weather (approved):
Nice! I like that this builds the node and returns that as the render content. That's way slicker than what I imagined. Also, how long did it take to find the way of customizing the image template? Whew.

I'm skipping the accessibility test on this one since the template is only a placeholder. We can do more thorough review once we build out the real template later.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

PR #1519, opened 2024-08-02
Notes on location

greg-does-weather:
## What does this PR do? 🛠️

Add notes on how we do location and why


sarahmhgov (commented):
thank you greg! this is enough and really helpful

eric-gade (approved):

Issue #1525, opened 2024-08-05
[AFD-Template] Create route and query param for AFDs


eric-gade:
### Description
Part of the work for #1475 
  
Add a route and (optional) querystring param for WFO code, and have that URL display the AFD page.
 
The new AFD page should be an actual Drupal page and template in the manner of the point location page.

### Acceptance Criteria
- [ ] Navigating to `/afd` will show the most recent AFD anywhere in the country
- [ ] Navigating to `/afd?wfo=XXX` will show the most recent AFD for the given WFO
- [ ] There is a custom (Drupal) Page and template for the AFD view

greg-does-weather:
Closed by #1538

PR #1538, opened 2024-08-06
Initial AFD route, template, and bundle

eric-gade:
## What does this PR do? 🛠️
This PR implements #1525, specifically:
* Creates a new route and route handler at `/afd`
* Has optional filtering by `?wfo=` querystring on the route
* Creates a new template and rendering pipeline for AFDs along that route
* Can display either the latest AFD anywhere (`/afd`) or the latest AFD from a specific office (`/afd?wfo=ETC`)

Note that per the ticket, we are keeping the formatting of the AFD body for the moment. Parsing will be in the next chunk of work.

### Bundler changes ###
We've also updated the proxy/bundler, so that it can appropriately read, display, and provide information about Products from the API (including AFDs). There is new sample data that has been lightly edited that is available in this PR.
  
### Testing ###
There are no e2e/browser tests yet, because subsequent tickets will change the structure of the AFD data and page templates dramatically as we move through the parent task.

Screenshot 2024-08-06 at 5 15 38 PM


tests/api/products.js
```diff
@@ -0,0 +1,179 @@
...
+  let size;
+  if (recordType === "all" || recordType === "location") {
+    size = data["@graph"].length;
+  }
+  if (recordType === "all") {
+    return `List of most recent ${productType}s (sample of ${size})`;
+  }
+  if (recordType === "location") {
+    return `List of most recent ${productType}s for office ${wfoCode} (sample of ${size})`;
+  }
```
greg-does-weather, line 85: Suggestion in the spirit of limiting scope pollution:

```suggestion
  if (recordType === "all") {
    const size = data["@graph"].length;
    return `List of most recent ${productType}s (sample of ${size})`;
  }
  if (recordType === "location") {
    const size = data["@graph"].length;
    return `List of most recent ${productType}s for office ${wfoCode} (sample of ${size})`;
  }
}
```

web/modules/weather_routes/weather_routes.routing.yml
```diff
@@ -48,3 +48,12 @@ weather_routes.alert-info:
   requirements:
     # Public endpoint
     _access: 'TRUE'
+
+weather_routes.afd:
+  path: '/afd'
+  defaults:
+    _controller: \Drupal\weather_routes\Controller\AFDController::content
+  options:
+    no_cache: 'TRUE'
```
greg-does-weather, line 57: :shakes_fist:

It'd be dandy if Drupal included query parameters in the cache key, but it appears that it caches the route, not the URL. Alas!

PR #1571, opened 2024-08-19
Make it simpler to build CSS and sprites

### LINKED ISSUE #1494: The process for updating our theme sprite sheet is undocumented and too complicated
### Description

Updating the sprite sheet is currently a multi-step process. We should instead make it simple like our other build/maintenance tasks.

### Acceptance criteria

- [ ] There is a single command that rebuilds our SVG spritesheet

greg-does-weather:
## What does this PR do? 🛠️

Instead of needing to install npm modules for compiling CSS and sprites, this PR moves those operations into containers. The make command for building CSS stays the same, and a new make command is added for sprites.

```sh
make build-css
make build-sprites
```

Prompted by @igorkorenfeld, so I'd especially be interested in hearing if these work for you. We can wait until after it's merged so you don't have to screw up your workflow, though.

closes #1494


partly-igor:
I was trying this out today and got 
```
  => ERROR [uswds 2/9] COPY zscaler-root.pem /etc/ssl/certs/zscaler-root.pem                                                       0.0s
------
 > [uswds 2/9] COPY zscaler-root.pem /etc/ssl/certs/zscaler-root.pem:
------
failed to solve: failed to compute cache key: failed to calculate checksum of ref e4c73592-b441-43f4-87aa-5ed1415fa520::qlxuz7caulsyo4gq4uu0kxhcy: "/zscaler-root.pem": not found
make: *** [build-css] Error 17
```

eric-gade:
I can't even `zap` on this branch because I get the following error:
```
Additional property dockerfile_inline is not allowed
```
  
This is probably because I have an out of date version of docker. I'm reluctant to change it, however, because I want my blow-up-env button to stay in the application...

greg-does-weather:
I think this is good to go now. I don't know how I didn't have the same problem with the USWDS container because it is ***obviously*** wrong. I moved everything out of `dockerfile_inline` as well, so it'll continue working with older engines.

jamestranovich-noaa (approved):
`make zap` works now. I also tried `make build-css` and `make build-sprites` and both worked great. LGTM

Issue #1529, opened 2024-08-05
[AFD-Template] Front-end component for retrieving past versions and navigating WFOs


eric-gade:
### Description
This is part of the work outlined in #1475   
  
It should be done adfter #1528 
  
Building upon the async routes created in #1528, this work should build a custom frontend component that is able to respond to changes in the select menus and asynchronously swap markup for the selected AFD.
  
Additionally, the we need to switch the select elements and transform them into full-fledged combobox components, where:
* Selecting a different WFO from the WFO dropdown will update the AFD selection dropdown options, retrieving information about some set number of most recent AFDs;
* Selecting a specific AFD from the AFD dropdown will fetch and update the markup for the selectd AFD, swapping out what is currently displayed

### Acceptance Criteria
- [ ] Select menus are now comboboxes
- [ ] Selecting a WFO updates the AFDs listed in the AFD combobox
- [ ] Selecting an AFD from the AFD combobox asynchronously fetches the markup for that AFD, then swaps it with what is currently displayed

greg-does-weather:
Closed by #1587

Issue #1528, opened 2024-08-05
[AFD-Template] Custom endpoint for async AFD retrieval


eric-gade:
### Description
This is part of the work outlined in #1475 
  
We need to create a custom ajax endpoint that -- when given the correct WFO and/or id parameters -- will return the markup for a _parsed_ AFD.
  
We can use the work on [this branch as a guide](https://github.com/weather-gov/weather.gov/tree/eg-785-dynamic-loading), since it was done back when we were first looking into the possibility of making more of the pieces of the site async.

### Acceptance Criteria
- [ ] There is an endpoint that takes a WFO and an optional `limit` number and returns the _parsed markup html_ for that number of most recent AFDs
- [ ] There is an endpoint that can return the _parsed markup html_ for a single AFD by id

PR #1587, opened 2024-08-21
AFD HTML-only endpoints and selection implementation

eric-gade:
## What does this PR do? 🛠️
This PR mainly implements #1528, which called for a custom endpoint for HTML-only retrieval of AFD markup (designed to be used by a frontend component).
  
In the course of working on this ticket, some other changes were made -- see the next section.
  
This PR also includes a partial implementation of #1529, though it is "switched off" here

### New routes ###
In order to facilitate the independent AFD viewer page with WFO and version switching, we need to add several routes and Drupal render-array `#theme` configurations. Note that we are now using the practice of starting routes with `/wx/` if they retrieve data that is designed for consumption by frontend components (and not the general public).
  
Here is a table describing the various routes and what they are for:
| route | responds with | description |
|------|----------------|------------|
| `/afd` | full page | Will check all available AFDs across the country and redirect to the URL representing the most recent for any WFO.  If passed a `wfo` querystring param, will redirect to the most recent for the given WFO. If passed an `id` querystring param, it will redirect to the specific page for that AFD and corresponding WFO. |
| `/afd/{wfo}` | full page | Will retrieve the most recent AFD for the given WFO and redirect to the corresponding page. |
| `/afd/{wfo}/{afd_id}` | full page | Will display the AFD viewer page for the with the given AFD presented. The specific version and wfo will also be selected by default in their corresponding select menus |
| `/wx/afd/{id}` | markup only | Responds with _only_ the HTML markup for the the given AFD. Designed to be used by a frontend component |
| `/wx/afd/locations/{wfo}` | markup only | Responds _only_ with HTML markup of `<options>` elements representing the available AFD versions for the given WFO. This is designed for use on the frontend, in order to repopulate the versions select menu when the WFO has changed |
  
### Non-JS functionality ###
This implementation does _not_ make use of any JS or frontend components for dynamically loading AFD information, even though the routes are present and tested. 
  
Instead, it uses basic form submission with querystrings to the `/afd` route described above. We did this in order to formally wire up the logic we would need anyway on the backend. Because of this change, we have added a submit button that is not in the original design. In consultation with @partly-igor , we are considering keeping this button even when we add the JS component, so that the interaction is clear.
  
### Bundling updates ###
We discovered an issue whereby the bundle data for products was incomplete. When getting AFD references for either the whole country (`/products/types/AFD`) or a given WFO (`/products/types/AFD/locations/OKX`), there will be references to product IDs that will eventually expire and no longer be available via the API after some time. To deal with this, we've decided to limit the number of references in each of these files to just 10, and then to add all the AFDs for each reference to the bundle. That way we have all the information in perpetuity, and tests will not call out to the real API looking for data that likely doesn't exist anymore.


web/themes/new_weather_theme/templates/layout/afd-page.html.twig
```diff
@@ -1,31 +1,57 @@
...
+
+            <br />
+            <input type="hidden" name="current-id" value="{{afd.id}}"/>
+            <button type="submit">Update selection</button>
+        </form>
+
         <h1>Area Forecast Discussion (AFD)</h1>
         {% if wfo is not null %}
-            <h2>latest for {{wfo}}</h2>
+            <h2>Issued by {{wfo}} at {{afd.issuanceTime}}</h2>
```
greg-does-weather, line 47:
```suggestion
            <h2>Issued by {{wfo | upper}} at {{afd.issuanceTime}}</h2>
```

web/themes/new_weather_theme/assets/js/components/afd-selector.js
```diff
@@ -0,0 +1,85 @@
...
+    if(!afdContainer){
+      return;
+    }
+    const response = await fetch(`/wx/afd/${id}`);
+    if(response.ok){
+      const markup = await response.text();
+      afdContainer.outerHTML = markup;
+    }
+    this.enableInputs();
+  }
```
greg-does-weather, line 70: This is _so slick_. Love it.

greg-does-weather (approved):
Gonna make a couple tweaks, but this is awesome. Great work!

PR #1615, opened 2024-08-23
Update the onboarding template to include adding the new teammate to the contributors list

greg-does-weather:
## What does this PR do? 🛠️

When we onboard a new teammate, we should add them to the contributors list. We forgot once. That's too much. So let's add this step to our template to make it more likely that we do it properly next time.


jamestranovich-noaa (approved):
:+1: :heart:

PR #1657, opened 2024-08-28
Select default WFO correctly, regardless of capitalization of WFO in path

### LINKED ISSUE #1653: When visiting an AFD page using a lowercase WFO code, the WFO select is not populated correctly
### Description

The WFO select is given a default value based on the WFO code provided in the URL. This should be uppercased in the select component so that it matches regardless of capitalization.

image

### Acceptance criteria

greg-does-weather:
## What does this PR do? 🛠️

Uppercases the WFO before building the list of form `<option>` elements.

- closes #1653


jamestranovich-noaa (approved):
LGTM

PR #1658, opened 2024-08-28
Add full width to observations section for current conditions

partly-igor:
## What does this PR do? 🛠️
This PR adds a utility class to the observations section of current conditions to fix a visual bug that occurs when the condition text is missing from current conditions.  In those cases the data below gets scrunched up and doesn't take up the full width like it should. 

Screenshot of before and after the fix for the observations width


eric-gade (approved):
👍

PR #1661, opened 2024-08-28
fix misspellings in ADR issue template

jamestranovich-noaa:
## What does this PR do? 🛠️
Does what it says on the tin.


greg-does-weather (approved):
Well I suppose we can sneak this one in just this once, but next time we need robust tests. :trollface:

Issue #1659, opened 2024-08-28
JSON:API: block anonymous access


jamestranovich-noaa:
### Description

Because JSON:API respects entity permissions, `anonymous` users can `GET` the API listing. We really have no use for a `GET` listing so let's block it outright. Note: previous attempts to do this via apache failed; it's likely we need to do something custom inside JSON:API -- a hook, perhaps?

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

jamestranovich-noaa:
Fixed. Only `uploader` roles can access the API.

PR #1664, opened 2024-08-29
alter json api routes to require role: uploader

jamestranovich-noaa:
## What does this PR do? 🛠️

Alters the route such that the role `uploader` is required to access resources. Addresses https://github.com/weather-gov/weather.gov/issues/1659 -- we can still prevent `GET` but I'm feeling happier about this approach. 

**Note**: depends on the previous JSON API PR

### DIFF
--- a/docs/dev/json-api.md
+++ b/docs/dev/json-api.md
@@ -19,7 +19,7 @@ We have configured JSON:API to only display `wfo_pdf_upload`s and `file`s. (The 
 
 We have also configured a new user type, `uploader`, which has no permissions except to create new `wfo_pdf_upload`s.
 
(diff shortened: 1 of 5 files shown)


greg-does-weather (approved):
This is a really clean solution. I love it!

One small question, and up to you whether you want to do anything with it. Would it make sense to move this into the `weather_cms` module instead of the `weather_login` module? Code organization is a neverending, unsolvable problem so there's no right answer. We also have `weather_routes` that might make sense? We should probably get together as a team and define some basic boundaries for these modules, so we have a shared understanding of what they’re for and what kinds of things go in them.

PR #1669, opened 2024-08-29
Handle weird extraneous newlines in AFD temp/pops table

greg-does-weather:
## What does this PR do? 🛠️

The policy document says the temp/pops table should follow immediately after the heading element, so that's what we parsed for, but it turns out sometimes there's a newline in between them. This PR adds a test to catch that case and slightly modifies the parser to handle it gracefully.


jamestranovich-noaa (approved):
LGTM

Issue #1549, opened 2024-08-09
Spike: implement "outside" testing proof of concept


jamestranovich-noaa:
### Description

We want to implement an "outside" testing proof of concept so that we can run Playwright against a fully developed containerized instance of weather.gov. This PoC will run locally for now. This is separate from Drupal tests. Spun off from https://github.com/weather-gov/weather.gov/issues/1352

- Suggested steps
  - Load CMS config from beta
    - `drush config:import` (we already do this in various places)
   - ~Load CMS content from beta~
     - `drush content:export --all-content` on beta
     - (This might be a large YAML file, so we may have to figure out how to load it from elsewhere if this is too large to commit) 
     - `drush content:import ...`
  - Load test users
    - `drush user:create newuser --mail=testuser@noaa.com' --password=’testpasswd'`
    - Assign this user a CMS role
  - Run a single Playwright test that e.g., logs in with the above test user

Ideally we would set up a separate drupal-test container in the docker-compose so we avoid mixing our development and test environments, but this can be punted for later if need be.

### Questions and Considerations

- We might need to figure out a solution for the content import if it is too big
- Are we missing any other CMS content or database content we need to import?

### Acceptance Criteria

- [ ] We can run Playwright tests against a full weather.gov container instance locally

jamestranovich-noaa:
This is mostly done, but still working on content import since that is a bit trickier than expected.

PR #1671, opened 2024-08-29
"outside" testing proof of concept

jamestranovich-noaa:
## What does this PR do? 🛠️

Duplicates our environment so we can run separate and outside tests on the testing environment. 

- You will need a fairly recent version of `docker-compose` (>=  `2.24.4`) to support the `!override` parameter. 
- You can examine the resulting configuration by doing:
  `docker compose -f docker-compose.yml config` 
  `docker compose -f docker-compose.test.yml --profile test config`

Addresses https://github.com/weather-gov/weather.gov/issues/1549


greg-does-weather (approved):
This is cool!

PR #1673, opened 2024-08-29
Update styles version to bust caches

greg-does-weather:
## What does this PR do? 🛠️

Just what it says on the tin.


partly-igor (commented):
Perfect!

partly-igor (approved):
Hitting the right button this time

PR #1695, opened 2024-09-04
Modify our Javascript to enable bundling

greg-does-weather:
## What does this PR do? 🛠️

This PR changes our Javascript to use ESM modules. It creates a single entrypoint script for each type of page (front page, AFD, and point) which then imports the necessary scripts. The primary theme's library list is updated to remove all the individual components and instead reference the page entrypoints. The page templates are updated to attach only their page entrypoints.

In development, all of the other scripts are still served independently, and there is no build step in order to see changes. The browser continues to load all of the component scripts individually as they are `import`ed from the entrypoint.

In sandbox deployments, the entrypoints are bundled, replacing the original entrypoint file, and delivered to cloud.gov. The practical result is that each page only needs to load a single Javascript file in order to import all of our custom functionality because the library list points at the entrypoint file which is now a bundle of all of the component files required by that page.

In CI/CD testing, the entrypoints are determined dynamically from the library list and are build into single bundled, minified scripts. Tests are then run against the monolithic Javascript. I opted to make this change for testing since I would like to eventually see this bundling and minification pushed to staging and beta, and it'd be good to test against what we're actually deploying.


web/themes/new_weather_theme/assets/js/charts/hourly-pops.js
```diff
@@ -1,98 +1,86 @@
 /* global Chart ChartDataLabels */
-(() => {
```
jamestranovich-noaa, line 2: Looks like we have some charting stuff that was accidentally included here

docs/dev/javascript-bundling.md
```diff
@@ -0,0 +1,41 @@
...
+We use ESM modules for our custom code. This helps create clean scope boundaries
+so we don't have to worry about polluting the global scope. It also helps us
+manage import order, since esbuild can just figure it out for us.
+
+## Drupal libraries
+
+We define what Javascript to deploy in the Drupal theme's libraries file at
+`web/themes/new_weather_theme/new_weather_theme.libraries.yml`. For custom code,
+there should be one entry per page type. The same code may be bundled in multiple
+page types, but we only want to deliver a single bundled file per page type.
```
jamestranovich-noaa, line 18: Can we add something about the fact that we need the yml key to end in `-page` to actually bundle?

jamestranovich-noaa (changes requested):

web/themes/new_weather_theme/assets/js/charts/hourly-pops.js
mgwalker: Not an accident! These were merged into main a little bit ago, so I needed to tweak them for the bundling to work correctly with them.

docs/dev/javascript-bundling.md
greg-does-weather: Yes, thank you! I was struggling to remember what needed to be documented. Slow start with the Monday brain and a gazillion meetings. 😬

jamestranovich-noaa (approved):
LGTM! excited to see this work land!

eric-gade (approved):
I hear ESM is the future.

LGTM 👍 

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

PR #1699, opened 2024-09-04
[Interop Layer] Test that only land alerts are returned

### LINKED ISSUE #1609: [Interop layer] Test only land alerts are returned
- Tests have been implemented or modified to capture the bug
  - [ ] when the API returns marine or other alerts, they are not returned by the interop layer

greg-does-weather:
## What does this PR do? 🛠️

Closes #1609

### DIFF
--- a/api-interop-layer/mocha.js
+++ b/api-interop-layer/mocha.js
@@ -1,0 +1,19 @@
+import sinon from "sinon";
+import * as mariadb from "mariadb";
+
+export async function mochaGlobalSetup() {
+  sinon.stub(global, "fetch");
+  sinon.stub(mariadb.default, "createConnection");
+}
+
+export async function mochaGlobalTeardown() {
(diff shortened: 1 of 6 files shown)


api-interop-layer/mocha.js
```diff
@@ -0,0 +1,19 @@
...
+  global.fetch.restore();
+  mariadb.default.restore();
+}
+
+export const mochaHooks = {
+  beforeEach() {
+    fetch.resetHistory();
+    fetch.resetBehavior();
+  },
+};
```
greg-does-weather, line 19: This file provides global stubbing for the `fetch` method as well as the Mariadb `createConnection` method. It also handles resetting them after every test. This way we don't have to handle mocking them everywhere.

jamestranovich-noaa (approved):

Issue #1447, opened 2024-07-22
Revisit logging in NewRelic for obs station errors


colinmurphy01:
- Is it still running?
- Is it running for OCONUS locations?
- Are we able to calculate the distance from the person's location to the successful obs station? (if not we may spin up a new ticket)

colinmurphy01:
@eric-gade where are we with this.

Last I heard is we found this was broken and needed to be fixed. We identified a fix but needed support, is that still true? Is this blocked?

eric-gade:
I need to ask a new colleague a question about it once I can get my environment set up to deal with NR locally again. There has been no movement on this issue otherwise

jamestranovich-noaa:
I have a fix at https://github.com/weather-gov/weather.gov/pull/1720 for the NR timestamps. Can't guarantee it fixes everything with NR but it should move us forward.

jamestranovich-noaa:
OK. I know why the NR errors are happening and I have a PR for the interop branch. https://github.com/weather-gov/weather.gov/pull/1767 

I confirm that 

- NR is broken in prod, still, but should be fixed when we deploy the api interop layer (we're removing the relevant PHP code from Drupal)
- We are able to calculate the distance from the person's location to the successful obs station (and log if we can't find any valid nearby observation stations)
- We are **not** able to query OCONUS -- but I believe this is unrelated to NR. This seems to be a bug in the api interop layer. Filing a separate issue.

jamestranovich-noaa:
I _think_ we can close this now because https://github.com/weather-gov/weather.gov/issues/1591 (interop layer) is the relevant epic and that branch now has all the NR fixes and logging. 

Was there anything else we needed to do here?

eric-gade:
I agree, we can close it

PR #1720, opened 2024-09-09
fix NR: use elvis instead of null operator

jamestranovich-noaa:
## What does this PR do? 🛠️

Send timestamps to New Relic properly.

Before: 

```json
[{"metrics":[{"name":"wx.observation","type":"gauge","value":2127.3710043474975,"timestamp":false,"attributes":{"withinGridCell":false,"stationIndex":0,"obsStation":"LBNL1","distance":2127.3710043474975,"usesReferencePoint":true,"applicationName":"local"}}]}]
```

After:

```json
[{"metrics":[{"name":"wx.observation","type":"gauge","value":2127.3710043474975,"timestamp":1725906130,"attributes":{"withinGridCell":false,"stationIndex":0,"obsStation":"LBNL1","distance":2127.3710043474975,"usesReferencePoint":true,"applicationName":"local"}}]}]
```

The Elvis operator is a shorthand for ternary operator and is so called because it looks like Elvis with his signature hairstyle. ([Wikipedia](https://en.wikipedia.org/wiki/Elvis_operator))

Addresses https://github.com/weather-gov/weather.gov/issues/1447

### DIFF
--- a/web/modules/weather_data/src/Service/NewRelicMetrics.php
+++ b/web/modules/weather_data/src/Service/NewRelicMetrics.php
@@ -58,7 +58,7 @@ class NewRelicMetrics
         $type = "gauge",
         $now = false,
     ) {
-        $now = $now ?? time();
+        $now = $now ?: time();
 
         $attributes["applicationName"] = $this->applicationName;
 


greg-does-weather (approved):
I've never heard of the "Elvis operator" before. 😂 Still, looking into how this works, this looks like it's definitely the right way to go.

PR #1743, opened 2024-09-10
[Docker Setup] Fix permissions after each zap

eric-gade:
## What does this PR do? 🛠️
Should fix ownership in the docker container plus permissions on the directories locally. 

This is kind of hacky, but the only way I could get it to work. We just run a couple of commands at the end of `zap`

### DIFF
--- a/web/themes/new_weather_theme/gulpfile.js
+++ b/web/themes/new_weather_theme/gulpfile.js
@@ -1,4 +1,4 @@
-const uswds = require("@uswds/compile");
+import uswds from "@uswds/compile";
 
 uswds.settings.version = 3;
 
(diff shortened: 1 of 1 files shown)


greg-does-weather (approved):
You call it hacky, but I call it... well... I mean, ingenious but also hacky, but only hacky out of necessity.

hacking

jamestranovich-noaa (approved):
Ayup, a bit unfortunate we have to do this, but makes sense

colinmurphy01:
Did this fix the docker issue?!?!?

eric-gade:
@colinmurphy01 This _might_ have fixed one or several of our issues (it's hard to tell without using the system for a while). There are some other process concerns we have talked about regarding the database and switching branches, and I think we will formulate some lower priority tickets for those.

Issue #1593, opened 2024-08-22
[Interop layer] Add NewRelic logging


greg-does-weather:
### Acceptance criteria

- Tests have been implemented or modified to capture the bug
  - [ ] metric sends are successful
  - [ ] number of observation stations attempted is recorded
  - [ ] distance between the requested location and the returned observation station is recorded

PR #1744, opened 2024-09-10
[Interop layer] First iteration of abstraction for logging and monitoring

greg-does-weather:
## What does this PR do? 🛠️

- Adds a super basic logger that just writes to the console, which may be sufficient for cloud.gov. The message format is meant to be machine-parseable even when it's multiple lines, but I don't care much about it, so if it's objectionable, we can change it.
- Removes `console.log` statements elsewhere in the interop layer and replaces them with the new logger calls. Also adds a bunch of verbose logs, just for funsies.
- Sets the default log level to `info` and supports changing it via environment variable.

I imagine #1593 could build on top of this?


jamestranovich-noaa (approved):
LGTM! I imagine when the time comes we can also add calls to our application monitoring tool via `monitoring/index.js` :+1:

PR #1757, opened 2024-09-11
[Interop layer] Parity+ error handling

### LINKED ISSUE #1755: [Interop layer] Add basic error handling, at parity with current handling
### Description

We need to do very basic error handling to get to parity with our current error handling before we switch over.

### Acceptance criteria

- [ ] if there is an error fetching or parsing observations, return an error boolean instead of failing the request
- [ ] if there is an error fetching or parsing the daily forecast, return an error boolean instead of failing the request
  - [ ] **Parity:** also return an error boolean if there is an error fetching or parsing the hourly forecast or alerts; or with putting them into the daily forecast object
  - [ ] **Enhancement:** target error booleans specifically to the pieces that failed, not the entire forecast object
- [ ] if there is an error fetching or parsing alerts, return an error boolean instead of failing the request

greg-does-weather:
## What does this PR do? 🛠️

This PR adds error handling to bring us to slighty-above-parity with current.

- Updates the `fetch` wrapper
  - Never throws. If there is an error, it will return an object with an `error` property set to `true`, along with anything else contained in the response
  - For 4xx errors, does not retry. Request errors are unlikely to resolve themselves, so there's no point retrying.
  - For 5xx errors or fetch exceptions, keep the retry logic

- The alert module gains some error handling and smarts
  - If there is an error fetching alerts, an `error` property is set on module metadata, but otherwise the alerts are not updated. This way, we can continue to serve the alerts we know about in the event that we have communication failures between us and the API temporarily
  - When alerts are successfully fetched, the `error` property is set to `false`.
  - Alert module metadata also includes a `updated` timestamp that is set any time the alerts are actually updated
  - The module metadata is included in information sent back to the client, so we will know whether the alerts have failed to update recently, and when alerts were last updated.
  - The alert module is unique in that it never returns an object of the form `{ error: true }`. It will always return a list of alerts, if any are cached and are applicable to the requested location. The `metadata` property is how we'll determine if alerts are outdated.

- The forecast module gets error handling too! But it tries to be smarter.
  - Only the `/gridpoints/{wfo}/{x},{y}/forecast` endpoint is required to be successful in order to return forecast data.
  - If the `/gridpoints/.../forecast/hourly` endpoint returns, then the hourly forecast for each daily component will be included
  - If the `/gridpoints` endpoint returns, then the hourly forecast for each daily component will be included
  - If both of them return, then we'll smoosh them together. Either way, we'll return as much data as we can get.

- The observation module
  - If we get errors fetching data from all three observation stations, we return `{ error: true }` along with a message indicating that we couldn't find any approved stations.
  - If we get invalid observations from all three stations, we return `{error: true }` along with a message indicating that the observations are invalid.

- Satellite
  - If either the connection fails or the response data is invalid, we return `{ error: true }`

And then this one is kind of a big shift in how we've worked so far: if we cannot get WFO grid coordinates for a lat/lon, we return `{error: true}` on all the top-level properties ***except*** alerts. In this case, rather than filtering alerts based on their intersection with the grid geometry, we intersect the point instead. That way, even if we can't show anything else, we can still show relevant alerts.

- closes #1755


api-interop-layer/data/index.js
```diff
@@ -12,90 +12,115 @@ export const getDataForPoint = async (lat, lon) => {
...
-
-    // The daily forecast endpoint includes elevation information, but that
-    // really belongs with the grid, so let's move that, too.
-    if (forecastData.daily.elevation) {
-      grid.elevation = forecastData.daily.elevation;
-      delete forecastData.daily.elevation;
-    }
+  let satellitePromise = Promise.resolve({ error: true });
+  let forecast = { daily: { error: true } };
+  let observed = { error: true };
```
jamestranovich-noaa, line 17: Bit confused as to why we're setting them to true here by default. Are these values overwritten later on?

api-interop-layer/data/satellite.test.js
```diff
@@ -5,44 +5,46 @@ import getSatelliteData from "./satellite.js";
...
   beforeEach(() => {
+    response.status = 200;
     fetch.resolves(response);
   });
 
   describe("returns a URL if everything goes well", () => {
     it("for GOES-16 imagery", async () => {
       response.json.resolves({ meta: { satellite: "GOES-West" } });
-      const actual = await getSatelliteData({ grid: { wfo: "wfo" } });
+      const actual = await getSatelliteData({ grid: { wfo: "wfo1" } });
```
jamestranovich-noaa, line 20: any reason why we changed to `wfo1` `wfo2` etc?

jamestranovich-noaa (approved):
looks good. only had a coupla questions

api-interop-layer/data/satellite.test.js
greg-does-weather: Yes! When debugging the tests, I set the logging to verbose so I could get a better sense of what was going on, and it was helpful to be able to differentiate which test was causing which set of log messages.

api-interop-layer/data/index.js
greg-does-weather: Yes, they get overwritten if there is grid data. They may get overwritten with the exact same thing if fetching them causes any errors. I thought about using an `Object.assign` kind of thing and chose not to, but I don't recall why.

greg-does-weather:
Added some diagrams and some stuff to the docs

PR #1767, opened 2024-09-12
add NR support for the api interop layer

jamestranovich-noaa:
## What does this PR do? 🛠️

New Relic support for the interop layer.

### Metric 

image

### Logs

image

Note that we omit two fields (see comment below).

Addresses https://github.com/weather-gov/weather.gov/issues/1593


api-interop-layer/data/obs/index.js
```diff
@@ -132,6 +133,17 @@ export default async ({
...
     await db.end();
 
+    sendNewRelicMetric({
+      name: "wx.observation",
+      type: "gauge",
+      value: distance,
+      timestamp: Date.now(),
+      attributes: {
+        stationIndex: stations.indexOf(station),
+        obsStation: station.properties.stationIdentifier,
```
jamestranovich-noaa, line 143: I've omitted two NR attributes: `usesReferencePoint` and `withinGridCell`. 

- `usesReferencePoint` depends on a stashed point in the PHP code, it's not clear if we can do this at the api interop layer
- `withinGridCell` uses `getGeometryFromGrid` and `getPlaceNearPolygon` which hasn't been ported to the api data layer yet (and doesn't seem immediately useful to me)

Open to discussion; we can add these but I want to make sure we need these data points in NR first.

eric-gade: I can provide some additional context about why I thought we needed these in the first place (and how we might be able to do without them).
  
`usesReferencePoint` was added when we had, optionally, the ability to load a forecast page by WFO grid cell _in addition to_ an axctual lat/lon point. I think this is now moot; all of out forecast pages -- and any corresponding observation requests -- will be done with the initial lat/lon point. From now on this value _should_ implicitly be true, so we can dump it.
  
`withinGridCell` is a little more complicated. For any user's lat/lon point, we know of a WFO grid cell in which that point resides. But it's not necessarily (or even usually) the case that the closest observation station is in the same WFO cell as the user's search location. I can't immediately recall the specific reason we needed this attribute. It might have had to do with the "grid vs point" issue mentioned above, which no longer matters. I think this is also safe to dump for now.

eric-gade (approved):
This is excellent! Gives me a lot of confidence in the interop as well. Thanks for working this all out 👍

api-interop-layer/data/obs/index.js
jamestranovich-noaa: Ahh gotcha! Thanks for the context, glad to hear that we can streamline this a bit!

PR #1769, opened 2024-09-13
[Interop layer] Add top-level error handling to Drupal

greg-does-weather:
## What does this PR do? 🛠️

In the current site, we catch API request failures and put an error block around them. With the switch to the interop layer, I forgot to add an exception handler. This PR adds a try/catch around the single call to the interop layer and modifies the point page template to show a top-level warning if that failed.

@partly-igor I just slapped some text into the error box. Right now it says
> ~Something went wrong loading this page. Unfortunately we don't have any more information to share right now, but we'll look into it. In the meantime, please try again in a few minutes.~

> Something went wrong loading this page. Please try again in a few minutes.

Happy to change it to whatever.

image

### DIFF
--- a/web/modules/weather_data/weather_data.module
+++ b/web/modules/weather_data/weather_data.module
@@ -13,9 +13,13 @@ function weather_data_template_preprocess_default_variables_alter(&$variables)
         $weatherMetadata["point"] = ["lat" => $lat, "lon" => $lon];
 
         $fetch = $container->get("http_client");
-        $data = $fetch
-            ->get("http://api-interop-layer:8082/point/$lat/$lon")
-            ->getBody();
+        try {
+            $data = $fetch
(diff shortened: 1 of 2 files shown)


partly-igor:
I think visually looks good, but wanted to tag in @kmranjo on the copy, she brought up some good questions:

>Do we have a way for it to ping us when something goes wrong? Do they need to wait a few full minutes, or could they refresh right away? Do we have any kind of feedback mechanism to report a problem? As it stands, without more info I'd just keep it to:

>> Something went wrong while loading this page. Please try again in a few minutes.

greg-does-weather:
Those are good points. Currently we don't have a way to get notified of these events, but the monitoring work that James is leading should eventually get us there. But even when we can, we probably shouldn't say we'll look into it because we might not.

I like Katrina's alternative a lot better.

eric-gade (approved):
👍

PR #1771, opened 2024-09-13
[Interop layer] Fix charts

greg-does-weather:
## What does this PR do? 🛠️

The data returned by the interop layer is nested under unit properties, like this:

```js
{
  ...
  temperature: {
    degF: 32,
    degC: 0
  }
}
```

So when we map in the data for the charts, we need to account for that. No biggie.

### DIFF
--- a/web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
+++ b/web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
@@ -1,7 +1,13 @@
-{% set times = hours | map(h => h.time) | json_encode %}
-{% set temps = hours | map(h => h.temperature) | json_encode %}
-{% set feelsLike = hours | map(h => h.apparentTemperature) | json_encode %}
-{% set pops = hours | map(h => h.probabilityOfPrecipitation) | json_encode %}
+{#
+  Available variables:
+    hours - An array of all the hours that should be represented in the table.
+    alerts - An array of all the alerts that intersect with those hours.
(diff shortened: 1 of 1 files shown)


web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
```diff
@@ -1,7 +1,13 @@
...
-{% set feelsLike = hours | map(h => h.apparentTemperature) | json_encode %}
-{% set pops = hours | map(h => h.probabilityOfPrecipitation) | json_encode %}
+{#
+  Available variables:
+    hours - An array of all the hours that should be represented in the table.
+    alerts - An array of all the alerts that intersect with those hours.
+#}
+
+{% set times = hours | map(h => h.hour) | json_encode %}
+{% set temps = hours | map(h => h.temperature.degF) | json_encode %}
```
jamestranovich-noaa, line 8: We can't toggle F/C yet?

jamestranovich-noaa (approved):
one minor question, LGTM

web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
greg-does-weather: Correct. There's an in-process design ticket for it, and we opted to postpone the engineering work until after the interop layer goes in because it transparently provides data in all the units we expect at once, so swapping between them can be completely client-side.

PR #1772, opened 2024-09-13
Wind Chart

eric-gade:
## What does this PR do? 🛠️
Implements #1634, the wind speed, gust, and directions chart

Well, this one was a little more complicated than the others. Drawing the arrow with the correct angle and the cardinal direction text is not something that ChartJS is equipped to do in a well-defined way. 

Instead, I made use of the `afterDraw` plugin, which gives you a canvas drawing context and some basic geometry about the labels and the overall chart. With this information I was able to paint the arrow and text to the screen.
  
Note that the wind info labels are _below_ the time labels, which is not according to the design. [This bug](https://github.com/chartjs/Chart.js/issues/11739) is preventing putting them above the time labels. In consultation with @partly-igor , we have put them below the times.

Screenshot 2024-09-13 at 1 20 36 PM
Screenshot 2024-09-13 at 1 20 45 PM


web/themes/new_weather_theme/templates/partials/hourly-table.html.twig
```diff
@@ -2,12 +2,17 @@
...
+{% set windDirections = hours | map(h => h.windDirection) | json_encode %}
 
 <wx-hourly-table class="display-block position-relative margin-top-2 bg-white">
   <h5 class="wx-visual-h3 font-heading-md text-normal text-primary-darker margin-top-0 margin-bottom-205">Hourly forecast</h5>
 
   <div class="margin-bottom-205">
-    <span class="font-family-mono text-primary-dark">Temperature</span>
+    <span class="font-family-mono text-primary-dark">
+      {{ "Temperature" | t }}
+    </span>
```
greg-does-weather, line 15: Ooh, good catch

greg-does-weather (approved):
That's some plain ol 'wizardry in the chart labels. Nicely done!

partly-igor:
It's looking really good. Visually two things: 

1. The direction arrow and text should match the Wind color (`secondary-darker`) instead of the gusts color. 
2. The text looks just a tiny bit off-center from the time and the direction arrow (example below). I am wondering if that text is left aligned, but the time text is centered, and so we should center-align the cardinal direction text? Or maybe it needs another 1-2px offset to the left?

Screenshot of text misalignment on wind chart

eric-gade:
> 2\. I am wondering if that text is left aligned, but the time text is centered, and so we should center-align the cardinal direction text? Or maybe it needs another 1-2px offset to the left?
  
Yeah, unfortunately we are manually aligning this text, since the chart is made using a 2d canvas. I can double check the measurements and post a follow up later today for your review

Issue #1551, opened 2024-08-09
Setup "outside" tests to run on every PR


jamestranovich-noaa:
### Description

Once we have "outside" tests working (https://github.com/weather-gov/weather.gov/issues/1549) _and_ documented how and why to use "outside" vs Drupal tests (https://github.com/weather-gov/weather.gov/issues/1550) then we should set up these tests to run on every PR as part of our usual CI workflow. 

### Questions and Considerations

None.

### Acceptance Criteria
- [ ] We have outside tests running against each PR/merge.

jamestranovich-noaa:
This isn't as straightforward as it seemed, because we have a fairly complicated Github Actions setup with Playwright that we will have to re-do and combine when we migrate to Gitlab CI. There's also a larger issue that might need dev discussion: eventually I think e2e testing should be fully on the dedicated test environment. And this kinda ties in with the deprecation of Cypress as well. 

I think I'm going to punt this for the time being since we can still run outside tests locally and currently only the API functionality is being tested via outside tests.

jamestranovich-noaa:
Wrote up larger issue thoughts here: https://github.com/weather-gov/weather.gov/issues/1795 -- I think we should address outside testing as part of that effort. Leaving this ticket open for a bit longer, in case we want discussion around outside testing functionality.

PR #1790, opened 2024-09-17
gh: run outside tests

jamestranovich-noaa:
## What does this PR do? 🛠️

Run outside testing. I don't think there's a need for us to set up a separate test environment since, well, GH actions also tear down/set up Docker containers every time. Addresses https://github.com/weather-gov/weather.gov/issues/1551


jamestranovich-noaa:
Closing because I'm realizing that this is part of a larger testing issue.

Issue #1681, opened 2024-08-31
AFD for Alaska.


coreypieper:
There is an issue with the forecast discussion link on the beta site for locations within the WFO Anchorage CWA. This is likely due to confusion with the dual domains they have. So they have dual GFE domains which have identifiers of (ALU and AER). However, products that cover their entire CWA, like the AFD, still utilize their overall WFO identifier AFDAFC. 

The link to the AFD on the beta site currently is: https://beta.weather.gov/afd/AER. This of course returns an error because AFDAER does not exist.

Consideration
- Could this be done in twig via urls

eric-gade:
Note to self that we also need to handle the 404 case here in such a way that the whole site doesn't break

jamestranovich-noaa:
I think the 404/500 is related to https://github.com/weather-gov/weather.gov/issues/1729

coreypieper:
At [this page](https://beta.weather.gov/point/61.217/-149.9?#daily) for Anchorage, the AFD link above the forecast doesn't work, and there is no Your Local Forecast Office link below the forecast.

PR #1804, opened 2024-09-20
Fix Anchorage WFO Promo links

eric-gade:
## What does this PR do? 🛠️
Fix for #1681 

We "normalize" the WFO code -- meaning we make ALU and AER into AFC -- in two places:
- In the WFOPromoBlock code, so that it attempts to retrieve the information for the Anchorage WFO info page and links to it on a given point location page;
- In a new custom Twig filter, which lets us apply this transformation from within templates as needed

## Testing
These two links should have both (a) a correct AFD link at the top of the forecast daily page, and (b) The correct link and text at the bottom linking to "Your weather office" (Anchorage/AFC):
- http://localhost:8080/point/55.204/-162.704?#daily
- http://localhost:8080/point/61.217/-149.9?#daily

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherEntityService.php
+++ b/web/modules/weather_data/src/Service/WeatherEntityService.php
@@ -123,15 +123,9 @@ class WeatherEntityService
      */
     public function normalizeAnchorageWFO(string $wfo): string
     {
-        $inAnchorageCodes = [
-            "aer",
-            "alu"
-        ];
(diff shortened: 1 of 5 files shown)


greg-does-weather (approved):
I am not here

jamestranovich-noaa (approved):
GREG. NO.

mgwalker:
Yes

Issue #1660, opened 2024-08-28
JSON:API ip address filtering


jamestranovich-noaa:
### Description

We need a way to filter JSON:API requests for specific ip ranges. Could be on the cloud.gov end. 

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

colinmurphy01:
Is this a blocker prior to briefing the DSSBuilder team on the #1434 API

jamestranovich-noaa:
Not a blocker. We need to talk with DSS Builder anyhow, to get more information: 1. their IP ranges (where they will upload from) 2. what other metadata they might also provide (we probably want to capture everything, even if we don't use it all)

colinmurphy01:
Cool. We have a meeting setup Monday AM. Hopefully we can get our questions answered then.

colinmurphy01:
I believe this is unblocked now?

jamestranovich-noaa:
Yes, this is now unblocked, with the caveat that we need to test and verify JSON:API on beta.weather.gov (which should happen Thursday).

PR #1805, opened 2024-09-20
ip address filtering for the `/jsonapi` endpoint

jamestranovich-noaa:
## What does this PR do? 🛠️

This PR configures our environment to add a standalone proxy app (using Caddy) that will reverse proxy to Drupal. This route service allows us to do IP address filtering for the `/jsonapi` endpoint only. Closes https://github.com/weather-gov/weather.gov/issues/1660

- This has only been deployed in the `james` cloud.gov space. 
- Not all environments will need this unless they have an `uploader` user. 
- We have an IP address from NWS, but let's not deploy in beta until we can double check the IP address: it could be NAT'd. 
- Architectural diagrams have been updated but `sandbox` and `beta` do not actually have this setup (yet)

## Testing

- Confirm that https://weathergov-james.app.cloud.gov/ is up and working
- `curl https://weathergov-james.app.cloud.gov/jsonapi` should return a 403
- [optionally] DM me and provide your IP address and I can push a proxy update; then the above endpoint will return JSON for you only.


eric-gade (approved):
Oops! I thought I had already approved this one! 👍

Issue #1818, opened 2024-09-24
[API] Product requests to fake WFO locations return 200 with empty result


eric-gade:
### Description

If you make a request for a specific product with an invalid WFO code, you still get a `200` (successful) result, just with an empty list as the result.
  
[Example](https://api.weather.gov/products/types/AFD/locations/OOOOO).
  
As a user of the API, I would expect some kind of error code response, maybe even a `404` with an explanation that the given location is invalid.

### Acceptance criteria

- [ ] Requests to product listings (like the AFD example above) return a `4xx` status code for invalid locations, along with an explanation message

colinmurphy01:
Did we discuss this in API sync and diagnose the issue?

Issue #1729, opened 2024-09-09
500 server errors when trying to retrieve an unknown AFD id


jamestranovich-noaa:
### Description

We recorded a 500 error in New Relic today. (NR query: `SELECT * FROM Transaction where http.statusCode in (500) AND request.headers.host = 'beta.weather.gov' SINCE today`)

The details of the above error:
- `GET` `/afd/PQR/f627960c-dde9-47ce-a8db-292dbf667141` 
- code: `WebTransaction/Custom/weather_routes.afd.wfo_id`

We know that AFDs can eventually "expire" but grabbing an expired id shouldn't break our site. More broadly, we should be more tolerant of user input in the url. To replicate: visit https://beta.weather.gov/afd/PQR/foobar -- this currently breaks and gives us a 500 server error.

### Acceptance criteria

- [ ] We return a 404, and not a 500, on invalid AFD ids.

PR #1820, opened 2024-09-24
AFD 404s (and other things!)

eric-gade:
## What does this PR do? 🛠️
Deals with #1729, in addition to more general problems with API response handling.

Before this PR, the `DataLayer`'s `fetch` function was treating API responses of 4xx and 5xx both as generic `500` errors. Those generic errors are difficult to properly "catch" and reason about in the Drupal routing system.
  
We have changed this so that we make the appropriate distinction. Additionally, the result of `fetch` is more comprehensive (an array of information, including whether or not the result was an error-level response).
  
Finally, we have wrapped some of the more API-request-heavy parts of the AFD controller in try blocks. All exceptions caught from API requests here will trigger a `NotFoundHttpException()`, which will lead to the generic 404 page.

## Surfaced Issues
In addition to the source ticket, we also identified #1818 (in which the API does not respond as one might expect).

### DIFF
--- a/web/modules/weather_routes/weather_routes.routing.yml
+++ b/web/modules/weather_routes/weather_routes.routing.yml
@@ -25,7 +25,7 @@ weather_routes.grid:
     # Public endpoint
     _access: 'TRUE'
     # Validate path parameters
-    wfo: '^[A-Za-z][A-Za-z][A-Za-z]$'
+    wfo: '^[A-Za-z]{3}$'
     gridX: '^[0-9]*$'
     gridY: '^[0-9]*$'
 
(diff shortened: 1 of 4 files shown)


web/modules/weather_routes/weather_routes.routing.yml
```diff
@@ -75,6 +75,7 @@ weather_routes.afd.wfo_id:
     no_cache: 'TRUE'
   requirements:
     _access: 'TRUE'
+    wfo_code: '^[A-Za-z][A-Za-z][A-Za-z]$'
```
jamestranovich-noaa, line 78: `^[A-Za-z]{3}$` isn't supported?

jamestranovich-noaa (approved):
one small question re: regexp but looks great

PR #1824, opened 2024-09-25
Remove fix permissions recipe

partly-igor:
## What does this PR do? 🛠️
This PR removes the `fix-permission` recipe from the Makefile because running that command is creating a large set of changed files with git.  Suggested by @eric-gade 

We may need to adjust file permissions manually without this in place.


jamestranovich-noaa (approved):
:crying_cat_face: LGTM

Issue #1600, opened 2024-08-22
[Interop layer] Test alert description parsing into markup


greg-does-weather:
### Acceptance criteria

- Tests have been implemented or modified to capture the bug
  - [ ] what/when/where headings are parsed correctly
  - overview text is parsed
    - [ ] when there is an overview
    - [ ] does not parse a string with embedded ellipses as an overview

Issue #1594, opened 2024-08-22
[Interop layer] Parse links in alert descriptions


greg-does-weather:
### Description

This logic already exists in PHP and needs to be translated over to the interop layer.

### Acceptance criteria

- Tests have been implemented or modified to capture the bug
  - [ ] turns valid .gov URLs into links
  - [ ] ignores non-.gov URLs
  - [ ] ignores URLs with embedded username/password
  - [ ] ignores non-HTTPs URLs
  - [ ] marks weather.gov URLs as internal/not external
  - [ ] marks non-weather.gov URLs as external
  - [ ] properly parses multiple links in the same description

PR #1827, opened 2024-09-26
Alert description / url parsing tests

eric-gade:
## What does this PR do? 🛠️
Implements tests outlined in #1600.
  
Note also that it appears the work from #1594 was already done. This PR implements the corresponding tests for those as well, finishing the port of tests from `WeatherAlertParser.php.test`

It's a straight translation.


jamestranovich-noaa (approved):
LGTM. One thing I'm not sure about, can we remove `api-interop-layer/_tests_to_copy/Test/WeatherAlertParser.php.test` though, if we've ported all of it over?

eric-gade:
> LGTM. One thing I'm not sure about, can we remove `api-interop-layer/_tests_to_copy/Test/WeatherAlertParser.php.test` though, if we've ported all of it over?

Oops! Yes, good catch

PR #1840, opened 2024-09-30
Fix precipitation table so it covers all hours of the forecast day

### LINKED ISSUE #1145: Discuss Options for Precipitation Tables going into following day
6-hour precipitation tables are based on Zulu time.  So, for instance in Medford, OR under Pacific Daylight Savings Time we have 6-hour precipitation time brackets as:

- 5am-11am (12Z-18Z)
- 11am-5pm (18Z-00Z)
- 5pm-11pm (00Z-06Z)
- 11pm-5am (06Z-12Z)

And for Fort Worth Texas (CDT) we have:
-7am-1pm (12Z-18Z)
-1pm-7pm (18Z-00Z)
-7pm-1am (00Z-06Z)
-1am-7am (06Z-12Z).

In its current state, beta.weather.gov shows this for Medford, OR on day 1:
image
and shows this all following days where precipitation amounts are available:
image

We probably want to choose the breakpoint that most closely corresponds with our day/night breakpoints in the daily forecast (6am-6am).
So for Medford, we'd want to use 12Z-18Z, 18Z-12Z, etc, instead of the current setup which starts at 18Z.

But what if the current time is 9am, so precipitation forecast for 6am-9am is no longer relevant.  Do we reduce the 6hr amount by half, to more accurately reflect the situation?  Or do we leave the full 6hour amount in there, even though we're showing only 3 hours.  Or do we keep the full 6 hour time in there, even though some of those hours are in the past?  @coreypieper and others, we need to discuss this and find a solution.

Here's the current setup for Fort Worth, TX for the first day:
image
and all other days:
image

### Acceptance criteria

greg-does-weather:
## What does this PR do? 🛠️

Quantitative precipitation (QPF) is forecast in 6-hour chunks, based on UTC rather than local time. As a result, timezones that are not an even multiple of six hours from UTC will be covered by more than 6 QPF periods (e.g., in UTC-7 MST or PDT, the QPF periods that cover a given forecast day will be 0400 day 1 to 1000 day 2, or five QPF periods). Previously, we only took the four QPF periods beginning with the first one that intersected 0600 day 1.

This PR changes the logic so each forecast day includes all of the QPF periods that cover the forecast day. Some QPF periods cover more than one forecast day, in which case that period is shown for both days.

- closes #1145

There is a lot of legacy rework in here regarding how daily forecasts are handled. Way back in the day, we had separate "detailed" and "extended" forecasts, but we harmonized those some time back. This PR originally did the work to harmonize them in the backend to make things simpler throughout. Rather than pull that out of this PR, I opted to leave it in since all of that work will soon be replaced by the interop layer anyway.

Note that the last QPF period extends beyond 6am:

image


tests/api/serve.js
```diff
@@ -55,9 +55,14 @@ const processDates = (obj, usingHourly = false) => {
...
+        //
+        // Alternatively, if we are not parsing hourly data but our parent key
+        // is "values" and our key is "validTime", then this is the hourly data
+        // stuffed inside the gridpoints API return. Do the same thing.
         if (
-          usingHourly &&
-          ["startTime", "endTime", "validTime"].includes(key)
+          (usingHourly &&
+            ["startTime", "endTime", "validTime"].includes(key)) ||
+          (!usingHourly && parent === "values" && key === "validTime")
```
jamestranovich-noaa, line 65: gosh, maybe it's just Monday but I'm having a hard time parsing this. can we pull out these as descriptive boolean variables? Something like:

```js
  const needsAlignment = ["startTime", "endTime", "validTime"].includes(key);
  const isHourlyInGridPoints = !usingHourly && parent === "values" && key === "validTime";

  if (usingHourly && needsAlignment || isHourlyInGridPoints) ...
```

and if so, is that boolean logic correct?

jamestranovich-noaa (changes requested):
Looks good save for a couple eslint errors and one confusing `if` statement (to me, anyways)

tests/api/serve.js
greg-does-weather: Oof, yeah, this is gnarly. And I wrote it long enough ago that I'm not certain anymore... 😅 

I'll spend some time walking through it today to understand it again and then I'll clean it up so it's not as hairy. Good catch!

greg-does-weather: This was way too convoluted. I did understand it after I sat with it for a minute, but it was a hack because I didn't think very long about how to determine if the data being processed was hourly forecast data. It turns out, when I spent a little time on that, there was a pretty easy regex I could use down below and then all of this wonkadoo `if` logic just goes away.

jamestranovich-noaa (approved):
LGTM, I think -- there's a lot of data changes but the code looks good to me.

greg-does-weather:
To be blunt about it, it would ultimately have been faster to strip out all the other changes I made and had only the qpf changes. :sigh: But all of this code, aside from the proxy stuff, will be thrown away soon so whatevs, I guess.

PR #1842, opened 2024-09-30
do not do a rolling strategy for multiple apps

jamestranovich-noaa:
## What does this PR do? 🛠️

TIL but `cf` does not allow one to set strategy when deploying multiple apps. Not going to merge this just yet because I'm not sure how this will impact beta deploys yet.


jamestranovich-noaa:
I'm going to close this in favor of a new PR that will push each app individually because that's the best way to handle rolling deployments.

PR #1843, opened 2024-09-30
Fix script bundling

greg-does-weather:
## What does this PR do? 🛠️

The Leaflet, ESRI and ESRI vector script tag `id` attributes were mistakenly removed, so the alert map module could never determine when those scripts had loaded. Additionally, Drupal controls the rendering order of our script tags so we cannot be certain that the DOM will be fully loaded when we load, so if Leaflet isn't already available, also attach a `DOMContentLoaded` event handler.

Alternatively, we could keep calling the Leaflet check on a timer, but in the even that Leaflet never loads, we'd just be burning CPU cycles on nothing. Using event-based loading is more processor-friendly.


jamestranovich-noaa (approved):
LGTM. we also need to revert the git revert that stops this from deploying on beta. but let's make sure this works on staging first.

greg-does-weather:
Revert the reversion!

Issue #1808, opened 2024-09-23
Ingest Weather Story XML files from DSS Builder


colinmurphy01:
@colinmurphy01 to add acceptance criteria for this ticket and regroup with @jamestranovich-noaa and size.

Related: #1793

colinmurphy01:
@colinmurphy01 to update the details (including XML metadata)

jamestranovich-noaa:
We ingest all details for the weather story now into drupal content types -- one per image. Only office, title, description, and full image are required. The rest of the attributes are optional.

jamestranovich-noaa:
Yup, I've let them know, with the caveat that this isn't in beta yet. But, soon!

PR #1853, opened 2024-10-02
weather story upload content type

jamestranovich-noaa:
## What does this PR do? 🛠️

Addresses https://github.com/weather-gov/weather.gov/issues/1808 -- sets it up so that we can ingest weather story attributes from DSS builder.

With weather stories, we want to ingest all attributes possible from DSS builder. But we require only a few attributes: namely, `title`, `field_office`, `field_fullimage` and `field_description`. 

For reference, here is what DSS builder currently outputs (note that these will change (almost) every day): 

- Sample weather story xml: https://www.weather.gov/source/hnx/WxStory/WeatherStory.xml
- Sample weather story image: https://www.weather.gov/images/hnx/WxStory/WeatherStory2.png

Note that, annoyingly, outside tests will sometimes fail with a 503 -- server unavailable error. This is because of file system locking inside Drupal. I haven't figured out a fix for this yet, but if this happens locally, just wait and try again. (tracked as https://github.com/weather-gov/weather.gov/issues/1857 )


tests/playwright/outside/api.spec.js
```diff
@@ -6,6 +6,34 @@ const { describe } = test;
 
 const API_ENDPOINT = "http://localhost:9080/jsonapi";
 
+// helper function to upload files to the JSON API
+const uploadFile = async (request, where, filename) => {
+  const binaryData = await fs.readFileSync(
```
greg-does-weather, line 11:
```suggestion
  const binaryData = fs.readFileSync(
```

It's a synchronous function, nothing to await. :)

greg-does-weather (approved):

greg-does-weather:
It'd be cool to see this in action sometime.

tests/playwright/outside/api.spec.js
jamestranovich-noaa: d'oh!

Issue #1766, opened 2024-09-12
Visualization datapoint labels cut off. 


coreypieper:
When values on the bar graphs that top out at 100% have data close to 100, the labels are cut off. 
![IMG_6757](https://github.com/user-attachments/assets/104e6c78-85fb-4067-93ec-5ae190229535)

Additionally, sometimes x axis values are not showing up at all on iphone SE and Android/Pixel 7a. We were only able to reproduce when bars are ≥95%.  

![image](https://github.com/user-attachments/assets/d23fad9f-0f9e-4228-aece-9e52ec36dc1e)

colinmurphy01:
@igorkorenfeld and @eric-gade do you have capacity to look into this when you pick up #1702?

colinmurphy01:
Hurricane this week might have conditions that can be used for testing.

partly-igor:
Noting that #1858 helped but didn't totally get rid of the issue for x-axis. 

The pattern I am seeing is that the x-axis labels disappear only for the "Today" bar charts whenever the first hour is odd numbered (for example 3pm, 5pm, etc..). Notably we skip labels for odd hours on the bar graph, and it may be somehow related to that. The x-axis labels do display when the first hour is even-numbered, so it doesn't seem like the chart width being too narrow is the issue, but perhaps specific chart widths do create that problem.

greg-does-weather:
To clarify, the thing where the labels on the chart itself were being cut off is resolved, right? The remaining issue is the X axis labels?

partly-igor:
That's right – and cool, I forgot we only did it for temperatures so that sounds like a good plan!

Issue #1702, opened 2024-09-05
Set chart widths dynamically


greg-does-weather:
### Description

In the first implementation of our charts, they are fixed at the width of their containers and all of the available data is smooshed into the charts. For charts showing longer periods of time, this can make the chart difficult to read.

Instead of fixing the chart's width to its container, we would like to dynamically determine the chart's width based on two factors:
- an explicit width for the hours on the X axis
- the number of hours being displayed

The chart container should clip the horizontal part of the container that extends beyond the container itself, and the container should support scrolling. Scrolling will also be accompanied by scroll buttons similar to the hourly details table.

The width of a single hour on the X axis should be variable based on which screen width break points we're currently in, and per type of chart. As a result, when the screen crosses break point thresholds, the chart width needs to be recomputed and the chart needs to be redrawn accordingly.

#### Chart widths
|Chart type|mobile|mobile-lg|tablet|tablet-lg|desktop|
|---|---|---|---|---|---|
|line|720|720|940|951|951
|bar|720|720|940|951|951

### Acceptance criteria

- [ ] Charts are sized appropriately according to the descriptions above
- [ ] Changes have been tested for accessibility

Related #1766

greg-does-weather:
tbd: we'll figure out what the widths should be and fill those in

partly-igor:
I have tested out different widths for the chart and believe that setting the chart width itself may allow us to achieve this effect. After manually testing out different options I have come with the set that I've added to the top of the ticket. This allows for the following number of visible hours at the smallest (hours min) and largest sizes (hours max) within a breakpoint. (Hours may be off by 1 :) )


Viewport Size | Breakpoint | Chart width | #Hours  Min | #Hours Max
-- | -- | -- | -- | --
Mobile | 0 | 720 | 9 | 13
Mobile-lg | 480 | 720 | 15 | 21
Tablet | 640 | 940 | 15 | 23
Tablet-lg | 880 | 951 | 21 | 25
Desktop | 1024 | 951 | 25 | 25

partly-igor:
Completed via #1858

PR #1858, opened 2024-10-03
Set chart widths (with scroll), height, and padding

partly-igor:
## What does this PR do? 🛠️
This PR sets a "data-width" for each data point at different breakpoints and wraps the charts in a scrollable container so that the data in the charts fits within the window. It also adds in padding to each chart so that labels are not cut off for charts that are representing percentages which have values nearing 100%. 

This should address both #1702 and #1766, though for #1766 I am unsure of the x-axis issue 

This PR doesn't add the scrolling arrows, which will be implemented in a separate PR. 

Screenshot of adjusted charts on desktop

Screenshot of adjusted charts on mobile


partly-igor:
Because the first day has dynamic widths, the first day needs to be set more dynamically.

eric-gade (approved):
Code looks good to me here.
  
There is one potential problem though, which is a larger issue: if you try to scroll on mobile (via touch), the other touch events on the charts are activated at the same time.
  
You can see this live in [my environment](https://weathergov-eric.app.cloud.gov/) if you want to test it yourself

partly-igor:
Good point. I am going to merge this in so we have a somewhat better version of the charts on mobile working for next week but will need to explore a solution to this. Maybe there is a way to disable the tooltip on drag/scroll.

Issue #1605, opened 2024-08-22
[Interop layer] Test hourly forecast


greg-does-weather:
### Acceptance criteria

- Tests have been implemented or modified to capture the bug
  ~- [ ] returned data matches the expected schema~ moved to #1848 
  - [x] the first hour begins at the start of the next whole hour
  - [x] each hour begins at the top of the hour
  - [x] if the current local time is before midnight, the list of hours extends through 6am the next day
  - [x] if the current local time is between midnight and 6am, the list of hours extends through 6am the *next* day
  - [x] future days are 6am to 6pm, except possibly the last day which may have fewer hours in it
  - [x] there are no past hours
  - [x] hours are presented with the correct UTC offset for the location's timezone

greg-does-weather:
closed by #1860

PR #1860, opened 2024-10-04
Interop hourly data tests

eric-gade:
## What does this PR do? 🛠️
Implements #1605 

Minor refactoring to the underlying code so that the ACs are met and also so parts are easier to test.


jamestranovich-noaa (approved):
Some eslint warnings here -- but code LGTM

PR #1866, opened 2024-10-07
Fix: QPF table should say "rain" if there is no ice or snow, instead of "water"

greg-does-weather:
## What does this PR do? 🛠️

- Just what it says on the tin
- Updates the bundle metadata so it's more descriptive of the available test data
- Adds tests to capture this behavior of the QPF table


jamestranovich-noaa (approved):
LGTM.

One minor comment, do we need to update the github bot here? It's rather annoying when it says:

> This PR modifies theme Javascript or CSS assets but does not update the theme libraries file. Did you mean to update the appropriate version information in the libraries file?

greg-does-weather:
Yeah. I want to get rid of this bot and instead update the libraries file automatically on deploy. Once we get through pushing out the interop layer, I'd love to spend some time on that!

greg-does-weather:
added #1878

PR #1868, opened 2024-10-07
Remove custom logic for determining which X axis ticks to display

### LINKED ISSUE #1766: Visualization datapoint labels cut off.

greg-does-weather:
## What does this PR do? 🛠️

When we were building the temperature chart, we originally implemented logic to only show the even-numbered hours on the X axis. However, we bumped into a lot of weird behavior with that, so we opted to just let Chart.js figure it out. That seemed to work pretty well.

However, for the hourly precipitation chances and humidity charts, we left in that "only evens" logic. This PR removes that logic so that these two charts also behave like the temperature chart.

- closes #1766

### DIFF
--- a/web/themes/new_weather_theme/assets/js/charts/hourly-humidity.js
+++ b/web/themes/new_weather_theme/assets/js/charts/hourly-humidity.js
@@ -64,7 +64,7 @@ for (const container of chartContainers) {
     },
 
     data: {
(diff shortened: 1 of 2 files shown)


partly-igor (approved):
This looks great and seems to fix, thank you!

PR #1871, opened 2024-10-07
Implementing (synced) scroll buttons for charts

eric-gade:
## What does this PR do? 🛠️
Adds scrolling arrow buttons to the hourly forecast charts in the 7-day.
  
In this implementation, the scrolling is synced for all charts via the `data-sync-scrolling` attribute on chart containers.

* Moved scroll buttons into their own template partial
* Moved chart area into its own template partial, and out of the hourly-table element
* Small cleanup

### DIFF
--- a/web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
+++ b/web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
@@ -33,7 +33,7 @@ const drawChart = (container, config) => {
 const setupScrollButtons = (container) => {
   const isSynced = container.dataset.syncScrolling === "true";
   const wrapper = container.closest('.wx-chart-wrapper');
-  const left = wrapper.closest('.wx-chart-wrapper').querySelector('.wx-scroll-button[data-direction="left"]');
+  const left = wrapper.querySelector('.wx-scroll-button[data-direction="left"]');
   const right = wrapper.querySelector('.wx-scroll-button[data-direction="right"]');
   if(!left || !right){
     return;
--- a/web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
+++ b/web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
@@ -30,5 +30,68 @@ const drawChart = (container, config) => {
   return new Chart(canvas, config);
 };
 
-export { drawChart };
-export default { drawChart };
+const setupScrollButtons = (container) => {
+  const isSynced = container.dataset.syncScrolling === "true";
+  const wrapper = container.closest('.wx-chart-wrapper');
+  const left = wrapper.closest('.wx-chart-wrapper').querySelector('.wx-scroll-button[data-direction="left"]');
+  const right = wrapper.querySelector('.wx-scroll-button[data-direction="right"]');
+  if(!left || !right){
+    return;
+  }
+
+  right.addEventListener("click", () => {
+    const canvasEl = container.querySelector("canvas");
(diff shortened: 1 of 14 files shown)


web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
```diff
@@ -30,5 +30,68 @@ const drawChart = (container, config) => {
   return new Chart(canvas, config);
 };
 
-export { drawChart };
-export default { drawChart };
+const setupScrollButtons = (container) => {
+  const isSynced = container.dataset.syncScrolling === "true";
+  const wrapper = container.closest('.wx-chart-wrapper');
+  const left = wrapper.closest('.wx-chart-wrapper').querySelector('.wx-scroll-button[data-direction="left"]');
```
greg-does-weather, line 36:
```suggestion
  const left = wrapper.querySelector('.wx-scroll-button[data-direction="left"]');
```
Shouldn't need the `.closest()` here, right? It should just match `wrapper` again? The right arrow one doesn't have it and seems to work fine.

web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
```diff
@@ -30,5 +30,68 @@ const drawChart = (container, config) => {
...
+  const right = wrapper.querySelector('.wx-scroll-button[data-direction="right"]');
+  if(!left || !right){
+    return;
+  }
```
greg-does-weather, line 40: Just a thought, but because this check is here, could we have the `drawChart` method also call `setupScrollButtons`? Then the only requirement to get the buttons to work is to include them in the twig template. No need to do anything with it here, but maybe a future refactor?

greg-does-weather (approved):
This is so cool. It is really impressive to see the charts scroll together! Code looks nice and tidy too! Great clean-up in here. Very cool!

There's an accessibility problem with the charts in general, but it is not introduced in this pull request. They don't using heading elements, so the charts are not findable, but they are still navigable. Even if you land on them, though, there's no information available. With this PR, screen readers will find the buttons without any context for what they do. Again, not a problem caused by this PR so I think it's reasonable to create a bug ticket to do a proper fix for the charts.

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Accessibility
    - [ ] Tab order is logical

> [!WARNING]  
> The charts have labels but the labels aren't headings, so they're not findable. I'm not sure a screenreader user would ever find them on purpose, and if they landed on them by accident, there's no context for what they're seeing.

## Device Matrix
    - [x] at 480px     (mobile)
    - [x] at 640px     (tablet)
    - [x] at 1024px    (desktop)

web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
eric-gade: Ah yeah that's a mistake, good eye

web/themes/new_weather_theme/assets/js/charts/WeatherChart.js
eric-gade: I thought about that but didn't want to overload that one function (and was being cautious -- without looking -- in case it was a "real" 16x per second canvas draw function or whatever). We can definitely get this on the next loop around

eric-gade:
@greg-does-weather thanks for reminding me, we meant to make a ticket about the headers thing yesterday when we did this.

https://github.com/weather-gov/weather.gov/issues/1880

PR #1872, opened 2024-10-07
[Interop layer] Fix overnight periods

### LINKED ISSUE #1852: [Interop layer] Investigate the logic for how we determine "overnight" periods
### Description

See https://github.com/weather-gov/weather.gov/pull/1823#discussion_r1785166989. The logic we use for determining whether we have an overnight period is incorrect. I suspect there is a simpler way to do this:

1. build the full list of calendar days
2. for the first calendar day...
  - one period, that period is night time
  - two periods, the first is day time and the second is night time
  - three periods, the first is overnight, the second is day, and the third is night

☝️ That should be guaranteed based on how we split the days around local midnight instead of around weather days. 

### Acceptance criteria

- [ ] We determine if the logic is correct
- If not...
  - [x] Tests have been added to capture the bug
  - [x] The bug is fixex

greg-does-weather:
## What does this PR do? 🛠️

Fixes a bug with how overnight periods were determined. The original logic was faulty because it assumed that if the first period of a day was not `isDaytime`, then the period must be overnight. However, that is only half the condition, because an overnight period only happens if there are also day and night periods in the same forecast day. It is totally valid for there to be only a single period in the forecast day, and our previous logic would call that an overnight period even though it cannot be.

This PR updates the logic to update the first period of the first forecast day to be overnight **IF AND ONLY IF** there are three periods in the first day. It also modifies the tests so that the times in the test are expressed in the local timezone so they are easier to reason about (hopefully 🤞).

For maintainability and readability purposes, it also removes the custom code for maintaining UTC offsets in Day.js objects, instead properly using the `dayjs().tz()` functions to convert arbitrary UTC offsets into the correct local timezone, and adds an additional test for the case where the offset does not match the local timezone.

- closes #1852

### DIFF
--- a/api-interop-layer/data/forecast/daily.js
+++ b/api-interop-layer/data/forecast/daily.js
@@ -3,36 +3,28 @@ import { convertProperties } from "../../util/convert.js";
 import { parseAPIIcon } from "../../util/icon.js";
 import { sentenceCase } from "../../util/case.js";
 
-const dayjsOffset = (iso8601) => {
-  const time = dayjs(iso8601);
-  const [, offset] = iso8601.match(/([-+]\d{2}:\d{2})$/) ?? [];
-
-  if (!offset || offset === "z") {
-    return time;
-  }
-
-  const [, direction, hours, minutes] =
-    offset.match(/([-+])(\d{2}):(\d{2})/) ?? [];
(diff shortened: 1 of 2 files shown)


jamestranovich-noaa (approved):
makes SO much sense. LGTM!

PR #1876, opened 2024-10-08
Eg 1604 interop daily alert tests

eric-gade:


api-interop-layer/data/alerts/index.test.js
```diff
@@ -1,6 +1,25 @@
...
+
+/**
+ * Test helper for creating arrays of mock
+ * day / hours objects
+ */
+const makeDayWithHours = (startTimestamp, numHours, timezone="America/New_York") => {
+  const hours = [];
+  const start = dayjs.utc(startTimestamp).tz(timezone);
+  // eslint-disable-next-line no-plusplus
+  for(let i = 0; i < numHours; i++){
```
greg-does-weather, line 15:
```suggestion
  for(let i = 0; i < numHours; i += 1){
```
I don't know why eslint prefers `+= 1` over `++`, but that's what it wants. 🤷

greg-does-weather (approved):
Looks good! Nice helper method there. 😄 

One minor note about the eslint thing to do with as you please. :shipit:

Issue #1870, opened 2024-10-07
Drupal upload locations for DSS builder


jamestranovich-noaa:
### Description

DSS Builder will upload the same file name repeatedly, and in this case Drupal will append numbers to the similar filename (`_1`, `_2`, etc) which might be confusing to both internal and external users so let's see if we can change the upload location in Drupal. 

Currently each filename is unique per WFO, so perhaps we can use a similar structure. Needs some investigation.

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

Issue #1857, opened 2024-10-03
outside tests will sometimes fail with a 503 -- server unavailable error


jamestranovich-noaa:
### Description

We have some kind of file system locking contention when running outside tests repeatedly, because we upload files again and again: 

```
120	03/Oct 18:59	php	Error	Symfony\Component\HttpKernel\Exception\HttpException: File "public://2024-10/test-upload_27.png" is already locked for writing. in Drupal\jsonapi\Controller\FileUpload->handleFileUploadFor
```

I'd like to fix the fundamental issue here, but if we can't address the root cause, the next best thing is to randomize the file upload names. 

### Acceptance criteria

- [ ] Outside tests can be run repeatedly in sequence without failing

colinmurphy01:
This feels important to fix prior to turning on the DSS integration?

Will randomizing file names impact the user experience?

jamestranovich-noaa:
Great question. This particular issue is (I believe) an odd Drupal/Docker interaction issue with tests. 

But good catch regarding user experience. I filed a separate ticket for figuring out a better way to handle upload of files with very similar names: https://github.com/weather-gov/weather.gov/issues/1870

colinmurphy01:
@jamestranovich-noaa has this been resolved?

jamestranovich-noaa:
Yes! Good catch. Closing

PR #1886, opened 2024-10-08
set weather story upload paths

jamestranovich-noaa:
## What does this PR do? 🛠️

Fixes https://github.com/weather-gov/weather.gov/issues/1870 _and_, as a bonus, #1857 as well.

This adds `filefield_paths` so we can configure where exactly uploads are placed in our filesystem.

The default behavior of Drupal prior to this change was:
- upload the file to the default location -- usually `[date:custom:Y]-[date:custom:m]` -- and append `_0`, `_1`, `_2`, etc, to the basename if a conflict is found
- this results in filenames like `2024-09/test-upload_67.pdf`
- this would be confusing and non-intuitive for users (which upload is for which WFO?)

This configuration changes it so that weather stories uploads are stored via `[node:field_office]/[node:content-type:machine-name]/field_fullimage/[file:ffp-name-only].[file:ffp-extension-original]` -- which is a mouthful but is probably best illustrated by an example: `WFO test office/wfo_weather_story_upload/field_fullimage/test-upload_6.png`

Note that we do a pretty nasty hack here. Because Drupal uploads the file to the default location first, then checks for conflicting filenames, and then moves the file afterwards, we set the initial upload to a random directory (literally `tmp/[random:number]` so that we are almost guaranteed to not conflict filenames. Then Drupal moves the file to the WFO directory above, and will re-do the conflict filename check again. If we didn't have this hack, then we'd get `WFO test office/wfo_weather_story_upload/field_smallimage/test-upload_57.png` even if that were the first file uploaded for that WFO. Ugh. We want to clean this up later but that most likely will involve overriding Drupal class functions. For now, anything stored in the public directory `tmp/` is considered, well, temporary.

Note that you will need to rebuild the test images.

`docker compose -f docker-compose.test.yml --profile test build`

And now you can re-run outside tests as many times as you wish, without 503 errors due to filesystem locking errors! :tada:


composer.json
```diff
@@ -24,6 +24,7 @@
         "drupal/core-composer-scaffold": "^10.3",
         "drupal/core-project-message": "^10.3",
         "drupal/core-recommended": "^10.3",
+        "drupal/filefield_paths": "^1.0@beta",
```
greg-does-weather, line 27: This currently doesn't meet our requirements for contributed modules:
- Do not use a beta or dev release of a module, even if it provides functionality we are seeking.

I think this might be something to discuss with product, around the risks and benefits.

greg-does-weather (approved):
Technically I'm on board with it. I'm less sure about the module. I'm gonna approve the PR because I see in Slack that Eric said to roll with it, but I'll defer to you/him on whether we should run it by product first just to be sure.

composer.json
jamestranovich-noaa: Good catch! @eric-gade let's discuss when you get back?

PR #1890, opened 2024-10-08
[Interop layer] Fix charts

greg-does-weather:
## What does this PR do? 🛠️

This PR fixes the data expectations between the interop layer and the charts. It also merges in the latest from `main` so that remaining interop layer work can happen on top of the latest release.

It probably makes the most sense to only review the last commit. Everything else is the merge from main.


greg-does-weather:
Should we perhaps rebase the interop layer onto main instead of merging? There are more commits captured in this PR than I expected. I only expected the commits to `main` since we last rebased.

jamestranovich-noaa:
I would suggest rebasing on `main` instead. It's confusing as-is...

greg-does-weather:
I agree. I'm gonna put this on ice until we rebase the interop layer again.

greg-does-weather:
superseded by #1898

Issue #1731, opened 2024-09-10
[Interop layer] Configure deployment to Sandbox and Staging cloud environments


greg-does-weather:
### Description

The interop layer is a new Node.js app that needs to be deployed to our cloud environments on all deployments.

- [ ] sandbox
- [ ] staging

Furthermore, networking needs to be configured so as to allow the `weathergov*` app to access the interop layer and for the interop layer to reach the public internet. If the configuration is reasonably straightforward, it might also be ideal not to expose the interop layer process to the internet.

### Acceptance criteria

- [ ] deploying to a sandbox environment also deploys the latest interop layer
- [ ] deploying to the staging environment also deploys the latest interop layer
- [ ] networking configuration is documented
- [ ] deployment and infrastructure diagrams are updated

jamestranovich-noaa:
Should we create a new cloud.gov space for the api interop layer? I ask because cloud.gov has [application security groups](https://cloud.gov/docs/management/space-egress/) that we can use for this purpose. Unfortunately these app security groups are applied only on a per-space basis (due to how cloud foundry is set up).

Basically, infra-wise, I propose we spin up a new space with a `closed-egress` asg (this is the default) and spin up a new interop layer app there. The documentation says,

- Requests being executed from within the space can only successfully be sent to other internal routes you have created in your organization.
- Any requests to the open internet or our brokered services will be blocked.

Thoughts?

greg-does-weather:
Can an application in one space be bound to a service in another space? The interop layer needs access to the database. It only needs access to the geospatial tables, though, not all of the other Drupal tables, so we might be able to get away with a second database that only serves the interop layer. We'd still need the `weathergov_geo_places` table for Drupal because we use that for one of our routing methods, though I suppose there's no reason that couldn't also be a call to the interop layer.

Of course, if the app in space A can be bound to the database service in space B, then there's no need to split up the database. 😛

What about deployment? I suppose deploying two apps to two spaces isn't much different from deploying two apps to a single space. 🤔 I'm already a little nervous about deployment timing – eg, what happens if the interop layer finishes deploying before Drupal does, and then the site is just nonfunctional for a few minutes while the Drupal deployment happens. I have no idea how you do an atomic update of multiple apps, but I doubt it matters much whether you use one space or two for that?

I'm not opposed to having a second space just for the interop layer, to be clear. I just haven't worked in an environment like that before and am curious!

jamestranovich-noaa:
[...]

PR #1897, opened 2024-10-09
API interop layer deploy actions

jamestranovich-noaa:
## What does this PR do? 🛠️

Sets up deploy for all environments except beta. Addresses https://github.com/weather-gov/weather.gov/issues/1731

It is not possible to have a rolling strategy for multiple apps defined in a manifest (see cloud.gov [documentation](https://docs.cloudfoundry.org/devguide/deploy-apps/rolling-deploy.html#deployment-process)). So here we are doing an app deploy one by one. Right now a "rolling" strategy doesn't matter that much because we only have one instance at any given time but it will matter when we add more instances down the line.


greg-does-weather (approved):
This all makes sense to me

PR #1900, opened 2024-10-10
[Interop layer] Fix alert start/end descriptions to include short timezone

greg-does-weather:
## What does this PR do? 🛠️

- Updates the way the start and end times of alerts are formatted to include the short timezone (eg, "CDT"). This requires another dayjs plugin, which I didn't previously enable so it wasn't working. Now it is. Also adds appropriate tests.
- Reviews the remaining "tests to copy" to remove the ones that have been covered. There are a few others left that we should document so we don't forget to add them.


jamestranovich-noaa (approved):
LGTM -- do we have interop test tickets for the last bit of `_tests_to_copy`?

greg-does-weather:
Ooh, not yet. Let me create those before I merge this, so I don't forget.

greg-does-weather:
Noted in #1904, which I will now pick up.

Issue #1904, opened 2024-10-10
[Interop layer] Missing tests


greg-does-weather:
- Alerts
  - [x] Alerts with only forecast zone(s) get the correct polygon
  - [x] Alerts with only fire zone(s) get the correct polygon
  - [x] Alerts with only county zone(s) get the correct polygon
  - [x] Alerts with only county SAME code(s) get the correct polygon
  - [x] We only return land alerts
  - [x] We do not return past (ended or expired) alerts

- Hourly forecast
  - [ ] We do not show gusts if they are less than 8mph more than sustained winds (***not even implemented***)
  - [x] ~"Feels like" temperature is properly derived from heat index, wind chill, or actual temperature (***not even implemented***)~ Non-issue. We use the `apparentlyTemperature` value from the API.

colinmurphy01:
Moving this to `doing` since you're working on it in #1907

PR #1907, opened 2024-10-11
[Interop layer] Add gridpoint tests, fix gust bug, fix alert bug

greg-does-weather:
## What does this PR do? 🛠️

- Fixes a bug where expired alerts could still be sent through
- Adds tests from #1904
- Fixes a bug where wind gusts would be preserved even if they were within 8mph of sustained winds
- Updates test data to have weird wind cases we can test again

This PR does *not* add a test for the wind gusts, because that's going to be a pretty involved test. That logic only makes sense in the `forecast/index.js` module, so it'll require building up sufficient dummy data to exercise all of the other forecast modules. Alternatively, if we can figure out how to get sinon to mock ESM exports, then we can mock out the other ones. Anyway, it was too much to add to this PR, so I propose we update #1904 to be just testing gusts.


api-interop-layer/data/alerts/index.test.js
```diff
@@ -73,11 +73,11 @@ describe("alert data module", () => {
...
-                sent: new Date().toISOString(),
-                effective: new Date().toISOString(),
-                onset: new Date().toISOString(),
-                expires: new Date().toISOString(),
-                ends: new Date().toISOString(),
+                sent: dayjs().subtract(1, "minute").toISOString(),
+                effective: dayjs().subtract(1, "minute").toISOString(),
+                onset: dayjs().subtract(1, "minute").toISOString(),
+                expires: dayjs().add(1, "minute").toISOString(),
+                ends: dayjs().add(1, "minute").toISOString(),
```
greg-does-weather, line 80: After fixing the bug with expired alerts, we also need to be more careful about the test data. Otherwise we get into a race condition where sometimes these tests pass and other times they fail, depending on just how fast the test executes.

api-interop-layer/data/forecast/index.js
```diff
@@ -56,6 +56,20 @@ export default async ({ grid, place }) => {
...
+    if (windGust && windSpeed) {
+      if (windGust.mph < windSpeed.mph + 8) {
+        // If the wind gust is less than 8 mph more than sustained winds, then
+        // we just set that to null. Get all the units.
+        Object.keys(windGust).forEach((unit) => {
+          windGust[unit] = null;
+        });
+      }
+    }
+  });
```
eric-gade, line 71: We could also just break this out into a helper / separate function, and/or make that a part of the hourly module. That way we could test it separately (rather than having to mock test data for the full forecast).

eric-gade (approved):
As far as I can tell, this specific change is working fine! 👍 
  
One thing that you might already be aware of: the wind directions are coming in as `undefined` in the wind chart. Probably a data structure issue:

Screenshot 2024-10-11 at 12 55 06 PM


## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

mgwalker:
Raging about the wind direction/arrows because they're working as expected on my branch, so now I gotta figure out what's up with my local caching... yay!

api-interop-layer/data/forecast/index.js
greg-does-weather: That was my original plan, but the way the hourly module works is it just iterates over all the data properties and inserts them into a map object that maps a time to different properties. As a result, when it's processing wind gusts, there's no way to know for sure whether the wind for that same time has been processed yet, or if it would happen later. After staring at it for a while, I decided it made more sense to me to keep the hourly module narrowly scoped to just normalizing the gridpoint data into, essentially, a list of hours.

But bumping it over to a helper function it :chef_kiss: perfect.

Issue #1703, opened 2024-09-05
Fix icon for Iphone OS/home screens


shadkeene:
We received feedback that when a user is saving beta.weather.gov on their Iphone, the NWS icon isn't showing.  I tested on Iphone SE and saved beta.weather.gov on the home screen via the Safari browser.  Here's what I see (see bottom two icons of beta.weather.gov on image):

![image](https://github.com/user-attachments/assets/945720ba-ffc6-4e12-819b-7e817fe3410c)

Acceptance criteria: 
- [ ] icon that displays on iphone home screens is same as other phone OSs

jamestranovich-noaa:
Ah, you beat me to it! (I swear I searched the issues for this particular issue). 

I found this in New Relic. Original issue description:

### Description

In New Relic metrics and dashboards, we have quite the number of requests for an apple touch icon (`/apple-touch-icon.png`). 

It looks like some folks are using beta.weather.gov as a website link on the iOS home screen. This could be important because apple touch icons are part of being mobile/SEO friendly. 

Apple touch icons are automatically generated by favicon generators these days -- so we can probably re-use the current favicon images.

### Acceptance criteria

- [ ] We have `apple-touch-icon.png` (and its variants, such as `apple-touch-icon-precomposed.png`, `/apple-touch-icon-120x120.png` and `/apple-touch-icon-120x120-precomposed.png`)

colinmurphy01:
As part of this we could deploy the app manifest capability to control this.

jamestranovich-noaa:
We are getting more and more errors in New Relic about this issue. To be specific, we have gotten over 30 "apple touch icon missing" errors _today_ alone. Agree that we should tackle this soon -- it's an easy lift once we agree on which icon to use. 

![image](https://github.com/user-attachments/assets/24bd893d-3e0d-425b-838b-e743679ec7ec)

colinmurphy01:
Thanks for the heads up on this. We'll reprioritize.

colinmurphy01:
@greg-does-weather #1924 includes the NWS logo as the app icon for now right?

greg-does-weather:
It's the NWS logo with a bright green background. Certainly not ideal! But icons cannot have transparent backgrounds, so I put in the green so we don't forget to make it something we actually want.

colinmurphy01:
This will be fixed #1924

PR #1924, opened 2024-10-15
Add an app manifest

greg-does-weather:
## What does this PR do? 🛠️

By adding an app manifest, we get some control over what it looks like when a user "installs" weather.gov to their device. One big gotcha is that the manifest icons cannot have transparency, so as a first pass, I gave them a green background so we could see how they render on-device.

rough start for #1703


jamestranovich-noaa (changes requested):
Are we addressing apple touch icons specifically in this PR? if so, don't we need a `rel="apple-touch-icon"` for some of these?

greg-does-weather:
No. iOS will use the icons defined in the app manifest. It only uses the `apple-touch-icons` if there is ***NOT*** a manifest.

greg-does-weather:
That's totally fair. I'll convert this to a WIP and we can revisit it later.

I _think_ for Android users, adding this manifest should cause the default browser to prompt users to install the app. For iOS users, it's more complicated because Apple hasn't seen fit to elevate web apps to first-class status yet (which is ironic, since that was the promise of iOS at the beginning, but 🤷‍♂️). But it should fix the icon issues anyway, and that's not nothing. 😛

colinmurphy01 (approved):

jamestranovich-noaa (approved):

Issue #1926, opened 2024-10-15
[R7 Implementation] Forecast item - 7 Days


partly-igor:
### Description
Implementation of the version "A" of the updates to the daily forecast item (day card). 

[Figma documentation ](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13026-159335&node-type=frame&t=O2OFQbCXtTC6Bzj8-11)

[Context/breakpoints](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13026-165127&t=O2OFQbCXtTC6Bzj8-1)

PR #1943, opened 2024-10-18
Hourly details toggle

eric-gade:
## What does this PR do? 🛠️
Implements the hourly details toggle from R7, as outlined in #1926.
  
**CAVEAT**: [We have opened a ticket to deal with the potential accessibility issues](https://github.com/weather-gov/weather.gov/issues/1944), discussed briefly below.

### Potential accessibility concerns
Per the initial discussion, we've implemented this toggle as a USWDS button group, with some small javascript for setting the appropriate toggle visuals / attribute changes.
  
However, I don't think this is the appropriate type of component from an a11y perspective. [Ticket here](https://github.com/weather-gov/weather.gov/issues/1944).
  
Additionally, we might need to consider our header structure given the updated designs. When displaying the table, we get a header describing the hourly details. But when displaying the charts, we only have the (further-down) headers describing each chart. 

<details>
<summary>Mobile</summary>

localhost_8080_point_21 305_-157 858(iPhone SE) (1)

localhost_8080_point_21 305_-157 858 (2)

</details>

<details>
<summary>Tablet</summary>

localhost_8080_point_21 305_-157 858(iPad Mini) (1)

localhost_8080_point_21 305_-157 858(iPad Mini) (2)

</details>

<details>
<summary>Desktop</summary>

localhost_8080_point_21 305_-157 858 (1)

localhost_8080_point_21 305_-157 858 (2)

</details>


web/themes/new_weather_theme/templates/partials/forecast-details-toggle.html.twig
```diff
@@ -0,0 +1,14 @@
+<div class="wx-forecast-details-toggle grid-row grid-col-12 margin-y-205" data-selected="table">
```
partly-igor, line 1: could you add `tablet:grid-col-6` here?

web/themes/new_weather_theme/assets/sass/components/forecast-details-toggle.scss
```diff
@@ -0,0 +1,13 @@
...
+    display: none;
+}
+
+.wx-forecast-details-toggle[data-selected="charts"] ~ *:not(wx-hourly-table) {
+    display: initial;
+}
+
+.wx-forecast-details-toggle[data-selected="table"] ~ wx-hourly-table {
+    display: initial;
+}
```
partly-igor, line 13: This works great for all but the QPF/precipitation amounts because the chart and table are combined in one include, so when you switch to "chart" it shows both the table and the chart and when you chose "table" it hides both. Probably need to pull chart into it's own partial.

partly-igor: Adding here that similar thing happens to the heading "Hourly forecast" because that one is inside the table partial, and I think we should probably pull it out into the `daily-forecast-list-item`

partly-igor (commented):
A couple of small issue, but looking good and very close.

web/themes/new_weather_theme/assets/js/components/forecast-details-toggle.js
```diff
@@ -0,0 +1,20 @@
...
+        buttons.forEach(btn => {
+          if(btn.dataset.selectionName === selectedName){
+            btn.classList.remove("usa-button--outline");
+          } else {
+            btn.classList.add("usa-button--outline");
+          }
+        });
+      });
+    });
+  });
```
greg-does-weather, line 20: This is very much a take-it-or-leave-it suggestion. It kind of evolved as I was thinking about it, so I'll lay out my thinking.

1. Attaching an event handler to a closure created inside this loop means we'll have a unique instance of the function for every button, or 14 instances of the function. It's not a lot, but it's still extra memory. This particular optimization is one that lives at the front of my brain because in Windows applications, it's a prime source of memory leaks. Not really an issue here, though, since the button's lifetime is the window's lifetime. 🤷‍♂️ 
2. One downside to moving to a single event handler is having to recompute the toggle and list of buttons each time. But it turns out `getElementsByTagName` and `getElementsByClassName` are at least one order of magnitude faster than `querySelector*`. TIL! So if these methods are sufficiently fast, then any performance hit is probably totally erased by not keeping the list of button nodes in memory forever, too, since they were being captured by the event handler closure.

But I don't know if this is actually significant enough to matter to anything.

```suggestion
// The click event for each toggle button will
// update the visual selection class and also update the
// selection data attribute
const toggleHandler = ({ target: button }) => {
  const toggleContainer = button?.closest(".wx-forecast-details-toggle");
  const selectedName = button?.dataset.selectionName;

  toggleContainer?.setAttribute("data-selected", selectedName);

  const buttons = Array.from(toggleContainer.getElementsByTagName("button"));
  buttons.forEach((btn) => {
    btn.classList.add("usa-button--outline");
    if (btn.dataset.selectionName === selectedName) {
      btn.classList.remove("usa-button--outline");
    }
  });
};

Array.from(document.getElementsByClassName("wx-forecast-details-toggle"))
  .forEach(toggleContainer => {
    const buttons = Array.from(toggleContainer.getElementsByTagName("button"));
    buttons.forEach(button => {
      button.addEventListener("click", toggleHandler);
    });
  });
```

web/themes/new_weather_theme/assets/sass/components/forecast-details-toggle.scss
greg-does-weather: Oops. That QPF chart and table being in the same template was my fault. 🤦‍♂️

greg-does-weather (approved):
Man, it took me a minute to understand how this works, but it's very slick. Nice use of CSS to simplify things! And today I learned about the `~` selector! Neat!

web/themes/new_weather_theme/assets/sass/components/forecast-details-toggle.scss
```diff
@@ -0,0 +1,22 @@
...
+    display: block;
+}
+
+.wx-forecast-details-toggle[data-selected="charts"] ~ *:not(wx-hourly-table) {
+    display: initial;
+}
+
+.wx-forecast-details-toggle[data-selected="charts"] ~ *:last-child {
+    display: none;
+}
```
greg-does-weather, line 17: So I understand, is this a workaround for the fact that the QPF table and chart are in the same container element? If so, it might be worthwhile to put a comment in here about it for future us's sake.

eric-gade: Yes it is, but this is all going to change anyway when we get to https://github.com/weather-gov/weather.gov/issues/1944 because we are going to have to put the different pieces in to different actual tabpanel elements

partly-igor (approved):
Looks good design wise!

Issue #1793, opened 2024-09-18
Update weather story component


colinmurphy01:
Placeholder Eng ticket for #1739 and blocked until then

Dependency - #1921 

This would require:
- Acting as DSS Builder, get the XML file from the specific office from prod ([example](https://www.weather.gov/source/lmk/wxstory/wxstory.xml))
- Image size compression is done for demo
- Priority for demo
  - Show placement on page
  - Show hierarchy of content within component
  - Show styling (BONUS, not required)
  
[Figma documentation](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13112-306091&node-type=frame&t=kqQsalojRQcyos73-11)
[Context/designs](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13113-313079&node-type=frame&t=kqQsalojRQcyos73-11)
[New icon for AFD](https://drive.google.com/file/d/1hCqhutIOZ-UrYubOhKqLMzSvepU8efxz/view?usp=drive_link)

colinmurphy01:
This is work that's been prioritized for 10/18.

colinmurphy01:
@colinmurphy01 to add detail for delivery by 10/18, likely without DSS builder integration completed.

colinmurphy01:
@eric-gade I think you were working on the desktop layout and changes @partly-igor had in [this slack thread](https://gsa-tts.slack.com/archives/C05DU11D4J2/p1729791044250799)

colinmurphy01:
Now that we've moved the AFD link within this component, we'd like for the AFD link to be present on the page even if there isn't a weather story.

greg-does-weather:
Closed by #2027, #2004, #2017, #2057

PR #1946, opened 2024-10-18
Jt/weather story integration

jamestranovich-noaa:
## What does this PR do? 🛠️

Addresses https://github.com/weather-gov/weather.gov/issues/1793 -- designing the weather story

- There's a mismatch between our WFO taxonomy and what DSS builder gives to us for the `Office` field -- I'm still working on this, but this is likely a fix for another PR. For now, we assume that all weather story uploads have the `field_office` set to the WFO code. 

- I also cannot seem to "publish" weather story uploads. I'm not sure why yet.

- No tests yet :grimacing: 

image

### DIFF
--- a/web/themes/new_weather_theme/templates/block/block--weathergov-weather-story-image.html.twig
+++ b/web/themes/new_weather_theme/templates/block/block--weathergov-weather-story-image.html.twig
@@ -1,0 +1,21 @@
+<div class="grid-container margin-top-2 padding-x-0 tablet:padding-x-2" wx-outer-radar-container>
+  <div class="grid-row tablet-lg:grid-col-8 tablet-lg:grid-offset-2 bg-accent-cool-lightest">
+      <h3 class="wx-visual-h2 text-normal text-primary-darker padding-x-2 tablet:padding-x-0 margin-top-3">{{ "From the forecaster" | t }}</h3>
+  </div>
+  <div class="grid-container padding-0 bg-accent-cool-lightest grid-col-12 tablet-lg:grid-offset-2 tablet-lg:grid-col-8">
+    <div class="grid-row grid-gap">
+      <h4>{{ content.title }}</h4>
+    </div>
+      <div class="grid-row grid-gap">
+        <p>Created: {{ content.updated.formatted }} by the {{ content.wfo }}</p>
+      </div>
+      <div class="grid-row grid-gap">
+        <p>{{ content.description }}</p>
+      </div>
+      <div class="grid-row grid-gap">
+        {% if content.image %}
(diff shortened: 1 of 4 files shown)


web/themes/new_weather_theme/templates/block/block--weathergov-weather-story-image.html.twig
```diff
@@ -0,0 +1,47 @@
...
+        <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+          <rect width="24" height="24" fill="transparent"/>
+          <path d="M6 5H18V6H6V5Z" fill="#0085CA"/>
+          <path d="M6 10H18V11H6V10Z" fill="#0085CA"/>
+          <path d="M18 15H6V16H18V15Z" fill="#0085CA"/>
+          <path d="M6 12H16V13H6V12Z" fill="#0085CA"/>
+          <path d="M16 17H6V18H16V17Z" fill="#0085CA"/>
+          <path d="M6 7H16V8H6V7Z" fill="#0085CA"/>
+          <path fill-rule="evenodd" clip-rule="evenodd" d="M4 2H20V22H4V2ZM5 3H19V21H5V3Z" fill="#0085CA"/>
+        </svg>
```
greg-does-weather, line 33: Just curious, but why is this SVG inline instead of an image tag?

greg-does-weather (approved):
LGTM. The tests shouldn't be too hard to add later since we already have the infrastructure for it, and they'll be similar to existing tests!

web/themes/new_weather_theme/templates/block/block--weathergov-weather-story-image.html.twig
jamestranovich-noaa: I pulled this svg from the figma design straight up, we don't have a pre-existing svg file. I looked at putting this in `./web/themes/new_weather_theme/assets/images/weather/` but it didn't seem to fit anywhere :shrug: what do you recommend we do here?

jamestranovich-noaa:
> When trying the upload script with [this example XML file](https://www.weather.gov/source/lmk/wxstory/wxstory.xml), I am getting the following error in the Python output:

Fixed. This is because the xml format for images is inconsistent, sometimes it provides relative paths and sometimes `https://`.

> What's even stranger is that the weather story is uploaded and will appear in my local system, with images, but the effective dates will be incorrect (and therefore it won't show on the point page):

The start and end times don't matter, we are grabbing all fields from weather stories regardless of whether we will use them or not. I checked the example XML you gave me and the epoch is `1722538860` which translates to, yep, `Thursday, August 1, 2024 7:01:00 PM` so ... :man_shrugging: I can ask the DSS builder team about this but it's not an error on our end as far as I can see.

PR #1971, opened 2024-10-23
config split: enable s3fs for uploads on cloud.gov

jamestranovich-noaa:
## What does this PR do? 🛠️

Enables `s3fs` for newer upload content types on cloud.gov. The config is copied from the pre-existing `config_split.patch.field.storage.node.field_image.yml` and _should_ be all we need here.

To test, `docker compose exec drupal drush config-split:activate cloudgov` (don't forget to deactivate afterwards!)

We will need to test uploads again on staging to make sure uploaded s3 files are persisted. Uploading the newer uploads to s3 is untested.


eric-gade (approved):
Ok I _think_ it's working locally here

PR #1984, opened 2024-10-25
Background alert processing

### LINKED ISSUE #1881: [Interop layer] Refactor alerts module to move top-level code into a function
### Description

The alerts module has top-level code that immediately loads alerts and sets up the timer to refresh them every 30 seconds. As a result of this top-level code, the alerts module is slightly more difficult to test. The top-level code should be moved into a function and that function should be called by the main script.

### Acceptance criteria
- [ ] Alerts initialization and updating are wrapped in a function
- [ ] The main script calls the init/check function
- [ ] Alert module tests are updated accordingly
- [ ] Module tests that dynamically import their modules because of the top-level alert function are modified to use static imports

### LINKED ISSUE #1976: [Interop layer] Use a point buffer for intersecting alerts instead of the WFO grid
### Description

Currently we return all alerts that intersect with the WFO grid for the user's location. This is okay, but it means people in different parts of the WFO grid will get the same alerts whether they apply or not. Instead, we should get alerts that intersect with a buffer around the user's location.

- the `ST_BUFFER` function requires that the distance be in the same units as the coordinate system. Our coordinate system is a geographic lat/lon, so we'll need to convert the user's location to SRID 3857 (web mercator), make the buffer, and convert back to 4269 (NAD83 – the geographic projection our other spatial tables are using) before doing the intersection.

### LINKED ISSUE #1977: Interop layer creates "holes" in alert polygons
### Description

The interop layer simplifies alert geometries to make them smaller going to users. However, because the geometries are actually geometry *collections*, the simplification process simplifies each individual geometry in the collection. The result is that the "overarching" geometry can have holes as the simplification causes inner geometries to pull away from each other. This can also result in users not being shown alerts that apply to them because they happen to have fallen into one of the holes.

#1975 would open up the possibility of progressively enhancing alert geometries. So in the first pass, we would not simplify them at all. Then at a later time, we could union the geometry collection into a single polygon and simplify that.

greg-does-weather:
## What does this PR do? 🛠️

- Moves alert updating and processing into a background thread
- Adds alert hashing to uniquely identify alerts
- Only processes alerts from the API if it has not processed them before
- Removes locally-stored alerts if they are no longer in the API data
- Moves some heavy spatial processing into code that is faster than MySQL

The code that gets a list of alerts for a location goes from about 1,000 milliseconds down to around 300 milliseconds (with around 30 active alerts), and should result in a quicker user experience.

---

- fixes #1977
- fixes #1976
- closes #1881

### DIFF
--- a/api-interop-layer/package.json
+++ b/api-interop-layer/package.json
@@ -11,6 +11,7 @@
   "author": "",
   "license": "CC0-1.0",
   "dependencies": {
+    "@turf/turf": "^7.1.0",
     "convert": "^5.3.0",
     "dayjs": "^1.11.12",
     "fastify": "^4.28.1",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,11 @@
 module.exports = {
   root: true,
   extends: ["airbnb-base", "prettier"],
-  ignorePatterns: ["tests/translations/**/*.js", "uswds*.js"],
+  ignorePatterns: [
+    "tests/translations/**/*.js",
+    "uswds*.js",
+    "web/sites/default/**",
+  ],
   rules: {
     // For imports in the browser, file extensions are always required.
     "import/extensions": ["error", "always"],
--- a/api-interop-layer/data/alerts/backgroundUpdateTask.js
+++ b/api-interop-layer/data/alerts/backgroundUpdateTask.js
@@ -1,0 +1,199 @@
+import { createHash } from "node:crypto";
+import { parentPort } from "node:worker_threads";
+import { createLogger } from "../../util/monitoring/index.js";
+import dayjs from "../../util/day.js";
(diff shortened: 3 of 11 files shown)


api-interop-layer/package.json
```diff
@@ -11,6 +11,7 @@
   "author": "",
   "license": "CC0-1.0",
   "dependencies": {
+    "@turf/turf": "^7.1.0",
```
jamestranovich-noaa, line 14: Can we add documentation as to why we chose this library? Maybe in `contributed-modules`?

greg-does-weather: Good call!

jamestranovich-noaa (approved):
tested and ran locally, seems to work well :tada:

Issue #1948, opened 2024-10-21
Revisit publishing workflow


jamestranovich-noaa:
### Description

The current custom publishing workflow we have is not generic enough to handle any content type, which leads to various issues, notably not being able to "publish" a weather story or PDF file manually. Currently the publishing workflow forces one to "preview" before actually publishing, but it seems somewhat buggy in that sometimes previewing does not enable the "publish" button for, well, publishing (this happens for any content type, not just uploads). But there might be a larger issue here -- at this point, if we have weather stories and pdf sit reps going through DSS builder, do we really need this publishing workflow anymore? 

### Acceptance criteria

- [ ] Decide whether or not to keep, modify, or remove the current publishing workflow
- [ ] Tests have been implemented or modified to capture the bug

colinmurphy01:
Is this blocking features like the sitrep briefing and weather stories we get from DSS Builder?

jamestranovich-noaa:
DSS builder uploads can and will continue to work as normal. _But_ this publishing issue is blocking us from adding, or indeed making any edits to DSS builder content.

Which is why I had to use the JSON:API endpoint to upload (modified) demo content to beta.weather.gov instead of making a new content type manually. Also, flagging that this functionality is just buggy :man_shrugging:

colinmurphy01:
Next steps:

- [ ] @kmranjo and @jamestranovich-noaa chat about the use case and options for fixing
- [ ] regroup with team for feedback
- [ ] implement change

jamestranovich-noaa:
Opened up a separate issue to discuss publishing workflow needs/wants, since that is uncertain at this moment: https://github.com/weather-gov/weather.gov/issues/2022

PR #2017, opened 2024-10-30
admin: remove weather story publishing workflow

jamestranovich-noaa:
## What does this PR do? 🛠️

Addresses https://github.com/weather-gov/weather.gov/issues/1948

We can still preview but we just don't enforce it as a requirement to publish anymore. Also, the older weather story content type is deprecated and we're not using it anymore. **Please** let me know if there's anything in here we should keep for the future.


greg-does-weather (approved):
yoink

PR #2018, opened 2024-10-30
Add county list to WFO pages

### LINKED ISSUE #1738: AFD - Accessible alternative to WCA map on WFO

greg-does-weather:
## What does this PR do? 🛠️

1. Updates the `weathergov_geo_counties` table to add the list of CWAs from the shapefile to our database
2. Updates the `new_weather_theme` variable-builder for the WFO Info content type to fetch the counties and include it in the data available to twig
3. Updates the WFO Info node template to add the list of counties in an sr-only paragraph inside the figure caption.

- closes #1738


greg-does-weather:
We'll need to run spatial update on all our environments before merging this one. It should be relatively quick at least, since it's only the counties table.

In future, it might be nice to add versioning to the data the same way we version the schema, so data updates don't necessarily have to be truncate-and-repopulate every time. But that's another time.

jamestranovich-noaa (approved):

jamestranovich-noaa:
> In future, it might be nice to add versioning to the data the same way we version the schema, so data updates don't necessarily have to be truncate-and-repopulate every time. But that's another time.

We can use a separate database migration tool such as `[dbmate](https://github.com/amacneil/dbmate)` -- we've had good success with this tool with other projects.

PR #2025, opened 2024-10-30
fix up the proxy thingy

### LINKED ISSUE #1613: [Interop layer] Add request ID so the API proxy can correctly record bundle data
### Description

Requests to the API proxy should include a `wx-gov-response-id` header that identifies the single user request being served by the various API calls. This allows the proxy to group the API responses into a single bundle. This was previously handled in the DataLayer, but the interop layer now needs to do it. 

### Acceptance criteria

- [ ] Outgoing requests from the interop layer have a `wx-gov-response-id` header
- [ ] Tests have been implemented to check that the header exists

greg-does-weather:
## What does this PR do? 🛠️

- Instead of manually building up a UI for the proxy by pushing lines of HTML to an array and then dumping them into the response, this PR adds mustache and a template. It's a lot easier to read!
- While I was doing that, I styled the thing up to be nicer to look at.
- I added a button for recording a new bundle. This button is attached to a text box. In that text box, you'll put a URL to a point location page. Behind the scenes, the proxy will parse that URL to find the lat/lon for the location page, and then it'll bundle that up. The new bundle's path will be the current Javascript timestamp (Unix epoch, in milliseconds)
- I also added a button to add a single point to an existing bundle. The main difference between the two (other than where on disk they store their files) is that adding a point to an existing bundle ***DOES NOT*** capture alerts. The reason for this is that there's only a single alerts file in a bundle now.
  - If you need to capture alerts, create a new bundle.
  - If you want the new alerts to be in an existing bundle, you'll need to manually copy them over. In that case, the best thing is probably to first add the point to the existing bundle, and then create a new bundle from that point. That way all of the point-specific data is in the existing bundle and you only have to migrate the alerts.
- For both kinds of recordings, this PR also records the observations for the first three observation stations since the interop layer optimistically fetches those. This way we'll be sure to also serve those from bundles.

- closes #1613, by a different route. There's no more "response ID" because the proxy handles all of the bundling itself instead of relying on other parts of the system to tell it what things go together. It ✨ just knows ✨

**Requires rebuilding your api-proxy container to install the new npm dependencies.**

image

### DIFF
--- a/docs/dev/intercepting-the-api.md
+++ b/docs/dev/intercepting-the-api.md
@@ -5,19 +5,71 @@ static set of well-defined data can be easier to work with in those cases. It
 can also allow us to capture data that causes bugs so we can add tests for them
 and fix them more easily without hoping the weather stays the same.
 
+Our proxy server runs in dev environments at
+[http://localhost:8081](http://localhost:8081).
+
 By default, our dev and CI/CD environments use our proxy server. The proxy is a
-small Node.js app using Express listening on port 8081. Our dev Drupal instance
-is configured to use this endpoint instead of `api.weather.gov`. When the proxy
-receives a request, it will first see if there is an existing JSON file that
-corresponds with the request. If there is, it will simply respond with that file
-and be done. Otherwise, it will forward the request on to the API and return
-whatever it gets back.
+small Node.js app using Express listening on port 8081. Both Drupal and the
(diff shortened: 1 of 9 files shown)


jamestranovich-noaa (approved):
This really is way too reasonable. Can we please have dev documentation, though? Basically copying and pasting the PR comments somewhere else for future reference would help a lot :pray:

Issue #1830, opened 2024-09-27
AFD Section Parse Errors


eric-gade:
### Description
This issue is actually two problems in one:
1. Whenever something unexpected comes back in parsed AFD nodes -- specifically when it comes to the temps/pops logic -- a twig failure breaks the whole site;
2. A specific case (bundled below) of questionably formatted AFD sections breaks twig assumptions

Do this at same time as #1796 

### Additional detail
You can see the formatting problem in this part of the captured AFD:
```
.PRELIMINARY POINT TEMPS/POPS...
TPA  89  79  88  78 /  10  10  50  40 
FMY  90  80  90  79 /  50  60  80  40 
GIF  92  77  91  77 /  20  20  70  30 
SRQ  90  79  89  78 /  20  30  60  50 
BKV  89  74  90  74 /  10  10  40  30 
SPG  90  82  89  81 /  10  20  50  50 

&&

Sea Breeze Thunderstorm Regime For Friday: 5
Sea Breeze Thunderstorm Regime For Saturday: 5

For additional information on sea breeze regimes, go to: 
     https://www.weather.gov/tbw/ThunderstormClimatology


&&
```

Either our reading of the directive is incorrect, or this is not properly formatted.
### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] The example bundle renders correctly
- [ ] Twig render errors for some subsection of the page do not break the whole site (could be kicked to a separate, more general issue if need be)
[bundle_66f6ccf90d614.zip](https://github.com/user-attachments/files/17166290/bundle_66f6ccf90d614.zip)

colinmurphy01:
This reads like a critical bug we should fix this sprint?

CC @greg-does-weather

jamestranovich-noaa:
side note, we have NR logging already setup for the api interop branch so parsing breakage should be easier to track going forward

mgwalker:
I dunno. Since the site isn't operational yet, I think I'd prefer making this fix only in the interop layer. It seems silly to me to fix code that we hope to retire in a couple of weeks.

colinmurphy01:
Sounds like a plan. Will that be included in #1796?

greg-does-weather:
That seems right! I'll note it over there too so we don't forget.

colinmurphy01:
Refinement: 
- Fix the breaking issue with the twig templates
- Update parsing to show text that's not under headers

colinmurphy01:
[...]

Issue #1796, opened 2024-09-18
[interop layer] Add AFD parsing and tests to interop


eric-gade:
### Description
We need to rewrite the AFD parser in JS and add to the interop layer. Additionally, we should transliterate the corresponding parser tests.

### Acceptance criteria

- [ ] AFD parsing is handled by the interop layer
  - [ ] the bug identified in #1830 is addressed
- [ ] Tests within the interop layer capture the same capabilities as the current tests
- [ ] PHP-side templates / classes are updated to use parsed markup from the interop, rather than creating markup themselves

colinmurphy01:
@eric-gade we're temporarily assigning this you for this sprint, we can regroup when you're back.

colinmurphy01:
@eric-gade assigning this to you for this sprint, let us know if you have questions or concerns.

greg-does-weather:
Closed by #2030

PR #2030, opened 2024-11-04
AFD parsing to interop

eric-gade:
## What does this PR do? 🛠️
This PR implement #1796 (moving AFD parsing to the interop layer), while also resolving and accounting for the bug in #1830.

This only moves the fetching of weather products to the interop layer, and specifically it only handles modifications (ie parsing the productText and attaching a parsed node array to the response object) to AFD products. If we use other products on the site in the future, we will need to update both the PHP side and the interop side to account for this, since the API's route for products does not distinguish the productType in the url (ie, it's usually just `/product/<uuid>` rather than, say, `/products/AFD/<uuid>`)


greg-does-weather (approved):
Wowza!
approved

PR #2040, opened 2024-11-04
Switch database library to mysql2

greg-does-weather:
## What does this PR do? 🛠️

The mariadb library seems to have been responsible for an unexpectedly large amount of our total app memory. This PR switches us to the mysql2 library, which is API-compatible, so an easy swap. This appears to fix the memory problems in my environment.

### DIFF
--- a/api-interop-layer/Dockerfile
+++ b/api-interop-layer/Dockerfile
@@ -8,4 +8,4 @@ ADD ./package-lock.json .
 
 RUN npm ci
 
-CMD npm start
+CMD npm run start-dev
--- a/api-interop-layer/data/alerts/backgroundUpdateTask.js
+++ b/api-interop-layer/data/alerts/backgroundUpdateTask.js
@@ -17,7 +17,7 @@ export const updateAlerts = async ({ parent = parentPort } = {}) => {
   parent.postMessage({
     action: "log",
     level: "verbose",
-    message: "updating alerts",
+    message: `updating alerts with ${KNOWN_ALERTS.size} known alerts`,
   });
 
   // The list of alert hashes in the current results from the API. We'll use
@@ -103,8 +103,6 @@ export const updateAlerts = async ({ parent = parentPort } = {}) => {
   });
 
   for await (const rawAlert of alertsToUpdate) {
-    KNOWN_ALERTS.add(rawAlert.properties.hash);
-
     const alert = {
       metadata: alertKinds.get(rawAlert.properties.event.toLowerCase()),
     };
@@ -136,6 +134,7 @@ export const updateAlerts = async ({ parent = parentPort } = {}) => {
         message: `Ignoring "${rawAlert.properties.event}" - not a land alert`,
       });
 
+      KNOWN_ALERTS.add(rawAlert.properties.hash);
       continue; // eslint-disable-line no-continue
     }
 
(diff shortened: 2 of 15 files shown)


api-interop-layer/Dockerfile
```diff
@@ -8,4 +8,4 @@ ADD ./package-lock.json .
 
 RUN npm ci
 
-CMD npm start
\ No newline at end of file
+CMD npm run start-dev
```
greg-does-weather, line 11: This pulls nodemon out of our cloud environments

api-interop-layer/data/alerts/backgroundUpdateTask.js
```diff
@@ -146,14 +145,23 @@ export const updateAlerts = async ({ parent = parentPort } = {}) => {
 
     alert.event = rawAlert.properties.event;
 
-    const { locations, description } = parseLocations(
-      rawAlert.properties.description,
-    );
+    try {
```
greg-does-weather, line 148: Twice today while I've been debugging there have been alerts that threw exceptions when trying to parse their locations. I wasn't able to capture either case, but now that particular bit of logic is wrapped in a try/catch and it logs the alert ID, which we should be able to lookup later from [IA State](https://mesonet.agron.iastate.edu/GIS/model.phtml)

api-interop-layer/data/alerts/backgroundUpdateTask.js
```diff
@@ -214,9 +225,12 @@ export const start = () => {
...
 
 if (parentPort) {
-  parentPort.on("message", (message) => {
-    switch (message.toLowerCase()) {
+  parentPort.on("message", ({ action, data }) => {
+    switch (action.toLowerCase()) {
       case "start":
+        data.forEach((hash) => {
+          KNOWN_ALERTS.add(hash);
+        });
```
greg-does-weather, line 233: When the background process starts up, it can now take a list of known alert hashes as seed. This is useful for cases where the background process needs to be restarted. The main process can send it the list of hashes it knows about and then the background process can start off skipping those.

api-interop-layer/data/alerts/index.js
```diff
@@ -69,13 +74,38 @@ export const startAlertProcessing = () => {
...
+        startAlertProcessing();
+      }, 5_000);
+    };
+
+    // If our background thread stops, restart it.
+    updater.on("exit", restart);
+    updater.on("error", (e) => {
+      backgroundLogger.error(e);
+      restart();
+    });
```
greg-does-weather, line 102: If the background worker exits or errors, we should note that and restart it.

docker-compose.yml
```diff
@@ -29,6 +29,11 @@ services:
     build:
       context: ./api-interop-layer
       dockerfile: Dockerfile
+    # Use the same resource limits as deployed
```
jamestranovich-noaa, line 32: :star_struck:

jamestranovich-noaa (approved):

eric-gade (approved):
Excellent!

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

api-interop-layer/data/db.js
```diff
@@ -1,4 +1,4 @@
-import database from "mysql2/promise";
+import database from "mysql2/promise.js";
```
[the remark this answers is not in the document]
greg-does-weather: This is definitely in our eslint config, but also it looks like Node requires it? https://nodejs.org/api/esm.html#mandatory-file-extensions - though looking at the [mysql2 `package.json`](https://github.com/sidorares/node-mysql2/blob/master/package.json#L52), `promise` is indeed a named export so yeah, this should just work. It's just the linter that's complaining.

PR #2056, opened 2024-11-07
make dump-spatial: fix access denied error

jamestranovich-noaa:
## What does this PR do? 🛠️

I had to zap my environment and got the following error when attempting to dump the spatial db:

```
# mysqldump: Error: 'Access denied; you need (at least one of) the PROCESS privilege(s) for this operation' when trying to dump tablespaces
```

Not sure why I'm getting this now.


greg-does-weather (approved):
This error has been there all along and it was way way back in the back of my brain to figure out how to suppress it at some point. It hadn't been a blocker so I just never could prioritize it. Thank you THANK YOU for fixing it!

PR #2061, opened 2024-11-07
When browsing to a place URL, don't redirect to a point URL

greg-does-weather:
## What does this PR do? 🛠️

Currently when you browse to a place page, such as https://beta.weather.gov/place/ca/oxnard, we find an appropriate lat/lon and redirect you to a point page: https://beta.weather.gov/point/34.1975/-119.17705

This is not ideal either for SEO or usability. This PR modifies that behavior.

1. If the place name has a space in it, we redirect to the same URL but with the space replaced with an underscore
2. If the state or place name are not capitalized the same as our database, we redirect to the same URL but with the capitalization normalized
3. If the state and place name match our database, we simply render the page instead of redirecting.

The result is that we should end up with mostly canonical URLs for all the places we know about. For example:

```
https://beta.weather.gov/places/ca/los angeles
```
would become
```
https://beta.weather.gov/places/CA/Los_Angeles
```

A natural successor and completion of #1514.

***NOTE*** that this screenshot is a little outdated because the PR was updated to redirect from the "ca/los angeles" place to "CA/Los_Angeles". But it still shows the gist of what's going on.

image

### DIFF
--- a/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
+++ b/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
@@ -129,11 +129,11 @@ final class LocationAndGridRouteController extends ControllerBase
             );
 
             if ($location !== false) {
-                $url = Url::fromRoute("weather_routes.point", [
-                    "lat" => $location->lat,
-                    "lon" => $location->lon,
-                ]);
-                return new RedirectResponse($url->toString());
+                // If we get a location, continue with the rendering process.
+                // If anyone figures out how to pass the lat/lon from here on
+                // through the process, that'd be dandy – then the template
+                // variable preprocessor thingy wouldn't have to redo this query
+                return [];
             }
         } catch (\Throwable $e) {
             throw new NotFoundHttpException();
--- a/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
+++ b/web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
@@ -113,28 +113,80 @@ final class LocationAndGridRouteController extends ControllerBase
         }
     }
 
-    public function redirectFromPlace($state, $place)
+    private function getKnownPlace($state, $place)
+    {
+        // We escape spaces with underscores and slashes with commas. We need
+        // to unescape them before querying. (None of the place names in our
+        // database have underscores or commas at the time of this writing.)
+        $place = str_replace("_", " ", $place);
+        $place = str_replace(",", "/", $place);
+
+        return $this->dataLayer->databaseFetch(
+            "SELECT
+                ST_X(point) AS lon,
+                ST_Y(point) AS lat,
+                state,
+                name
+            FROM weathergov_geo_places
+            WHERE
+                state LIKE :state
+                AND
+                name LIKE :place",
+            [":state" => $state, ":place" => $place],
+        );
+    }
+
+    public function servePlacePage($state, $place)
     {
         try {
-            $location = $this->dataLayer->databaseFetch(
-                "SELECT
-                    ST_X(point) AS lon,
-                    ST_Y(point) AS lat
-                FROM weathergov_geo_places
-                WHERE
-                    state LIKE :state
-                    AND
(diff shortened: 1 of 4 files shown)


jskinne3 (changes requested):
It appears that URLs may contain spaces, for example: `/ca/los angeles`

I believe there are circumstances under which spaces will be escaped to `%20` resulting in the less-readable URL `/ca/los%20angeles`

I suggest replacing spaces with underscores -- provided that there are no, or few, valid place names with literal underscores in them

web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
```diff
@@ -113,28 +113,80 @@ public function redirectFromGrid($wfo, $gridX, $gridY)
...
+                // (There are two places in American Samoa with slashes in
+                // their names, but slashes are URL path characters so we
+                // don't want to persist them!)
+                $url = Url::fromRoute("weather_routes.place", [
+                    "state" => $known->state,
+                    "place" => str_replace(
+                        "/",
+                        ",",
+                        str_replace(" ", "_", $known->name),
+                    ),
```
eric-gade, line 179: Nice, I didn't know about this kind of thing before

web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
```diff
@@ -113,28 +113,80 @@ public function redirectFromGrid($wfo, $gridX, $gridY)
...
                 ]);
                 return new RedirectResponse($url->toString());
             }
+
+            // If we're here, we know the location and the requested URL is
+            // already canonical, so continue with the rendering process.
+            // If anyone figures out how to pass the lat/lon from here on
+            // through the process, that'd be dandy – then the template
+            // variable preprocessor thingy wouldn't have to redo this query
```
eric-gade, line 188: Thinking off the top of my head: it might be possible now to do away with the preprocess hook altogether for the point page route, and instead use the "theme object" structure to define what kinds of data the template for the point page expects (it has a confusing name, it's more like a schema for the expected assoc array a given template or node would want to render). You can see [these examples for the AFD page](https://github.com/weather-gov/weather.gov/blob/main/web/modules/weather_routes/weather_routes.module#L5) where that happens. I believe all AFD data fetching occurs now in the route handler. We could consider doing the same for our location pages.
  
If you think that makes sense, I'll write up a ticket for the work and take it from there.

eric-gade (approved):
This is great!
  
I've gone ahead and created https://github.com/weather-gov/weather.gov/issues/2077 just to add a few browser tests for these cases.
  
I also left an inline comment about possibly abandoning the preprocess hook and just fetching all data from within our route handlers using "theme" objects. If you agree, I can make a follow up ticket for that.

Thanks for this final hurrah!

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
greg-does-weather: That seems dope. That variable prerender hook thing is so ugly, and it is completely undiscoverable. Future devs will be thankful if the code is in a sensible place like a route handler, where it totally makes sense.

greg-does-weather:
Do I merge this, or does that baton pass on to someone else now? 😇

eric-gade:
@greg-does-weather Let it be yours, for one last time...

web/modules/weather_routes/src/Controller/LocationAndGridRouteController.php
eric-gade: I've documented in a separate ticket here https://github.com/weather-gov/weather.gov/issues/2078

PR #2064, opened 2024-11-07
Spatial data management tweaks action

greg-does-weather:


greg-does-weather:
The approaches from this PR won't work. I was trying to just reuse our existing script, but it depends on the Docker runner being the same machine that the cf SSH tunnel is listening on. That's fine in our local dev environments, but it breaks in GitHub where the Docker runner is a remote host.

I think a workable solution would be to create another container that hosts the cf cli and that container would establish the SSH tunnel to cloud.gov. Running that container on the same virtual network as the spatial container should solve the problem of them being able to communicate. And I think this approach is more robust anyway, making fewer assumptions about the systems external to Docker.

Issue #2060, opened 2024-11-07
Map weather story images to the right WFO


jamestranovich-noaa:
### Description

The weather stories need to be mapped to the appropriate WFO code in order to show up on the daily forecast page. Unfortunately the `Office` field in the XML seems to be inconsistent and even varies over time. Example: the weather story XML documents for `TAE` sometimes lists the `Office` field as `Tallahassee, FL` -- which makes sense -- and at other times as the generic `National Weather Service` -- which, I suppose, is correct but not accurate at all. Some `Office` fields are prefixed with "NWS Forecast Office", such as `NWS Forecast Office Medford, OR`, but some are not (as in `Old Hickory, TN`). In other words, there's a great deal of variance here in the XML data.

For a first pass, let's see how much we can disambiguate based on the `Office` field. Or maybe form a heuristic based on other fields: if the `Office` does not match anything (e.g., the generic "National Weather Service"), the `cwa_center_*` latitude and longitude fields might be workable. 

### Acceptance criteria

- [ ] We have a way to reliably map the XML attributes to the right WFO
  - [ ] _Or_, if not, we request DSS Builder to make changes so that we know how to associate the weather story image with the right WFO
- [ ] We have a test for mapping XML attributes to WFO codes

colinmurphy01:
Does Jeremy know of this issue? This feels like something they should fix on the DSS Builder side, no?

jamestranovich-noaa:
For now, here is the best heuristic I've come up with:

- regexp match against `Office`: `^(NWS Forecast Office )?([^,]+)` and look it up via our taxonomy
- if that fails, regexp match against `WebURL`: `^http://www.weather.gov/([a-zA-Z]+)/?$`
- if that fails, regexp match against `WebURL`: `^/([a-zA-Z]+)/$`
- if that fails, regexp match against `WebURL`: `^/FXC/wxstory.php\?wfo=([a-zA-Z]+)$`
- if that fails, give up.

This gives me ~92.14% error rate, which is the best I can do without making the heuristic more complicated. I'll go ahead and implement this for matching while we wait for a reply or better solution (if any) from the DSS builder team.

We also might need to modify our taxonomy a bit: currently, "Charleston" and "Jackson" have the same name and we need to disambiguate. So there's a bit of taxonomy work to do as well.

colinmurphy01:
@coreypieper @shadkeene are you okay with this success rate for now?

[...]

PR #2082, opened 2024-11-15
derive WFO for weather story uploads

jamestranovich-noaa:
## What does this PR do? 🛠️

For weather story uploads, we attempt to derive the WFO based on, primarily, the `WebURL` XML attribute -- this gives us a ~92% success rate. **This is not a permanent solution.** The proper fix/approach needs to come from the DSS builder team, and this is a stopgap measure until we figure out the proper way to correlate weather story uploads and WFOs.

Note: I tried doing a computed property in Drupal but found it overly complicated and brittle with a lot of boilerplate code besides. So, I ended up going with a `presave` hook approach.

- Adds a new `field_derived_wfo` that can be overridden manually (that is, if the derived WFO is incorrect or unknown)
- `presave` hook will examine the weather story upload and attempt to derive the WFO
  - if the derived field is already set, we bail out. 
  - if the web url field is not set, we bail out. 
  - there are roughly five regexp matches we can attempt to derive the wfo from (all are covered with outside tests)
  - if we cannot determine the WFO then we set it to `unknown`

Closes https://github.com/weather-gov/weather.gov/issues/2060

Please run outside tests manually with this PR -- they do not run as part of the CI (yet).

### DIFF
--- a/tests/playwright/outside/api.spec.js
+++ b/tests/playwright/outside/api.spec.js
@@ -157,6 +157,11 @@ describe("API tests", () => {
     const secondJson = await secondResponse.json();
     expect(secondJson).not.toHaveProperty("errors");
     expect(secondJson).toHaveProperty("data");
+    // verify that `weather_cms_entity_presave` could not derive the WFO due to
+    // insufficient information.
+    expect(secondJson.data).toHaveProperty('attributes');
+    expect(secondJson.data.attributes).toHaveProperty('field_derived_wfo');
+    expect(secondJson.data.attributes.field_derived_wfo).toEqual('unknown');
   });
 
   test("uploader cannot upload wfo weather story images with missing attributes", async ({ request }) => {
@@ -224,7 +229,7 @@ describe("API tests", () => {
           title: pngFilename,
           field_office: "WFO test office",
           field_description: "a blank uploaded image",
-          field_weburl: "/test/wxstory.php?wfo=test",
+          field_weburl: "/FXC/wxstory.php?wfo=test",
           field_frontpage: true,
           field_graphicnumber: 1,
           field_order: 1,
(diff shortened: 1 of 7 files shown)


tests/playwright/outside/api.spec.js
```diff
@@ -157,6 +157,11 @@ describe("API tests", () => {
     const secondJson = await secondResponse.json();
     expect(secondJson).not.toHaveProperty("errors");
     expect(secondJson).toHaveProperty("data");
+    // verify that `weather_cms_entity_presave` could not derive the WFO due to
+    // insufficient information.
+    expect(secondJson.data).toHaveProperty('attributes');
+    expect(secondJson.data.attributes).toHaveProperty('field_derived_wfo');
+    expect(secondJson.data.attributes.field_derived_wfo).toEqual('unknown');
```
eric-gade, line 164: This is great, thanks

eric-gade (approved):
Good work on this -- I know it's a brutal (temporary) solution 👍 

One non-blocking question: do we have a way in the CMS UI to manually map the "unknown WFO" weather stories to their proper WFO?

## Code Review Checklist

## The Basics
- [x] Checks are passing
- [x] I read the code
- [x] I ran the code

## Documentation

## Security

## Reliability

## Infrastructure

## Accessibility
- [x] New pages have been added to cypress-axe file so that they will be tested with or automated accessibility testing
    - [ ] Tab order is logical

## Device Matrix
    - [ ] at 480px     (mobile)
    - [ ] at 640px     (tablet)
    - [ ] at 1024px    (desktop)

jamestranovich-noaa:
> One non-blocking question: do we have a way in the CMS UI to manually map the "unknown WFO" weather stories to their proper WFO?

Yes! You can manually edit the `WFO derived` field in the CMS and save. This is is partly why we don't override if we see that the derived field is already set.

Issue #2021, opened 2024-10-30
remove all references to the older weather story content type


jamestranovich-noaa:
### Description

The older weather stories are, at this point, vestigial. We want to remove all references to these:

- [ ] weather story content type definitions
- [ ] weather story twig templates
- [ ] weather story tests (or repurpose for weather story uploads)
- [ ] exported weather story ymls (from beta.weather.gov)
- [ ] delete all current references on beta.weather.gov so it's easier to migrate/update

This can happen either before or after the DSS builder integration happens for weather stories (TBD).

### Acceptance criteria

- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

PR #2083, opened 2024-11-15
remove old weather story content type

jamestranovich-noaa:
## What does this PR do? 🛠️

Closes https://github.com/weather-gov/weather.gov/issues/2021

Let me know if there might be any stray references lying around, please. I'm not confident I caught everything.

On beta.weather.gov we have to remove the older weather stories manually. There's about 14 of them. But we don't have these weather stories in our `scs-export` directory. So I've exported them just in case.


eric-gade (approved):
I think this is fine for now, and we can delete any references we find elsewhere as we come across them (a cursory search on my end didn't come up with anything)

PR #1938, opened 2024-10-17
Quick forecast nav

eric-gade:
## What does this PR do? 🛠️
Implements the "quick forecast" navigation outlined in #1925

## What does the reviewer need to know? 🤔
For the moment, we are doing most of the styling inside of a component SASS file (rather than with lots of utility classes). We might want to reconsider this in the future.
  
Also note we have kicked the can on a11y issues here, and those should be explored in #1937

## Screenshots (if appropriate): 📸

<details>
<summary>Mobile</summary>

localhost_8080_point_21 305_-157 858(iPhone SE)

</details>

<details>
<summary>Tablet</summary>

localhost_8080_point_21 305_-157 858(iPad Mini)

</details>

<details>
<summary>Desktop</summary>

localhost_8080_point_21 305_-157 858

</details>

PR #2095, opened 2024-11-19
Eg 1967 alernative approach

### LINKED ISSUE #1967: Finish UX Changes (desktop tabs, accordions)
Our plan is to implement the quick forecast and updates to the day forecast item incrementally – this ticket is for the second version of these components. The new version will change the quick forecast on mobile and table to act as an accordion, and as a set of tabs (or buttons) on tablet-lg and desktop. 

Mobile
[Figma documentation](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13538-178393&node-type=frame&t=11ZH72DCjsHUttWh-11)
[Context/Design](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13538-175875&node-type=frame&t=11ZH72DCjsHUttWh-11)

Tablet-lg/Desktop
[Documentation ](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13538-174248&node-type=frame&t=11ZH72DCjsHUttWh-11)
[Context/Design](https://www.figma.com/design/w2QI2uTgaTWUPLHLi40mm1/Weather.gov-%E2%80%93-v0.2?node-id=13538-174279&node-type=frame&t=11ZH72DCjsHUttWh-11)

eric-gade:
**_Choo-choooo!_ long train of thought / PR description ahead!**

## What does this PR do? 🛠️
This PR finalizes the UX design changes made after the most recent research round, summed up in #1967 
  
Due to the complexity of the design, as well as how we split this work up into two phases, I am including an extensive PR description below.

### Refresher: what we had before this ###
In order to get this large chunk of work out the door, we split the task up into two pieces. The first was the "interim" approach of creating a "quick forecast" component, and having it gracefully cascade from horizontal to vertical form as the browser shrinks.
  
The main functionality was as a nav, whose links navigated in-document to the corresponding daily forecast item. This work was done in #1938 
  
### What we want now: two modes ###
This PR finishes the work, which really splits the whole Daily tab into two "modes": tab mode and accordion mode.
  
#### "Tab" mode ####
In tab mode, the quick-forecast nav component is displayed. Instead of simply linking to the days within the page, however, each element is a "tab" and each matching daily forecast item is its corresponding "tabpanel". Clicking a given tab activates (displays) the matching daily forecast item, turning the whole section into a tabbed view.
  
This mode follows the [WCAG Tabs Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).
  
This is the mode we expect at the desktop width and higher.
  
It looks like this:
  
Screenshot 2024-11-20 at 3 07 25 PM
  
#### "Accordion" mode ####
In accordion mode, the quick-forecast nav is hidden entirely. Instead, each daily forecast item is preceded by a special button acting as an accordion controller for the rest of the forecast item content. Note that these accordion buttons are hidden at the desktop breakpoint.
  
[This mode follows the WCAG Accordion Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).
  
This is the mode we expect for everything _below_ desktop width.
  
It looks like this:
  
Screenshot 2024-11-20 at 3 11 44 PM

#### `<wx-daily-forecast>` component ####
We introduce a new web component, which replaces the `div` that served as the "Daily" tabpanel. This component knows how to switch the whole "Daily" tab between tab-mode and accordion-mode.
  
In short, switching between these modes visually is _mostly_ handled by CSS. But in order to make tab-mode work in a WCAG compliant way, we add `role` and `aria-` attributes to the daily forecast items that, when switched to accordion-mode, need to be removed. That can only be accomplished in JS, hence the primary use of this custom component.
  
#### Hey, how can we get JS to respond to our USWDS breakpoints anyway? ####
JS is not the normal place to listen for media query changes. But in order to respond to the changes between tab-mode and accordion-mode -- and to adjust the a11y attributes -- we need to respond to those changes in javascript.
  
We found a seemingly compliant way to do this using CSS Custom Properties and the window's built-in [matchMedia capability](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia). These will be noted via inline comments in the PR below.
  
Please let me know if you see any problems with this approach, as a lot of the functionality here hinges on it.

## What about tests?
We will implement the tests for this work in #2100, which will be the top priority ticket.
  
## Concerns
We are starting to approach peak complexity between our JS components, CSS layouts, and use of utility classes in the markup. We might need to consider a general refactoring, lightweight frameworks, design changes, or some combination of these. This is just an early warning.
  
## Closes
#1967


web/themes/new_weather_theme/assets/js/components/DailyForecast.js
```diff
@@ -0,0 +1,168 @@
...
+   * immediately, so as to setup the component for
+   * the given media (desktop or everything else)
+   */
+  setupMediaEvents(){
+    const computedStyle = getComputedStyle(this);
+    this.desktopBreakpoint = computedStyle.getPropertyValue("--wx-media-breakpoint-desktop");
+    this.desktopQuery = window.matchMedia(`(min-width: ${this.desktopBreakpoint})`);
+    this.desktopQuery.addEventListener("change", this.handleDesktopMediaChange);
+    this.handleDesktopMediaChange();
+  }
```
eric-gade, line 56: The actual measurements for the `desktop` breakpoint in our USDWS setup get pulled from the CSS custom property here. This is also where we configure the `matchMedia` object and set up a function to react to changes on it. That ends up being a listener for when the desktop breakpoint changes.

web/themes/new_weather_theme/assets/sass/components/daily.scss
```diff
@@ -53,3 +55,24 @@ table tr:nth-child(2) th {
...
+nav and the sibling daily forecast ol list when both
+items have the appropriate ARIA tab related roles
+set.
+Handles the hiding/showing of tab content based
+on the selected tab.
+*/
+wx-daily-forecast {
+    --wx-media-breakpoint-mobile: #{units("mobile")};
+    --wx-media-breakpoint-desktop: #{units("desktop")};
+}
```
eric-gade, line 70: This is where we set the CSS custom properties that give us the measurements for the desktop (and mobile -- for now unused) breakpoints.
  
These properties are read in by the DailyForecast web component.

jamestranovich-noaa:
Did we accidentally break wind direction? (May not be the fault of this PR)

image

jamestranovich-noaa (changes requested):
suggestion: 

```patch
diff --git a/web/themes/new_weather_theme/new_weather_theme.libraries.yml b/web/themes/new_weather_theme/new_weather_theme.libraries.yml
index b9974a7d..5c4719c4 100644
--- a/web/themes/new_weather_theme/new_weather_theme.libraries.yml
+++ b/web/themes/new_weather_theme/new_weather_theme.libraries.yml
@@ -25,7 +25,7 @@ front-page:
         defer: true
 
 point-page:
-  version: 2024-10-22
+  version: 2024-11-20
   js:
     assets/js/point.page.js:
       attributes:
```

jamestranovich-noaa:
Very nice PR! Just needs that one javascript version bump for everything to work and then I can approve :trophy:

web/themes/new_weather_theme/templates/point/forecast.html.twig
```diff
@@ -11,18 +11,42 @@
     </div>
   {% else %}
     <div class="grid-container padding-x-0 tablet:padding-x-2">
-      <div class="grid-col-12">
-        {% include '@new_weather_theme/partials/quick-forecast.html.twig' with { days: forecast.days } %}
+      {% include '@new_weather_theme/partials/quick-forecast.html.twig' with { days: forecast.days } %}
+      <div class="wx-quick-forecast-header desktop:display-none" aria-hidden="true">
+        <div class="wx-quick-forecast-header-top margin-right-3 table:margin-right-0">
```
partly-igor, line 16: Header can be a touch smaller `font-mono-xs`

web/themes/new_weather_theme/templates/partials/daily-forecast-quick-toggle.html.twig
```diff
@@ -0,0 +1,84 @@
...
+{% endif %}
+<button
+  class="wx-quick-toggle-item bg-white flex-align-center display-flex flex-wrap width-full padding-105"
+  id="{{dayId}}-quick-forcast-button"
+  role="button"
+  aria-controls="daily-content-{{itemId}}"
+  aria-expanded="true"
+>
+  <div class="wx-quick-toggle-item-top display-flex flex-1 flex-justify flex-align-center text-primary-darker font-mono-sm">
+    <time class="text-align-left flex-1 text-primary-dark" datetime="{{day.periods[0].startTime}}" aria-label="{{"Date" | t}}">
```
partly-igor, line 24: I think the aria-label here may need to be the date value instead of "Date" because otherwise the SR repeats "Date" for every item in the quick forecast, and doesn't announce the actual value

Screenshot of voice over text

web/themes/new_weather_theme/templates/partials/daily-forecast-quick-toggle.html.twig
```diff
@@ -0,0 +1,84 @@
...
+      <span>{{day.periods[0].dayName}}</span>
+      <span class="display-block">
+        {{day.monthNumericString}}/{{day.dayNumericString}}
+      </span>
+    </time>
+
+    <div class="wx-quick-toggle-condition flex-1" aria-label="{{"Condition" | t}}">
```
partly-igor, line 31: Similar to date the `aria-label` probably should be the actual condition, because it currently doesn't get announced

partly-igor (changes requested):
Great work on this it's looking really good. I've noted a few a11y and style issues:

## Tab/a11y
- [x] I noted inline but I think the aria-roles are not quite right because the words "Date" and "Condition" are repeated for each day in the QuickForecast, but the values are not announced
- [x] Without VO turned on, if I tab to the QuickForecast, it doesn't totally follow the WCAG keyboard pattern. The following are missing: 
  - Can't use the left/right arrows to navigate the tabs, I have to use the tab-key to do so.
  - Enter selects a day, but space doesn't
  - Home/end keys to go the start/end respectively 
  
  ## Style
- [ ] The starting breakpoint for the tab version should be `tablet-lg` instead of `desktop`.  However, if  that's going to cause a punch of pain to adjust, let me know. 
- [ ] The expand/collapse button the accordions isn't quite centered on tablet
screenshot showing the center line for the accordion
- [ ] On the accordion version there is shadow on the top portions of the details section, which makes it look a little odd. Wondering if we should just have the shadow on the outer list of all the accordions.
Screenshot of on the top portions of the details section
- [ ] Similarly with there is a drop-shadow but no left/right border on the expanded portion which makes it look misaligned. On mobile and tablet (regular) there should be a border on the expanded portion (matching the style of the border around the quick-toggle)
Screenshot of the shadow on the side of the accordion
- [ ] Noted in-line but the the headings above the quick toggles should be a little smaller, can be the `xs` size

web/themes/new_weather_theme/templates/partials/quick-forecast.html.twig
```diff
@@ -1,34 +1,9 @@
...
-    <div class="wx-quick-forecast-header-bottom">
-        <div class="wx-quick-forecast-header-empty"></div>
-    </div>
-</div>
-<nav class="wx-quick-forecast margin-bottom-4">
-    {% for day in days %}
-        {% include '@new_weather_theme/partials/quick-forecast-link-item.html.twig'
-        with { day: day }%}
-    {% endfor %}
+<nav class="wx-quick-forecast grid-col-12 margin-bottom-4" aria-label="{{'daily forecast tab navigation'}}">
```
partly-igor, line 4: Want the tab controls to feel a little closer to the data, could you change the margin bottom here to `margin-bottom-2`.

Should the `aria-label` have the translation filter added to it?

eric-gade:
@partly-igor these should be taken care of now, aside from the breakpoint. I've made https://github.com/weather-gov/weather.gov/issues/2110 for that

partly-igor:
Getting really close. Still seeing a couple of small issues:

## Screenreader
I saw you updated the aria-labels so I am not sure why it's not announcing it correctly but
- [ ] On desktop I'm still hearing "Date Condition" instead of the values for each day.
- [ ] On tablet VO announces "HIGH LOW" for each day instead of the values
- [ ] On tablet the announced state is always "expanded" even when the button / accordion is collapsed (that is in both states it announces "expanded"). 
Screenshot of voiceover announcement for the quick forecast on desktop

Screenshot of voiceover announcement for the quick forecast on tablet

## Style
- [ ] I noticed that there is a nested `grid-container` with the same classes as the outer one, so want to make sure that's not an accidental duplicate?  The result is the correct padding for desktop (total of 4 units), but incorrect for tablet (has 4 should be 2), which creates a bit of misalignment on tablet. HTML structure might be right so don't have to mess with that, but padding needs to be adjusted. 
Screenshot of the nested `grid-container` divs
Screenshot of the misaligned elements on tablet
- [ ] On desktop the first day has the correct distance between the Quick Forecast and the data, but all of the other days seem to get extra margin/padding 
Screenshot comparing margin below day 1 of quick forecast and day 2

eric-gade:
@jamestranovich-noaa for whatever reason, the libraries file is not showing up in the "changed files" tab, even though I've bumped the version several times. You can even see it in this most recent commit: https://github.com/weather-gov/weather.gov/pull/2095/commits/bb0c377e54ed04094841429a8e0b27b5d72224c5
  
Github will not let me merge this until your requested changes are approved. If either you or @partly-igor want to go ahead and approve it, feel free to merge this in!

partly-igor (approved):
Testing approving again to see if it works

The library change shows up for me under the files changed

image

igorkorenfeld:
Wondering if @jamestranovich-noaa has to be the one to approve since he requested the changes?

jamestranovich-noaa (approved):

PR #2097, opened 2024-11-20
set up a separate network for the test env

jamestranovich-noaa:
## What does this PR do? 🛠️

Let's not reuse the same network for our dev/test environments -- this causes errors.


eric-gade (approved):
💯

Issue #1887, opened 2024-10-08
DSS Builder PDF upload parity changes with Weather Stories


jamestranovich-noaa:
### Description

We have made a lot of changes to Weather Story upload content types that should be copied over to PDF uploads as well. 

- [ ] Optional attributes
  - `StartTime` (Unix epoch time)
  - `EndTime` (Unix epoch time) 
  - `Order` (integer)
  - `FrontPage` (boolean)
  - `Title` (CDATA string)
  - `Description` (CDATA string)
- [ ] Filesystem upload locations (after #1886 goes in)

### Acceptance criteria

- [ ] Outside tests for optional attributes
- [ ] Changes have been tested for accessibility

colinmurphy01:
Moving to the backlog for consideration next sprint.

colinmurphy01:
Is this `done`?

colinmurphy01:
@jamestranovich-noaa is this `done`?

jamestranovich-noaa:
Not yet. I need to finish up weather stories, and then we can circle back to this. I've been holding off on parity for PDF uploads until we get the newer DSS workflow settled in (particularly overwriting of uploaded files).

PR #2103, opened 2024-11-21
dss builder pdf uploads: add optional attrs

jamestranovich-noaa:
## What does this PR do? 🛠️

Adds optional attributes for PDF uploads. Also overwrites previous PDF files (similar to weather story uploads).

Closes https://github.com/weather-gov/weather.gov/issues/1887


eric-gade (approved):
Code LGTM! 👍

Issue #2084, opened 2024-11-15
[Bug] Alerts are the wrong color on the forecast item


partly-igor:
### Description
In a number of instances the `wx-alert-link` is the wrong color for on the forecast item in the 7 days tab. It is displaying in the gray color that is meant for advisories/statements (i.e. the default), rather than the color for a warning or a watch.  This seems to be an issue when there is only one hazard for the day – as you can see in the second example below. In the second example the alert for Saturday, Nov 16 is in the correct color, given that there is both an advisory and a watch. But the alert for Sunday, Nov 17 is wrong for the watch. 

Unfortunately I wasn't able to capture bundles for these. 

Examples:
![Screenshot of forecast for NYC with the wrong color for fire weather warning](https://github.com/user-attachments/assets/68482d3b-c1ea-485d-a97c-9baa5a400281)

<img width="1011" alt="Screenshot of forecast for Stanford, MT with the wrong color" src="https://github.com/user-attachments/assets/e0dfffc8-3286-4c6a-b16f-564c02b19e7d">

### Acceptance criteria

- [ ] On the daily forecast items alerts for warnings show the correct red colors, alerts for watches show the correct yellow colors, and alerts for anything else show the correct gray colors. 
- [ ] Forecast items with  multiple alerts have the color corresponding to the highest alert level.
- [ ] Tests have been implemented or modified to capture the bug
- [ ] Changes have been tested for accessibility

PR #2105, opened 2024-11-21
Adjust alert link color and alerts padding

partly-igor:
## What does this PR do? 🛠️
This small PR updates the alert link color noted in #2084 and makes an unrelated adjustment  to the padding of the alerts container so that it aligns better to the other elements on desktop 

Screenshot with corrected alert colors

Screenshot of the adjust padding on the alerts container


sspj-does-weather (approved):
LGTM 👍

PR #2123, opened 2024-11-25
Fix the alert metadata page

### LINKED ISSUE #1993: Updating the hazard priority list (with interop service)
With the update to the interop layer - [this page](https://beta.weather.gov/info/alerts) is likely going to break.

greg-does-weather:
## What does this PR do? 🛠️

- Adds a new route controller for site metadata
- Update the `/info/alerts` route to use the new controller
- The controller fetches alerts from the interop layer
- Maps the route to a "theme"
- Renames the Twig template
- Update the Twig template
- Removes some unnecessary `attach_library` statements from the location search template

- closes #1993

image


web/modules/weather_routes/weather_routes.routing.yml
```diff
@@ -53,7 +53,7 @@ weather_routes.grid.legacy:
 weather_routes.alert-info:
   path: '/info/alerts'
   defaults:
-    _controller: \Drupal\weather_routes\Controller\LocationAndGridRouteController::noop
```
jamestranovich-noaa, line 56: Do we still need the `noop`? OK with leaving it in but just wondering as this was (afaics) the last usage of it

jamestranovich-noaa (approved):

web/modules/weather_routes/weather_routes.routing.yml
greg-does-weather: Good question! If it's not being used, I can just yoink it out. If it ever needs to come back, it's not exactly hard, and we'll also have git history to refer to. :)

Issue #2001, opened 2024-10-28
[Interop] Use uncaught exception handler to gracefully restart


greg-does-weather:
### Description

The interop layer currently has a handler for exceptions that are not caught elsewhere. Its behavior is to log the error and continue running. However, because unhandled exceptions necessarily indicate a previously unknown error state, we cannot be certain that the interop layer's internal state is consistent and we should not trust it to keep operating correctly. Indeed, the [Node.js documentation](https://nodejs.org/api/process.html#warning-using-uncaughtexception-correctly) makes this exact argument.

Instead, we should consider a very slim launcher application whose whole job is to monitor for unhandled exceptions in the interop layer, gracefully shut it down, and relaunch it.

### Acceptance Criteria
- [ ]

` Tag the related milestone to this story`

colinmurphy01:
It's hard to grok the impact or priority of this work. @eric-gade would be interested in your thoughts, moving to `backlog` for consideration.

jamestranovich-noaa:
I still plan on doing this, with the caveat that my priority is still finishing up weather stories though.

eric-gade:
@colinmurphy01 let's prioritize this I think it's a good idea

PR #2125, opened 2024-11-26
Setup PM2 for interop layer

jamestranovich-noaa:
## What does this PR do? 🛠️

With PM2, we can now gracefully restart when the API interop layer goes down, and so we exit (error code 1) every time we get an uncaught exception. Closes https://github.com/weather-gov/weather.gov/issues/2001

Does not fix our memory issues but might give us more insight on what's happening under the hood.

- Added a new Makefile target: `im`, so we can peer inside the pm2 and see the metrics/logs. (This is for dev purposes only)
- We are not using `nodemon` anymore for dev, but are using PM2 for both dev/prod, which makes our environments more consistent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,10 @@ ct: check-translations
 check-translations: ## Check the consistency of translations
 	npm run check-translations
 
+im: interop-monit
+interop-monit: ## Check interop API	metrics and logs
+	docker compose exec api-interop-layer npx pm2 monit
+
 ### Composer management
 ci: composer-install
 composer-install: ## Installs dependencies from lock file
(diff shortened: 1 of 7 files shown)


eric-gade (approved):
This is awesome! 🥳 
  
I am bypassing the normal checklist here, since there's nothing specifically web related.
  
One question I have is this: when using the monitor command in live environments, is the expectation that we are already SSH'd into an interop instance and we run it from there, or is there some kind of remote client we use?

jamestranovich-noaa:
> One question I have is this: when using the monitor command in live environments, is the expectation that we are already SSH'd into an interop instance and we run it from there, or is there some kind of remote client we use?

Yes, we will need to `cf ssh` in to use this command. PM2 does offer remote monitoring via pm2.io but we cannot use it here -- annoyingly, pm2 will prompt you to login pm2.io if you type in `pm2 monitor` instead of `pm2 monit`.

Issue #2044, opened 2024-11-05
Tonight’s low in the wrong column


coreypieper:
The tonight’s low is in the “High” temperature column in the Quick Look at the top of the page. It’s rightly listed as the Low lower on the page. 

![IMG_8232](https://github.com/user-attachments/assets/c66af678-642a-4c35-9369-a0c16ca7ba20)
![IMG_8233](https://github.com/user-attachments/assets/c2c87ec0-24fb-4cfb-bb82-b30586fbd68e)

coreypieper:
Looks as though now the low is in both columns the first period (and maybe is should say “Tonight”?)
![IMG_8278](https://github.com/user-attachments/assets/6d3e2efd-9c7c-4df8-9289-ad757053894c)

colinmurphy01:
What time did you see this occurring?

eric-gade:
Closed by #2140

PR #2140, opened 2024-12-03
Deal with Today evening time issues

### LINKED ISSUE #2073: Date on 7-day forecast incorrect
### Description

If a user is looking at beta.weather.gov in the evening of 11/11, they currently see the date of 11/12, and they see the words "Today" instead of "Tonight", and they see the high temperature and low temperature instead of only the low temperature.  

image

### Acceptance criteria

A user should only see the current date.  If it is nighttime, the descriptor should read "Tonight" instead of "Today", and we should explore/discuss only showing the overnight low temperature instead of equal high and low temperatures.

eric-gade:
## What does this PR do? 🛠️
This PR directly addresses #2073 and #2044.

The issues were, respectively:
* For "nighttime" periods (between 6pm-Midnight) on the current day, the high temp was being displayed in the wrong place / doubled from the low in both the quick forecast and the quick toggles
* The numeric day was wrong for these same kind of periods -- the day number would have the same value as the following day

### Day number issue ###
The day numbering issue found in #2073 is an interesting one, and it's worth writing out what the underlying problem was.
  
In the interop code prior to this PR, we formatted the numeric day based on dayjs timestamps that were _not_ being appropriately aligned to the given timezone. The result was that the system saw a UTC timestamp like:
```
2024-12-03T01:30:00.000Z
(1:30 AM on the 3rd UTC time)
```
as being on the 3rd, even though the timezone for the period in question is offset from UTC in most places. In the case of the bug in question, an example matching timestamp would be:
```
2024-12-02T08:30:00-05:00
(8:30PM on the 2nd in NY)
```
These timestamps are for the same time. But in UTC it is after midnight, therefore it's the 3rd, and in local time it's before midnight, therefore it's the 2nd.
  
The problem was we were formatting the day and month numbers based on the UTC time and not the local/timezone mapped time.
  
### Labelling issues with the "night" period ###
Similarly, we have updated some of the labels for the current day where the user is viewing the "night" period (between 6pm and Midnight) locally.
  
Instead of the label "Today," we have now changed this specific time period to have "Tonight" as the label.
  
Instead of displaying the High temp as equivalent to the Low temp, we give the High temp a label of "N/A"

### Updating the proxy bundle format ###
One problem we faced was reliably reproducing this issue for testing and development purposes. The options were to either wait until it was nighttime wherever I happen to be in order to work on the problem, or to be able to reliably "fake" the time locally.
  
Currently, our bundles save all their data with timestamps that will be relative to the time that the bundle is loaded, ie, the time you view the bundle in the browser.
  
With this PR, we introduce a new meta-property on bundles called `now` that takes an ISO-8601 timestamp as its value.
 If that property is present, the proxy will load all bundle timestamps as if they are relative to the given "now" time (rather than the actual time). This allows us to reproduce the issue.
  
Included in this PR is also some extra point information for Wahoo, Nebraska (the "home office," for those in the know). It contains this metadata and therefore loading that page reliably shows the "night" period at all times.

<details>
<summary>Tablet/Mobile (quick-toggle version)</summary>

Screenshot 2024-12-03 at 2 43 42 PM

</details>

<details>
<summary>Desktop (quick forecast version)</summary>

Screenshot 2024-12-03 at 2 43 25 PM

</details>

## Closes
#2073 
#2044


tests/api/serve.js
```diff
@@ -111,6 +111,13 @@ export default async (request, response) => {
     console.log(`LOCAL:    serving local file: ${filePath}`);
     const output = JSON.parse(await fs.readFile(filePath));
 
+    // If the bundle contains a `now` key, override any
+    // configured value for the current time with the timestamp
+    // present in the bundle
+    if(output["@bundle"]?.now){
+      config.now = output["@bundle"].now;
+    }
+
```
eric-gade, line 120: Here's where we read and set the "now" time from the bundle metadata, if present

jamestranovich-noaa (approved):
LGTM! Great investigative work! :male_detective:

PR #2150, opened 2024-12-04
rename zscaler pem as it is a public cert

jamestranovich-noaa:
## What does this PR do? 🛠️

Gitlab will not allow us to push pem files. This changes the zscaler pem name to clarify that it is indeed a public certificate.

Nada


sspj-does-weather (approved):
LGTM 👍

Issue #2152, opened 2024-12-05
Consolidate tab components


eric-gade:
### Description
Right now, we have several independent web components that each implement the WCAG Tabs pattern, but in slightly different ways.
  
This ticket calls for the creation of a generic Tabs component that can be reused -- either through inheritance, custom event handling, or composition.

### Acceptance criteria

- [ ] We implement a generic Tabs web component that can be reused with custom styling
- [ ] Existing tabs components are extended from, composed from, or use custom events from this generic tabs component
- [ ] Changes have been tested for accessibility

Issue #1944, opened 2024-10-18
Make R7 forcast details toggle more accessible


eric-gade:
### Description
With #1926 and #1943, we implemented a toggle for the daily forecast details, where users can quickly switch between viewing just the table or just the charts (to reduce clutter on the screen).
  
However, I believe there are a11y problems with the way we've implemented this. For the moment, it is simply a USWDS button group with some javascript that enables the switching of what is displayed. But what we probably want is something more like the [WCAG tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) under the hood.
  
We still want the visual behavior to be the same as in the designs, but we want to ensure that screen reader users can understand what is going on, what is or is not being displayed, and how to switch properly between the displays.
 

### Acceptance criteria

- [ ] The forecast details toggle system is WCAG tabs compliant

partly-igor:
Writing down thought so I don't lose it – I wonder if in place of tabs we might use radio buttons, styled as the toggle. The benefit is that it would announce which tab is selected, and allow for the more default key behavior.

colinmurphy01:
This is blocked by #1967

Issue #2153, opened 2024-12-05
Forecast charts announce "Empty group"


partly-igor:
### Description

When a user using a screen reader navigates to the charts view in a forecast-item and moves focus to a chart, the screenreader announces "empty group" (when tested using VO). We currently don't have an aria-role or text alternative within the canvas element, and the table that contains the text data is hidden, so screenreader users are not able to get any information from the chart. 

As discussed with @eric-gade possible solutions include:
- Refer to table rows by aria-labelledby
- Refer to the whole table by aria-labelledby
- Custom SR only explaining that they should switch to table view
  

### Acceptance criteria

- [ ] Navigating to a chart a screen reader user is able to get meanigful information - either the data or a direction to the table containing the data.

PR #2158, opened 2024-12-05
Make forecast details buttons WCAG tabs (a11y)

eric-gade:
## What does this PR do? 🛠️
As discussed in #1944, the toggle buttons for switching between charts and the table for forecast details are not accessible. This PR does several things (noted below), but mainly changes those buttons to be [WCAG compliant tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

### Generic Tabs component, and proposal for cleanup ###
Right now, we have **three** separate components that each implement the tabs functionality. This is a lot of repetition, especially when it comes to binding the keyboard navigation events.
  
For this work, we have created a _generic_ Tab component (`wx-tabs`) that we hope the other tab components can be refactored from. We've even made a [ticket for that work here](https://github.com/weather-gov/weather.gov/issues/2152).

### Tests ###
We have added some JSDom based component tests. We need to make sure these are running properly as part of the code-standards CI.
  
### Other issues ###
We had to modify some of the markup. There was one place in which all of the forecast details tables had the same id across the document -- huge NO-NO!

## How to test?
Pull down this branch and then use VoiceOver to ensure that the Table/Charts button group is being interpreted as a tablist. Test the keyboard navigation events that are specified in the [WCAG recommendation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

Nope -- it looks exactly the same!


web/themes/new_weather_theme/tests/components/generic-tabs-tests.js
```diff
@@ -0,0 +1,172 @@
+/* eslint no-unused-expressions: off */
+import { expect } from "chai";
```
sspj-does-weather, line 2: For writing a11y-aware tests in JS, I'm a pretty big fan of https://testing-library.com/ ... not suggesting to rewrite these tests, but maybe to include it later on?

For example, your first button should have an accessible label of `Tab 1 content`. With chai, you have to write a specific assertion for that, but with testing library, you just grab the button by its accessible label (rather than `querySelector`), so if the label doesn't exist in the accessibility tree, the test fails and you don't have to write any extra code at all.

eric-gade: Good idea. I might roll that into the tab consolidation ticket #2152

partly-igor (approved):
This looks great from design perspective. All the tabs/button groups continue to work well. 

The things that I'll note that might be an issue (and we can confirm during our a11y review), but can be covered by follow-up tickets.
- On desktop, when any day is selected under 7 days we only display 1 day of content but VO announces a list of 7 items in the tab panel. 
- I think it's probably ok as is, but flagging that for the table/charts between the tablist and the tabpanel is the `h4` "Hourly forecast". I think this ok because it's descriptive, but noting that we should listen for this when we do our review. 
- Under the Today tab, the today forecast is a list of 1 item – not sure if it's odd for SR users to hear that list of 1 item. 
- The charts are still announced as empty group (this will be covered in work for #2153)

PR #2165, opened 2024-12-06
Remove fill from condition icons

partly-igor:
## What does this PR do? 🛠️
This PR removes fills from a few of the icons for weather conditions that had them included as part of their path. This was causing the icons to render with the incorrect fill when placed on on the dark background within the quick forecast, which was causing contrast issues making the icon difficult to see. 

When testing these changes locally I was encountering a problem with the browser rendering the old icon after updating. I believe this may be a caching issue – though running `make cc` and emptying the cache did not always resolve the problem.  What seemed to work somewhat consistently is rebuilding, running `make cc`, "Empty cache and hard reload", and then navigating to a new location. 

Screenshot of the hurricane icon
Screenshot of the dust icon
Screenshot of the snow-sleet icon

Screenshot of the windy-cloudy icon


jamestranovich-noaa (approved):
yay! LGTM

Issue #1675, opened 2024-08-30
Fix flaky tests


colinmurphy01:

colinmurphy01:
Related to #1440

colinmurphy01:
Considerations:
- Transition all Cypress tests over
- Identify the tests themselves that should be prioritized
  - Anything dealing with location combo-box is typically going to be flaky.

colinmurphy01:
This probably blocks #1440

colinmurphy01:
This has been a PAIN recently. Moving up so we can tackle it in the next sprint.

colinmurphy01:
Eric's got some ideas to try, let's implement those this sprint and see if it solves.

PR #2177, opened 2024-12-11
Browser test tweaks

eric-gade:
## What does this PR do? 🛠️
This PR addresses the ongoing ticket #1675, and appears to ameliorate _some_ of our browser test flakiness.

Here's the rub: currently, any custom element components are all bundled up with the functional (non-custom element) components on the point page. These are loaded into a single script tag in the body that is given the "defer" attribute. For custom elements, this means that the functionality of the components might not be fully loaded at page "load" time. That means we are testing for functionality that in a lot of cases hasn't loaded yet (and so the tests fail). It's a matter of timing.
  
To get around this, we have separated actual custom elements into a separate library bundle for the point page called `point.page.components.js`. This is now loaded without the "defer" attribute, and is also present in the head of the page to ensure priority loading.
  
Additionally, we have annotated most of the `page.goto` calls in our tests to explicitly wait for the `load` event. Note that this is actually the default behavior for `goto` anyway. We simply wanted to be explicit, so that later on if we need to tweak these tests further for performance reasons, we know where we can change things.
  
Testing this locally, all of the e2e tests seem to pass with pretty good regularity (where they weren't necessarily before).

### DIFF
--- a/tests/playwright/e2e/afd-page.spec.js
+++ b/tests/playwright/e2e/afd-page.spec.js
@@ -33,6 +33,7 @@ describe("AFD Page Tests", () => {
     }) => {
       const response = await page.goto(
         "http://localhost:8080/afd/OKX/invalid-afd-id.json",
+        { waitUntil: "load" }
       );
 
       expect(response.status()).toBe(404);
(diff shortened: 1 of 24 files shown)


jamestranovich-noaa (approved):
amazing work! LGTM

PR #2178, opened 2024-12-11
Interop alert loop optimization

eric-gade:
## What does this PR do? 🛠️
This PR partially satisfies interop layer refactoring (in the hopes of reducing memory consumption in the alert caching).
  
So far, the effects on memory consumption seem minimal. 

In our [proposal doc](https://docs.google.com/document/d/1xhjscbNdmVywgZgvMksBKP-cJEpfomSMHAIvS7Zyw30/edit?usp=sharing), the first item we decided to try was to "chunk" how complex geometries are union-ed in memory. We decided that we should process a maximum of 5 zones at a time, which is what this PR implements.
  
Be aware that only forecast zones and county zones are relevant to this optimization.

### DIFF
--- a/api-interop-layer/data/alerts/geometry.js
+++ b/api-interop-layer/data/alerts/geometry.js
@@ -1,6 +1,14 @@
 import { simplify, union } from "@turf/turf";
+import openDatabase from "../db.js";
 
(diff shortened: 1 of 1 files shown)


jamestranovich-noaa (approved):
Nice! Easier to scan and understand.

Issue #1708, opened 2024-09-06
Gitlab transition (phase 1): Setup CI pipelines


jamestranovich-noaa:
### Description

We want to port over Gitlab Actions to Gitlab CI Pipelines so that we can test and vet Gitlab runners. The [`code-standards.yaml` ](https://github.com/weather-gov/weather.gov/blob/main/.github/workflows/code-standards.yaml) has several steps and dependencies that should be relatively easy to move over to pipelines. There are other Github actions, but this one is probably the most used Github Action for new commits and PRs. 

We can ignore ADR workflows for now. These will be added later. 

We can ignore caching for now, as that is a process that doesn't work for us all the time, and might be solved with Gitlab docker repositories.

### Acceptance criteria

- [ ] We have a single Gitlab pipeline that does all of `code-standards` that runs on PRs and commits.

jamestranovich-noaa:
Closing as gitlab#28 has more details and we are going to subdivide this into two tickets: docker-in-docker playwright tests and deployment steps.

PR #2179, opened 2024-12-12
Gitlab CI: add javascript, php, style linting

jamestranovich-noaa:
## What does this PR do? 🛠️

Partially addresses https://github.com/weather-gov/weather.gov/issues/1708 by adding a preliminary CI pipeline for Gitlab. Note that Gitlab CI caching is not available (we have asked them to turn on this feature), which is why we essentially work around this by having our own `utility-node` container install all the npm dependencies.

Two things of note:

- I am leveraging the Gitlab container repository feature so that we can run our Docker images/containers directly. 
- I have also added junit style reports for linting so we can see the test results right in the MR.

I'll be documenting how to push up to the Gitlab container repository in the next PR/MR because I'm not sure how versioning should work. Right now the only tag is `latest`. TBD.

Visit [gitlab#1](https://vlab.noaa.gov/gitlab-licensed/NWS/Systems/DIS/Weather.gov-2.0/-/merge_requests/1) for what it looks like when tests pass.


eric-gade (approved):
Just checked the gitlab outputs, looks really good. There is only one part from the config I am confused about -- how often do we need to push to the container repository in order for all of this to work? Is that going to run every time before CI?

PR #2180, opened 2024-12-13
Alerts cache table

eric-gade:
## What does this PR do? 🛠️
This PR attempts to partially address interop layer refactoring. It implements the alert cache table strategy outlined in the [interop alert refactoring doc](https://docs.google.com/document/d/1xhjscbNdmVywgZgvMksBKP-cJEpfomSMHAIvS7Zyw30/edit?usp=sharing).

### A new spatial table
Alerts are now cached -- with their geometries, if present -- in a new spatial table called `weathergov_geo_alerts_cache`. This table will be dropped and recreated whenever the interop process (re)starts. It's schema is defined in code (see inline comments).
  
### `AlertsCache`
The interface into the alerts cache table is a helper object called `AlertsCache`. Instances of this object contain immutable functions for interacting with the cache table. The advantage is that you can make several instances of this object and they will not interfere with each other -- a necessity in our case since we have a main thread (pulls out alerts by point) and a worker thread (updates/invalidates cached alerts).
  
### How it works
Previously, we were storing relevant/processed alerts in memory, including any complex geometries for larger alert areas. With these changes, _no alert data is stored long term in memory_, and everything is handled by the database table.
  
The background worker's job is now to:
* Hit the alerts endpoint, generating a unique hash for each alert
* If a hash is already in the db table, skip
* If there are hashes in the db table that are _not_ in the new alerts, then we remove those from the table (they are now invalid)
* New hashes are processed, including mapping for kind and computing geometries from zones/counties if needed
* Land alerts are fully processed and stored in the db with their geometries
* All other alert types are stored in the db _without_ geometries. This ensures that we don't loop through these "seen" alerts again on the next run, but also that we don't return those alerts as valid when requesting all alerts for a lat/lon pair
  
Then in the main thread, we perform an `AlertsCache` call that takes a given geo point and queries for all alerts that intersect that point. This geographic computation is done within the database, so it's one query.

### DIFF
--- a/api-interop-layer/data/alerts/cache.js
+++ b/api-interop-layer/data/alerts/cache.js
@@ -1,0 +1,114 @@
+/**
+ * AlertsCache
+ * -----------------------------------
+ * Serves as the interface into the alerts cache database table,
+ * where active alerts and their geometries are stored.
+ * To avoid potential memory conflicts in the threaded environment,
+ * actual alert data is not stored on this object and methods are
+ * mostly immufable.
+ * Consumers can use multiple instances of this object to read vs write
+ * to the cache table.
+ */
+export class AlertsCache {
+  constructor(tableName="weathergov_geo_alerts_cache", dbConnection){
+    this.db = dbConnection;
+    this.tableName = tableName;
+
+    // Bound methods
+    this.getHashes = this.getHashes.bind(this);
+    this.determineOldHashesFrom = this.determineOldHashesFrom.bind(this);
+    this.removeByHashes = this.removeByHashes.bind(this);
+    this.removeInvalidBasedOn = this.removeInvalidBasedOn.bind(this);
+  }
+
+  async createTable(){
+    const sql = `CREATE TABLE IF NOT EXISTS ${this.tableName} (
+id INT AUTO_INCREMENT,
+hash TEXT NOT NULL,
+alertJson JSON NOT NULL,
+shape GEOMETRY NOT NULL,
+PRIMARY KEY(id)
+) DEFAULT CHARSET=utf8mb4`;
+    return await this.db.query(sql);
+  }
+
+  async getHashes(){
+    const sql = `SELECT hash FROM ${this.tableName};`;
+    const [ result, _ ] = await this.db.query(sql);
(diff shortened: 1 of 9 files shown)


api-interop-layer/data/alerts/cache.js
```diff
@@ -0,0 +1,113 @@
...
+    const sql = `CREATE TABLE IF NOT EXISTS ${this.tableName} (
+id INT AUTO_INCREMENT,
+hash TEXT NOT NULL,
+alertJson JSON NOT NULL,
+shape GEOMETRY DEFAULT NULL,
+alertKind TEXT DEFAULT NULL,
+PRIMARY KEY(id)
+) DEFAULT CHARSET=utf8mb4`;
+    return this.db.query(sql);
+  }
```
eric-gade, line 28: Schema for table defined here

api-interop-layer/data/alerts/cache.js
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Given an array of current hashes and an array of incoming
+   * hashes, return a list of hashes from the current set that are
+   * not in the incoming set.
+   * This is how we determine old/invalid hashes during an
+   * all active alerts request.
+   * NOTE: node does not yet have standardized Set methods
+   * for determining difference, so we can't do the "normal"
+   * thing here and use Sets
```
jamestranovich-noaa, line 44: boooooo this was the first thing i thought of

api-interop-layer/data/alerts/geometry.js
```diff
@@ -180,7 +180,7 @@ export const generateAlertGeometry = async (db, rawAlert) => {
     }
   }
 
-  // we cannot generate a geometry.
+  // we cannot generate a geometry.;
```
jamestranovich-noaa, line 183: typo?

jamestranovich-noaa (approved):
LGTM i like the overall structure, very clean

api-interop-layer/data/alerts/geometry.js
eric-gade: oops!

PR #2188, opened 2024-12-19
Style adjustment - Dec 2024

partly-igor:
## What does this PR do? 🛠️
This PR makes a few style adjustments and fine tuning: 
- Align the WFO promo with the other elements on the page. Now that the other element are not offset, the WFO shouldn't be Either
- Changes the capitalization of "Chance of precipitation" to match the capitalization of other chart titles
- Adds margin before the "Dewpoint" chart to match the spacing of the other charts
- Centers the "conditions" label above the quick toggles so that it's better aligned with the icon

There is a translation file update here – noting in case there is an extra step here for a release (may already be covered by other translation work). 

Screenshot of the updated alignment of the WFO promo

Screenshot of the updated spacing before Dewpoint

Screenshot of the update alignment of the "Conditions" label


web/modules/weather_i18n/translations/en.po
```diff
@@ -75,7 +75,7 @@ msgid "daily-forecast.labels.today.01"
 msgstr "Today"
 
 msgid "daily-forecast.text.chance-precip.01"
-msgstr "chance of precipitation"
+msgstr "Chance of precipitation"
```
sspj-does-weather, line 78: Sorry for causing the merge conflicts here! This string is used in several places, with different capitalization needs, so I split it into different keys.

partly-igor: No problem, thanks for explaining!

sspj-does-weather (approved):
LGTM 👍

PR #2189, opened 2024-12-19
gitlab ci: generalize build and add documentation

jamestranovich-noaa:
## What does this PR do? 🛠️

Does what it says on the tin: generalize build (does not work for `database` yet) and documents Gitlab CI as it is right now ([rendered](https://github.com/weather-gov/weather.gov/blob/fd5bf0118f2688fcc532768d481f59f0f1a6df8b/docs/dev/gitlab-ci.md)). 

- There is no good way to name an image -- it is derived from the docker compose yaml. So I add a derived variable but we might be able to clean this up later.
- A dot before the gitlab pipeline means "do not execute directly" -- here, we use `.build-image` as a template to be extended.

### DIFF
--- a/docs/dev/gitlab-ci.md
+++ b/docs/dev/gitlab-ci.md
@@ -1,0 +1,24 @@
+# Gitlab CI Configuration
+
+## General approach
+
+With Gitlab, we have an opportunity to re-use our existing Docker infrastructure for CI, especially in conjunction with the [Gitlab container registry](https://vlab.noaa.gov/gitlab-licensed/NWS/Systems/DIS/Weather.gov-2.0/container_registry). This simplifies our Gitlab CI configuration since we can build and test in CI exactly the same way as we do locally. (This approach differs from Github Actions because Actions sometimes uses alternative approaches to install the required environment, such as installing `php` via `shivammathur/setup-php@v2`.) Some changes to keep in mind:
+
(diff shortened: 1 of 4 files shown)


docs/dev/gitlab-ci.md
```diff
@@ -0,0 +1,24 @@
+# Gitlab CI Configuration
+
+## General approach
+
+With Gitlab, we have an opportunity to re-use our existing Docker infrastructure for CI, especially in conjunction with the [Gitlab container registry](https://vlab.noaa.gov/gitlab-licensed/NWS/Systems/DIS/Weather.gov-2.0/container_registry). This simplifies our Gitlab CI configuration since we can build and test in CI exactly the same way as we do locally. (This approach differs from Github Actions because Actions sometimes uses alternative approaches to install the required environment, such as installing `php` via `shivammathur/setup-php@v2`.) Some changes to keep in mind:
+
+- We want Gitlab CI to create the Docker images for us. Generally, pushing direct images to the Gitlab container repository is discouraged, as we want to ensure images were generated from a known and fixed commit for reproducibility and security reasons.
+- We may want to default to pulling from the Gitlab container repository instead of building locally (this will save time.) Of course, you can always build locally.
```
eric-gade, line 8: Does "locally" here refer to local development, or in the live environments, or something else?

eric-gade (approved):
👍 LGTM

docs/dev/gitlab-ci.md
jamestranovich-noaa: updated the wording to be clearer -- i merely meant that we can pull down the images instead of building the images locally. (totally optional, ofc)

PR #2199, opened 2024-12-31
dss builder: make wfo weather story template nicer

jamestranovich-noaa:
## What does this PR do? 🛠️

We seem to be using the weather story template for debugging, so I made it nicer and also removed a warning `The "weathergov_location_search" was not found` that we see repeatedly in production. We do not need to translate this page, I think, because, again, this is for debugging only.

Screenshot: 

image

We probably should note that this template is for debugging purposes only. It should not be accessible unless hit directly. Or, we can remove it altogether. :man_shrugging:


web/config/sync/block.block.new_weather_theme_locationsearchblock.yml
```diff
@@ -1,26 +0,0 @@
...
-    - system
-    - weather_blocks
-  theme:
-    - new_weather_theme
-id: new_weather_theme_locationsearchblock
-theme: new_weather_theme
-region: content
-weight: -2
-provider: null
-plugin: weathergov_location_search
```
jamestranovich-noaa, line 15: Do we need this at all? Not sure if we are using it...

eric-gade (approved):
I like it

jamestranovich-noaa:
@sspj-does-weather since you are taking over weather stories, please also let me know if this PR looks good (or not).

sspj-does-weather (approved):
Looks good to me as well. :+1: Thanks for making this look a little nicer, James!

PR #2200, opened 2025-01-03
bump up api interop to 1024M

jamestranovich-noaa:
## What does this PR do? 🛠️

We are currently using ~4gb out of our memory quota. Let's go ahead and give the api interop some extra memory. 

we got another instance of the api interop layer crash-loop of doom :tm: so this is (likely) a short term solution


sspj-does-weather (approved):
LGTM 👍

PR #2206, opened 2025-01-07
Update Drupal to version 10.4.1

sspj-does-weather:
## What does this PR do? 🛠️

Just what it says on the tin.

My minor concern is needing to bump s3fs but not knowing how to test that on my local. I think it will be easier to test on staging.

My major concern (that I need to solve before merging) is how database upgrades are handled in our current deployments.


loganmcdonald-noaa (approved):
phew, doesn't look like too many other things needed updating

jamestranovich-noaa:
Side note, we can test `s3fs` by doing, e.g., 

```
docker compose exec drupal drush config-split:activate cloudgov
```

to activate the cloudgov setup locally (don't forget to deactivate when you are done!). But testing `s3fs` in staging is fine too.

I think before we merge this PR, we need to preserve pre-existing weather stories (otherwise, if this PR goes in first, the database will be wiped and then we lose weather stories).

loganmcdonald-noaa (changes requested):
Blocking for now, running into issues in test. May open a different PR depending on how that goes.

loganmcdonald-noaa (approved):

PR #2223, opened 2025-01-15
Fix expired weather stories

sspj-does-weather:
## What does this PR do? 🛠️

This PR tries to limit stories selected for display to those where the current server time is within the UTC start/end times given by the weather story XML.

James, I faked some timestamps to test this on local, but I would love if we could copy a bunch of actual stories from beta. Do you already have a script that can do that?

### DIFF
--- a/web/modules/weather_data/src/Service/WeatherEntityService.php
+++ b/web/modules/weather_data/src/Service/WeatherEntityService.php
@@ -101,8 +101,10 @@ class WeatherEntityService
             ->accessCheck(false)
             ->condition("status", 1)
             ->condition("type", $nodeType)
+            ->condition("field_starttime", time(), '<=')
+            ->condition("field_endtime", time(), '>=')
             ->condition("field_office", '%' . $wfo, 'LIKE')
-            ->sort("changed", "DESC")
+            ->sort("field_order", "ASC")
             // Only get the first one.
             ->range(0, 1)
             ->execute();


web/modules/weather_data/src/Service/WeatherEntityService.php
```diff
@@ -101,8 +101,10 @@ public function getLatestWeatherStoryImageFromWFO($wfo, $nodeType)
             ->accessCheck(false)
             ->condition("status", 1)
             ->condition("type", $nodeType)
+            ->condition("field_starttime", time(), '<=')
+            ->condition("field_endtime", time(), '>=')
             ->condition("field_office", '%' . $wfo, 'LIKE')
-            ->sort("changed", "DESC")
+            ->sort("field_order", "ASC")
```
sspj-does-weather, line 107: I don't think this is exactly right: I actually think the "right" story is the tab numbers in the filename, but I'm a) not 100% confident of that, and b) not sure how to filter on that.

jamestranovich-noaa:
Thanks for pushing this up! :raised_hands: 

> James, I faked some timestamps to test this on local, but I would love if we could copy a bunch of actual stories from beta. Do you already have a script that can do that?

We can manually create weather story content types (and upload its images) via the Drupal admin, that's probably the easiest way to do it. We can use any current XML, e,g., https://www.weather.gov/source/top/wxstory/wxstory.xml -- I'll test shortly.

web/modules/weather_data/src/Service/WeatherEntityService.php
jamestranovich-noaa: I think this is fine for a first pass. We can always add derived fields later if need be.

web/modules/weather_data/src/Service/WeatherEntityService.php
```diff
@@ -101,8 +101,10 @@ public function getLatestWeatherStoryImageFromWFO($wfo, $nodeType)
             ->accessCheck(false)
             ->condition("status", 1)
             ->condition("type", $nodeType)
+            ->condition("field_starttime", time(), '<=')
+            ->condition("field_endtime", time(), '>=')
```
jamestranovich-noaa, line 105: Hmm. This is logically sound but does not work when testing this out.

I have a timestamp of `1736984938` and a very recent weather story that starts `1736906460` and ends `1736960460` so we are missing something here -- timezones? If I adjust for my timezone (GMT-8) then it works. Ugh. I don't know what is happening here. 

image

For now, I'd advise just comparing against `field_starttime` for now.

sspj-does-weather: I've spent some time playing with this, but I think this is the right behavior. If the weather story starts at `Wednesday, January 15, 2025 2:01:00 AM`, then ends at `Wednesday, January 15, 2025 5:01:00 PM`, if you tested it at `Wednesday, January 15, 2025 11:48:58 PM`, it makes sense you won't see it.

Side note: 2 am - 5 pm UTC is 6 pm - 9 am PT.

eric-gade: This also looks correct to me. The epoch times are all in the same timezone. James it looks like the timestamp you have is after the end time of this particular story, so I agree that you wouldn't be seeing it

eric-gade (approved):

web/modules/weather_data/src/Service/WeatherEntityService.php
jamestranovich-noaa: @eric-gade the issue is that the same weather story was still alive on weather.gov when I checked. I already went over this with Seamus and we tested again and it seems to be OK now. But something to keep in mind.

PR #2224, opened 2025-01-15
Fix weather story timestamps

sspj-does-weather:
## What does this PR do? 🛠️

Uses the starttime attribute rather than when Drupal created the weather story. Tests for `0` because starttime is sometimes `0`.

### DIFF
--- a/web/modules/weather_blocks/src/Plugin/Block/WeatherStoryImageBlock.php
+++ b/web/modules/weather_blocks/src/Plugin/Block/WeatherStoryImageBlock.php
@@ -41,7 +41,7 @@ class WeatherStoryImageBlock extends WeatherBlockBase
                 // turning that into a proper web URL for us.
                 $fullimage = $story->get("field_fullimage");
                 $image = [
-                    "alt" => $description,
+                    "alt" => "",
                     "uri" => $story
                         ->get("field_fullimage")
                         ->entity->get("uri")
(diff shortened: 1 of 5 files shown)


web/modules/weather_blocks/src/Plugin/Block/WeatherStoryImageBlock.php
```diff
@@ -41,7 +41,7 @@ public function build()
                 // turning that into a proper web URL for us.
                 $fullimage = $story->get("field_fullimage");
                 $image = [
-                    "alt" => $description,
+                    "alt" => "",
```
sspj-does-weather, line 44: The description is on the page in text, so I think it would be better to hide the image than to repeat the description twice. Does that sound right?

web/modules/weather_blocks/src/Plugin/Block/WeatherStoryImageBlock.php
```diff
@@ -53,6 +53,11 @@ public function build()
                     $story->get("changed")->value,
                 );
 
+                $starttime = \DateTime::createFromFormat(
+                    "M j, Y, g:i A",
+                    $story->get("field_starttime")->value,
+                );
+
```
eric-gade, line 60: Is the `starttime` field here a unix epoch integer?

sspj-does-weather: Yep, an int enforced by the db.

eric-gade (approved):

PR #2225, opened 2025-01-16
documentation: get all weather stories

jamestranovich-noaa:
## What does this PR do? 🛠️

Code to grab all weather stories from an csv file and dump them into a directory indexed by date. This code is intended for debugging purposes only; this is how I download all weather stories locally. It will download ~1.2gb and take a while.

### DIFF
--- a/docs/dev/get-all-weather-stories.py
+++ b/docs/dev/get-all-weather-stories.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+#
+
+import requests
+
+import csv
+import os
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


sspj-does-weather (approved):
Thank you for sharing this, @jamestranovich-noaa !

I guess if I didn't want 1.2 G of stories, I could just Ctrl-C out of this script?

jamestranovich-noaa:
> I guess if I didn't want 1.2 G of stories, I could just Ctrl-C out of this script?

Yup! That's what I just did :)

PR #2252, opened 2025-02-12
add opt out for weather stories

loganmcdonald-noaa:
## What does this PR do? 🛠️
We want to be able to give admins in the CMS the ability to opt certain WFOs out of having weather stories show up. This PR adds a bool field to the weather field office taxonomy, sets the default to true, and allows admins to check a box if they want to opt an office out.

Before merging, I'll test in staging where we have some stories and create documentation for our admins.


eric-gade (approved):
👍 Code looks good. I was not able to easily get a demo weather story in my env locally in order to test display. We can monkey around on staging

PR #2310, opened 2026-02-18
Add stress tests for alert geometry computation

### LINKED ISSUE #2138: API proxy: add a stress test case
### Description

We want to add a stress test (or two!) to the API proxy designed specifically to provoke a worst-case alert geometry scenario for the API interop layer. In a real world example we observed recently, we had an alert spanning 16 forecast zones, with each zone having its own complex geographic shape. The geographic query and resulting  point collection (over 10M of responses from the database) exhausted all memory on the API interop layer. 

This stress test can then be used to measure improvements to the API interop layer -- how fast do we respond with that query? How much memory do we take up? Are our optimizations for this kind of query working?

CC @eric-gade 

### Acceptance criteria

- [ ] A stress test (or two) has been added to the API proxy
- [ ] We can "measure" the API interop layer response to the above stress test(s)

nick-leroy:
## Summary
- Adds stress tests for `generateAlertGeometry()` with realistic large polygon fixtures (dense rectangles tiled to force real Turf.js union work)
- Three scenarios: typical (5 zones, 50 pts), worst-case (16 zones, 200 pts), and extreme (24 zones, 500 pts)
- Measures execution time with configurable thresholds; reports heap usage when `--expose-gc` is available
- Establishes baseline metrics for performance comparison after future refactors

## Test plan
- [ ] Run: `cd api-interop-layer && LOG_LEVEL=silent npx mocha data/alerts/geometry.stress.test.js --require mocha.js`
- [ ] All 3 stress tests pass within timing thresholds
- [ ] For precise memory metrics: `node --expose-gc node_modules/.bin/mocha data/alerts/geometry.stress.test.js --require mocha.js`

Closes #2138

### DIFF
--- a/api-interop-layer/data/alerts/geometry.stress.test.js
+++ b/api-interop-layer/data/alerts/geometry.stress.test.js
@@ -1,0 +1,218 @@
+/**
+ * Stress tests for alert geometry computation.
+ *
+ * Simulates worst-case scenarios where alerts span many
+ * forecast zones, each with complex polygon shapes. A real
+ * incident involved 16 zones producing 10MB+ of point data
+ * that exhausted the interop layer's memory.
+ *
+ * These tests verify that the geometry pipeline handles
+ * large inputs within acceptable time and memory bounds,
+ * and serve as a baseline for measuring future optimizations.
+ */
+import { expect } from "chai";
+import { generateAlertGeometry, ZONE_CHUNK_SIZE } from "./geometry.js";
+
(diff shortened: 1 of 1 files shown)


nick-leroy:
Closing — these PRs targeted drupal-main and were based on the pre-rewrite Drupal codebase. The current api-interop-layer architecture (PostGIS-side ST_Union via SQL fragments) addresses the underlying memory concern via a different design, so the original patch is no longer applicable. Thanks for considering!

PR #2311, opened 2026-02-18
Refactor alert geometry loop for cleaner chunking

### LINKED ISSUE #2139: Refactor interop alert geometry loop
### Description
As part of the effort to ameliorate memory issues during alert fetching, parsing, and caching (as discussed in [this working document](https://docs.google.com/document/d/1xhjscbNdmVywgZgvMksBKP-cJEpfomSMHAIvS7Zyw30/edit?usp=sharing)), we need to refactor the alert geometry loop.
  
You can find the relevant code [here](https://github.com/weather-gov/weather.gov/blob/main/api-interop-layer/data/alerts/geometry.js#L21).
  
The issue is that geometry fetching can, in the worse cases, return megabytes of point data. These points are then put into an in-memory `union` function in order to compute the encompassing geometry of all constituent (county or zone) geometries.
  
This ticket calls for _limiting the number of requested_ geometries we process at a time when fetching county or zone information. The size of each chunk should be configurable, but as a first pass we want to set it to 5 at the most.
  
Here is a possible algo:
1. Make a RESULT_LIST to hold the computed chunks
2. Fetch the next CHUNK_SIZE county/zone geometries from the database
3. Union the fetched chunks into a single geometry
4. Append computed union to RESULT_LIST
5. If there are county/zones remaining, go back to step 2
6. If there are no more county/zones remaining, union the geometries in the RESULT_LIST

That should, when all is said and done, get you one single union for all constituent county/zone geometries without having to do one single and potentially memory intensive fetch and union computation.

### Acceptance criteria

- [ ] Code has been refactored to only fetch and process X (5) number of county/zone geometries at a time
- [ ] Test the code with the worst case scenario locally in a low-memory docker instance, and observe memory consumption

nick-leroy:
## Summary
- Replaces the manual `while` loop in `fetchAndComputeZoneGeometries()` with a clean two-phase approach: chunk the zone list with a `chunkArray` helper, compute per-chunk unions, then merge results via reduce
- Each chunk's raw DB data can be GC'd after its local union, reducing peak memory for large alerts
- Adds guard clause to `chunkArray` for non-positive sizes
- Fixes pre-existing test bug: `Array(n).map()` sparse array pitfall replaced with `Array.from()`
- Renames misleading `numChunks` test variable to `totalZones`

## Test plan
- [ ] Run: `cd api-interop-layer && LOG_LEVEL=silent npx mocha data/alerts/geometry.test.js --require mocha.js`
- [ ] All 8 tests pass
- [ ] Re-run stress test from #2138 to compare performance before/after

Closes #2139

### DIFF
--- a/api-interop-layer/data/alerts/geometry.js
+++ b/api-interop-layer/data/alerts/geometry.js
@@ -115,42 +115,71 @@ const getUnion = (firstShape, secondShape=null) => {
 };
 
 /**
- * Here a 'zone' can refer to either a forecast zone or
- * a county fips id. Either one will produce a geometry set
- * that is stored in the database.
+ * Splits an array into groups of at most `size` elements.
+ * Avoids the manual index juggling of the old while-loop
+ * approach, and the output is easy to iterate over.
+ */
+const chunkArray = (array, size) => {
+  // Guard against non-positive chunk sizes which would
+  // cause an infinite loop in the for-loop below
+  if (size <= 0) {
+    return [array];
+  }
+  const chunks = [];
+  for (let i = 0; i < array.length; i += size) {
+    chunks.push(array.slice(i, i + size));
+  }
+  return chunks;
+};
+
+/**
+ * Fetches and merges zone/county geometries in bounded
+ * batches to keep memory usage under control.
+ *
+ * The algorithm (per the issue description):
+ * 1. Slice the zone list into chunks of ZONE_CHUNK_SIZE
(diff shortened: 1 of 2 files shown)


nick-leroy:
Closing — this PR targeted drupal-main and refactored a fetchAndComputeZoneGeometries function that no longer exists in the post-rewrite api-interop-layer (the union now happens in PostGIS via ST_Union). The architectural concern is addressed; the patch is moot. Thanks for considering!

PR #2376, opened 2026-06-02
Fixup links from the wa-wa map dev tool

mgwalker:
The URLs on beta changed from `/points/[lat]/[lon]` to `/forecast/points/[lat]/[lon]`. This PR updates the dev tool version of the wawa map accordingly.


jamestranovich-noaa (approved):
thanks! ❤️

